Issue #15, opened 2020-11-05
Support direct .__await__


uwinx:
Hey there, seems like a nice library. It's just a small suggestion and opinion, but you can add `__await__` for the queries, instead of kind of an annoying `.run` every time

```python3
class Q:
    async def _actual_call(self) -> T:
        # .... do_some_work .... '
        return 

    def __await__(self) -> Generator[None, None, T]:
        return self._actual_call().__await__()

q = Q()
await q
```

Suggestion related to `picollo/query/base::Query`

dantownsend:
@uwinx That's a great suggestion - thanks. The original design intention for Piccolo was to make it as explicit as possible when database queries were actually being made. Some existing ORMs do database queries when accessing properties etc, which can lead to performance issues. But I have to admit, the explicit `.run()` can be a pain, and I sometimes forget it.

And as you're calling `await` on the query, it's still explicit that you're expecting it to do something.

I've created a pull request for the change. I'll mull it over tonight, in case there's some edge case I haven't considered.

sinisaos:
@uwinx Maybe is annoying and sometimes I forget to call `.run()` but I like explicit way to run query because I know exactly what need to do. And like docs say's - "Until you execute one of those methods, you can chain as many methods onto your query as you like, safe in the knowledge that no database queries are being made." I look at that like advantage.

dantownsend:
@sinisaos Yeah, there are pros and cons to each approach.

The chaining of methods will still work, with or without the explicit `.run()`:

```python
await MyTable.select().first().run() 
await MyTable.select().first()
``` 

I'm thinking of merging the changes in, and having it as a convenience. What do you think?

The question then becomes - would you do the same thing for the synchronous version. You could set the `__call__` method for `Query`, and then do this:

```python
MyTable.select()()
# Equivalent to:
MyTable.select().run_sync()
```

It looks a bit weird though.

sinisaos:
@dantownsend Async way looks good and there is good convenience while the functionality stays the same. But Piccolo provide both Async and Sync support and you're right about the Sync way looks weird.

Personally, I like the explicit `.run()` and `.run_sync()`.

dantownsend:
[...]

PR #16, opened 2020-11-05
adding __await__ magic method to Query

dantownsend:
Following this suggestion from Martin:

https://github.com/piccolo-orm/piccolo/issues/15

### DIFF
--- a/piccolo/query/base.py
+++ b/piccolo/query/base.py
@@ -87,6 +87,13 @@ class Query:
         """
         pass
 
+    def __await__(self):
+        """
+        If the user doesn't explicity call .run(), proxy to it as a
+        convenience.
+        """
+        return self.run().__await__()
+
     async def run(self, in_pool=True):
         self._validate()
 
--- a/tests/query/test_await.py
+++ b/tests/query/test_await.py
@@ -1,0 +1,18 @@
+import asyncio
+
+from ..base import DBTestCase
+from ..example_project.tables import Band
+
+
+class TestAwait(DBTestCase):
+    def test_await(self):
+        """
+        Test awaiting a query directly - it should proxy to Query.run().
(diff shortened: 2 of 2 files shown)


piccolo/query/base.py
```diff
@@ -87,6 +87,13 @@ def _validate(self):
         """
         pass
 
+    def __await__(self):
+        """
+        If the user doesn't explicity call .run(), proxy to it as a
+        convenience.
+        """
+        return self.run().__await__()
```
uwinx, line 95: if you want to keep respecting `run`'s `in_pool` flag in `__await__`, you can add `await_in_pool` property flag (setter+getter) to `Query`

dantownsend: @unwinx How do you envisage that working?

Like this, or something else?

```python
query = MyTable.select()
query.in_pool = False
await query
```

uwinx:
exactly. that'd be "just in case user wants X" 18:40, 6 November 2020, Daniel Townsend <notifications@github.com>:
@dantownsend commented on this pull request.



In piccolo/query/base.py:
> @@ -87,6 +87,13 @@ def _validate(self):
         """
         pass
 
+    def __await__(self):
+        """
+        If the user doesn't explicity call .run(), proxy to it as a
+        convenience.
+        """
+        return self.run().__await__()

@unwinx How do you envisage that working?
Like this, or something else?
query = MyTable.select()
query.in_pool = False
await query

piccolo/query/base.py
uwinx: yes, exactly. as a "just in case" feature

dantownsend: I'll leave this out for now - people can use `.run()` if they need more control.

PR #21, opened 2020-11-30
JSON field

dantownsend:


dantownsend:
- Added a basic JSON and JSONB column type
- Aliases can be used in select queries

PR #30, opened 2021-01-01
Foreign key references as string

dantownsend:
The ``ForeignKey`` ``references`` argument can now be specified using a string, or a ``LazyTableReference`` instance, rather than just a ``Table`` subclass. This allows a ``Table`` to be specified which is in a Piccolo app, or Python module. The ``Table`` is only loaded after imports have completed, which prevents circular import issues.

Other issues fixed:

 * Faster column copying, which is important when specifying joins (e.g. ``await Band.select(Band.manager.name).run()``).
 * Fixed a bug with migrations and foreign key contraints.
 *  Increased test coverage.


scott2b:
I am still seeing circular import issues even when using lazy references. I have a project with structure like this:

users
    tables.py
    views.py

accounts
    tables.py
    views.py

users.tables.User is a subclass of BaseUser, and accounts.tables.UserAccount has the following reference:

```
    user = ForeignKey(
        references=LazyTableReference(
            table_class_name="User", module_path="users.tables"
        )
    )
```

The users.tables module does not reference the accounts package at all, and there are no  explicit imports of users into accounts, just the lazy reference.

Now, the problem occurs in users.views where there are explicit imports of both User, and UserAccount in order to make use of both of these in my user profile views. When that import is attempted, I get an error like:

```
ImportError: cannot import name 'UserAccount' from partially initialized module 'accounts.tables' (most likely due to a circular import) (/usr/src/./accounts/tables.py)
```
I realize there is not a lot to go on here without building out a demonstrating isolated version of the problem, but any ideas that come to mind as to how I might work through this would be appreciated. I have tried re-ordering the User and User account imports in the view, but to no avail. I suppose if I really have to, I can consolidate these classes into a single module, but organizationally I see these as distinct concepts that I'd like to keep separate if at all possible.

fwiw I was on piccolo 0.73.0, have updated to 0.82.0, and am still seeing this problem.

scott2b:
For the time being, I have moved my UserAccount import into the view function that uses it. That will have to do for now.

PR #52, opened 2021-01-26
Timestamptz column

dantownsend:
Adding a `Timestamptz` column type for storing timezone aware datetimes.

Needs fully testing with the Piccolo Admin before merging in.


dantownsend:
Tested with Piccolo Admin, and works as expected.

Issue #50, opened 2021-01-26
Windows may not support file names ":"


zhanghaofei:
![image](https://user-images.githubusercontent.com/14348610/105808693-294c1900-5fe3-11eb-89fc-3f22d11a6c25.png)

dantownsend:
@zhanghaofei Thanks for reporting this.

I haven't tested it much on Windows, but this is a fairly easy fix.

I'll replace the colons with underscores. The file names don't really matter - just what's inside.

Out of interest, are you using PowerShell, or Windows Subsystem for Linux?

zhanghaofei:
Yes, but it makes it impossible to clone the code.I am using PowerShell, I have tried WSL2 is normal, can be cloned into the code, but back to PowerShell and the file manager is still a problem, as shown below:
![image](https://user-images.githubusercontent.com/14348610/105826719-01b57a80-5ffc-11eb-9875-0507c0ce5f87.png)

dantownsend:
I see - I should be able to fix this today. I just need to rename those files to they include underscores instead of colons, and make it so new migration files don't include colons.

dantownsend:
@zhanghaofei I've changed any filenames containing colons. Can you verify that cloning the repo works for you? If so I'll push the changes to PyPI - thanks!

zhanghaofei:
I have verified it. It is normal. Thank you

dantownsend:
The changes are on PyPI now. If you encounter any other issues with Windows let me know.

I need to try and setup Windows testing as part of the CI pipeline.

PR #53, opened 2021-01-26
Windows fixes

dantownsend:
Colons aren't allowed in filenames on Windows, so changing the name of existing migration files, and making sure new migrations don't contain colons.


dantownsend:
https://github.com/piccolo-orm/piccolo/issues/50

Issue #55, opened 2021-02-02
Implementing custom column types


Fingel:
Hello,

I just recently stumbled upon Piccolo and I'm very impressed so far with how much thought and effort has been put into it. Modern Python desperately needs an ORM that can keep up with the next generation of web development libraries. Piccolo looks to be well on it's way to fill that need.

I have a project in mind that I could use Piccolo with, but it does have a use case that requires the use of a PostGIS geography type column. I've been looking through https://github.com/piccolo-orm/piccolo/blob/master/piccolo/columns/base.py and it doesn't look too daunting to subclass `Column` and go from there. If I can get something useable, even with using raw for selects, it would be pretty great.

Since there doesn't seem to be any documentation on implementing custom columns, I'd figure I'd post an issue, even if there isn't really a problem yet just in case anyone had any advice or thoughts.

dantownsend:
@Fingel Thanks for the kind words.

You're right - it should be possible to create custom column types without too much hassle by subclassing `Column`. I should mention this in the docs.

I don't know much about the Geography column type, but you'll probably need something like this:

```python
class Geography(Column):

    value_type = str

    def __init__(
        self,
        shape: str = 'POINT',
        number: t.Optional[int] = None,
        default: t.Union[str, t.Callable[[], str], None] = "",
        **kwargs,
    ) -> None:
        self._validate_default(default, (str, None))

        self.shape = shape
        self.number = number
        kwargs.update({"shape": shape, "number": number})
        super().__init__(**kwargs)

    @property
    def column_type(self):
        if self.number:
            return f"GEOGRAPHY({self.shape}, {self.number})"
        else:
            return f"GEOGRAPHY({self.shape})"
```

I'm interested to know how you get along. If you're able to implement PostGIS functionality, I'd like to merge it into the main piccolo repo, or have a separate piccolo_gis repo.

Fingel:
Hi @dantownsend ,

I did start our writing something that looks like your example. I wanted to make it easy on myself to I defined a `Point(Column)` which looked a lot like your example, with plans to generalize into full Geography type later.

The first issue I ran into when trying to create a migration was this: 

`module 'piccolo.columns.column_types' has no attribute 'Point'`

https://github.com/piccolo-orm/piccolo/blob/fcf58ac329114ceb269732b6a8d888f79ea137d4/piccolo/apps/migrations/auto/migration_manager.py#L192

[...]

PR #57, opened 2021-02-04
Added WhereRaw

dantownsend:
This allows raw SQL to be executed in where clauses.

It's an escape hatch, allowing database features to be accessed which are otherwise not possible when using Piccolo.

For example, GIS and JSON functions.

https://github.com/piccolo-orm/piccolo/issues/55


Fingel:
I think this is a great addition. Could it be dangerous? Yes. But when dealing with databases like Postgres, it's almost a necessity if you want to take advantage of the billions of features it provides that would be unreasonable to expect Piccolo to implement natively.

Full text search is another one of the killer features from Postgres that this feature might enable some people to take advantage of flexibly.

PR #61, opened 2021-02-05
docs: Fix mistake in docs for select

dbanty:


dantownsend:
@dbanty Could catch! Thanks

Issue #76, opened 2021-03-11
[ENHANCEMENT] Add before_create, after_create, at_update, on_delete hooks for tables(schemas).


coder3112:
## Minimal Use Case
``` python
class User(Table, tablename='my_users'):
    username = Varchar(length=100, unique=True)
    password = Secret(length=255)
    first_name = Varchar(null=True)
    last_name = Varchar(null=True)
    email = Varchar(length=255, unique=True)
    active = Boolean(default=False)
    admin = Boolean(default=False)

class Profile(Table, tablename='foo'):
   uuid = UUID()
   user = ForeignKey(references=User)
   ......
```
Here, I would want that a function ``` foo_func and foo_func2 ``` be executed when it is deleted, updated or created.

## Proposed solution
I can use a function like so:
``` python
def foo_func():
   do_something

def foo_func2():
   do_something_else

def foo_func3():
   do_something_interesting

class User(Table, tablename='my_users'):
    username = Varchar(length=100, unique=True)
    password = Secret(length=255)
    first_name = Varchar(null=True)
    last_name = Varchar(null=True)
    email = Varchar(length=255, unique=True)
    active = Boolean(default=False)
    admin = Boolean(default=False)
    
    class Meta:
         on_create = ['foo_func', 'foo_func2']
         one_update = ['foo_func2']
         on_delete = ['foo_func3']
```
Now whenever I run ``` INSERT ```, ``` UPDATE``` or ```DELETE```, these functions will run accordingly.

##### PS - I can start working on a PR if you think this should be done

dantownsend:
@coder3112 It's a nice idea. What kind of things would you run in the functions though?

coder3112:
These would fundamentally do the same things as ``` pre_save ``` & ``` post_save``` signals in django

dantownsend:
@coder3112 OK, I can see this being useful.

I guess that by having these functions, it's cleaner than overriding the `save` method for a `Table`.

I can imagine people putting some validation logic in there.

The limitation with the Django implementation is that it doesn't trigger for bulk updates. The same would have to be true here.

One thing to be aware of is Piccolo doesn't currently use a `Meta` class as a child of a `Table`, like in other ORMs. It uses a fairly modern Python feature where you can pass arguments into a class, using `__init_subclass__`. So the easiest implementation would be:

```python
class User(Table, tablename='my_users', pre_save=[some_function]):
    username = Varchar(length=100, unique=True)
    password = Secret(length=255)
[...]

PR #83, opened 2021-03-26
Added pre_query and post_query for #76 and get() for #84

coder3112:
I have added the following things for #76 :
- ``` pre_insert```
- ``` post_insert```
- ``` pre_update```
- ``` post_update```
- ``` pre_delete```
- ``` pre_delete```
All tests passed locally for both sqlite and postgres.
Will make new tests soon.

### DIFF
--- a/piccolo/query/methods/update.py
+++ b/piccolo/query/methods/update.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
-from dataclasses import dataclass
+
 import typing as t
+from dataclasses import dataclass
 
 from piccolo.custom_types import Combinable
 from piccolo.query.base import Query
--- a/piccolo/query/methods/update.py
+++ b/piccolo/query/methods/update.py
@@ -32,15 +32,27 @@ class Update(Query):
 
     def validate(self):
         if len(self.values_delegate._values) == 0:
-            raise ValueError(
-                "No values were specified to update - please use .values"
-            )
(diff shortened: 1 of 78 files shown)


coder3112:
I still need to allow for arguments and async functions. These will be coming soon

dantownsend:
@coder3112 Thanks for this. I'll try and review ASAP.

coder3112:
Ok, so I have also added the fix for #77 in this PR.
Realize its a mistake but did not realize what I was doing. (Should have created a new branch as it is separate issue)

piccolo/query/methods/update.py
```diff
@@ -32,15 +33,27 @@ def where(self, where: Combinable) -> Update:
...
         for column, value in self.values_delegate._values.items():
             if len(column._meta.call_chain) > 0:
-                raise ValueError(
-                    "Related values can't be updated via an update"
-                )
+                raise ValueError("Related values can't be updated via an update")
+
+    async def run_pre_functions(self):
+        for function in self.table._meta.pre_update:
+            function()
```
aminalaee, line 44: @coder3112 This is more of a question, shouldn't these be awaited?

coder3112: Good Catch! I just ran it and yes you are right. A couple changes need to be made so i guess I'll close this.

PR #86, opened 2021-04-18
Fix example_schema.rst

teners:
Correct `Band.manager` being a `Varchar` instead of `ForeignKey` field in the example schema documentation part


dantownsend:
@teners Good catch, thanks!

PR #122, opened 2021-06-29
Json improvements

dantownsend:
Made some improvements to how Piccolo handles JSON.

Taking this table as an example:

```python
class MyTable(Table):
    column_a = JSON()
```

Piccolo is now better at switching between JSON strings and Python objects.

```python
# The dictionary will be automatically converted to a JSON string, for storing in the database.
await MyTable.update({MyTable.column_a: {'a': 123}).run()
```

You can also tell Piccolo to deserialise JSON coming from the database.

```python
>>> await MyTable.select().output(load_json=True).run()
[{'id': 1, 'column_a': {'a': 123}]
```


heliumbrain:
Looks great!

piccolo/utils/sql_values.py
```diff
@@ -0,0 +1,27 @@
...
+from enum import Enum
+import typing as t
+
+from piccolo.utils.encoding import dump_json
+
+if t.TYPE_CHECKING:
+    from piccolo.columns import Column
+
+
+def convert_to_sql_value(value: t.Any, column: Column) -> t.Any:
```
heliumbrain, line 11: When using the newly added deserialize_json param for my pydantic models in create_pydantic_model I get the following error when trying to create objects in the database:

```python
  File "asyncpg/protocol/prepared_stmt.pyx", line 171, in asyncpg.protocol.protocol.PreparedStatementState._encode_bind_msg
    raise exceptions.DataError(
asyncpg.exceptions.DataError: invalid input for query argument $4: {} (expected str, got dict)
```

Looks like this would fix it, or is it possibly another issue?

dantownsend: Yeah, you're right - once this MR is merged in, it should fix that problem.

dantownsend: The current release of Piccolo doesn't automatically convert a dict to a string for storing the database, but this new release will.

dantownsend: @heliumbrain I just pushed a new build to PyPI (0.24.0) - hopefully should fix that issue.

heliumbrain: Yeah, seems to be working fine now 👍

PR #132, opened 2021-07-15
update-engine-docs-piccolo-conf-example

### LINKED ISSUE #102: PICCOLO_CONF environment variable clarification
I'm working on a project where I want to maintain a directory structure like
```
api
├── db
│   ├── __init__.py
│   ├── piccolo_conf.py
```
Since my `piccolo_conf.py` file is no longer at the root of the project, I need to specify the `PICCOLO_CONF` environment variable. For the sake of clarity (and I'm happy to make a PR for this), is it valuable to add an example to the [docs](https://github.com/piccolo-orm/piccolo/blob/master/docs/src/piccolo/engines/index.rst) highlighting that the required format for the path looks like this:
```api.db.piccolo_conf```
rather than this?
```api/db/piccolo_conf.py```
I kept running into errors the when using the latter format, but the former is accepted.

aminalaee:
Fixes #102 
Add example of `PICCOLO_CONF` env variable inside sub-modules.

### DIFF
--- a/docs/src/piccolo/engines/index.rst
+++ b/docs/src/piccolo/engines/index.rst
@@ -95,7 +95,10 @@ variable accordingly.
 
     DB = SQLiteEngine(path='my_test_db.sqlite')
 
-.. hint:: You can also specify sub modules, like `my_module.piccolo_conf`.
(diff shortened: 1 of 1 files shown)


dantownsend:
@aminalaee Great, thanks for this!

PR #133, opened 2021-07-16
add-scripts-folder

aminalaee:
I think it'd be clear to move all the scripts to the `scripts` folder following [this](https://github.com/github/scripts-to-rule-them-all).
I've moved all existing scripts and updated the github workflow. But here some questions/ideas:

- Should we move `piccolo.sh` from root of project too?
- Coverage is not enforced, should we add a minimum coverage so pipeline fails on missing coverage using something like Codecov?
- `release.sh` can be a workflow in github to upload to PyPi when github release is created.


dantownsend:
@aminalaee This is much cleaner, thanks!

I've moved `piccolo.sh` into the scripts folder.

I used to have Coveralls configured when I was using Travis for CI. I couldn't easily get it working on GitHub Actions, and didn't have time to investigate further. If we could get it working, that would be great.

Likewise, with releasing to PyPI - I didn't have time to investigate how to store secrets with GitHub Actions. I moved over in a hurry when Travis CI became unbearably slow!

aminalaee:
@dantownsend Glad to be of help.
I'll give it a try tomorrow and let yoy know how it goes.

aminalaee:
@dantownsend

- I removed `coverage-report.sh` and added a `tests.sh` script to be used in `test-postgres.sh` and `test-sqlite.sh` to run tests, check minimum coverage and generate reports (avoiding duplication).
- I've added the `release.yaml`, you need to get PyPI API token and set up `PYPI_TOKEN ` secret for the project or organization.
- If you don't mind I added `codecov`. The results are not uplaoded. You need to set `CODECOV_TOKEN` secret for the project or organization

dantownsend:
@aminalaee Thanks for all your hard work with this - it's a big improvement.

With isort and black, I heard they're sometimes incompatible. Have you ever experienced this?

aminalaee:
@dantownsend To be honest I don't have that much experience but yes they can be incompatible. My experience was that `black` prefers imports like this in `__init__` :
```
from .base import (
    A,
    B,
)
```

But `isort` wants to do that `from .base import A, B`.
Based on my experience `isort` can do the sorting but `black` can reformat the imports.  Looking at some other repos putting black after isort should be fine.

aminalaee:
@dantownsend Added `isort` profile for `black`. The profile makes them pretty compatible. The only difference is line length, which `isort` and `flake8` use 79 by default, but `black` uses 88. Since `piccolo` was already using 79, I set 79 explicitly for `isort`, `black` and `flake8`.

dantownsend:
@aminalaee Thanks again for all this. I've added secrets for Codecov and PyPI, so hopefully they should work as expected.

Issue #63, opened 2021-02-05
Confusion on primary key behavior


dbanty:
It seems like even though `primary` is an available argument to a `Column`, using it messes with migrations, preventing auto from working properly. Through some trial and error, I seem to have found that an `id` column is automatically added as the primary key to models. This wasn't at all clear to me in reading through the documentation.

I would like the ability to declare my own primary key column called `pk`, as `id` is a builtin function in Python and shadowing it (e.g. in a param to a function) can cause accidental mistakes. At the very least, it would be good to make this default behavior more obvious in the documentation.

dashsatish:
@dbanty Please do have a look at https://github.com/piccolo-orm/piccolo/issues/32. I tripped on the same kinda issue.

dantownsend:
You're both right - I didn't realise this functionality was completely undocumented.

I've added docs for now, but hope to make it more configurable at some point.

aminalaee:
@dantownsend Can this be closed?

dantownsend:
@aminalaee Yes, in v0.29.0 a custom primary key column can now be specified.

Issue #32, opened 2021-01-04
[Need a suggestion/Fix] "id" field seems to be too tied to integer type 


dashsatish:
Hi,
Can the `id` field be allowed to be redefined in the child table/model with a different type. 
- Need it to be UUID which is also supported in postgres? But piccolo doesn't allow it to be changed. So all I had to do was use my own field with a `unique` constraint. 
- Moreover, can we not have composite primary keys? is there a way to define?

dantownsend:
@dashsatish This is a tricky change, which I'll need to think carefully about. I've added it to the roadmap, and will see if I can build a proof of concept. For now, having a separate UUID column is the best workaround.

aminalaee:
@dantownsend Can this be closed?

dantownsend:
@aminalaee yep

PR #136, opened 2021-07-20
support-custom-field-primary-key

### LINKED ISSUE #107: Allow using uuid instead of integer for the id column
Right now piccolo admin returns a weird ID when uuid is used as a primary key

4C6CC5B0-40D8-4AF6-804C-F1B6567D39F7

aminalaee:
Fixes #107 #32 and maybe #63 
Here's my approach to support UUID priamry key:

- Removed `PrimaryKey` column as it only supported `Integer` value types
- Didn't define `id` for `Table` by default and checked if no `PrimaryKey` was defined for table then `Serial` PK would be added.
- I think `Serial` column is Postgres specific, maybe change it to something like Django `AutoField`?

This works fine in Postgres, but in `sqlite` after insert we're returning `cursor.lastrowid` which does not return our `id` but the special virtual column `ROWID` from sqlite. I could come up with two approaches:
- Sqlite has added `RETURNING` clauses so we can do some sane INSERT query the same way as Postgres. But we will only support Sqlite 3.35 forwards: [here](https://www.sqlite.org/releaselog/3_35_0.html)
- Or we have to get the `cursor.lastrowid` and do another query to get our id based on ROWID like this `SELECT id FROM table WHERE ROWID = <cursor.lastrowid>`.
I wanted to get some feedback on this.

Any of these should work now:

```
class MyTableDefaultPrimaryKey(Table):
    name = Varchar()


class MyTablePrimaryKeyInteger(Table):
    id = Integer(null=False, primary=True, key=True)
    name = Varchar()


class MyTablePrimaryKeyUUID(Table):
    id = UUID(null=False, primary=True, key=True)
    name = Varchar()
```


piccolo/engine/sqlite.py
```diff
@@ -427,12 +427,14 @@ async def _run_in_new_connection(
             connection.row_factory = dict_factory  # type: ignore
             async with connection.execute(query, args) as cursor:
                 await connection.commit()
-                response = await cursor.fetchall()
 
                 if query_type == "insert":
+                    # await cursor.execute(f"SELECT id from my_table_primary_key_uuid WHERE rowid = {cursor.lastrowid};")
+                    # response = await cursor.fetchone()
+                    # return [{"id": response["id"]}]
                     return [{"id": cursor.lastrowid}]
```
aminalaee, line 435: This is where we're getting `ROWID` of last insert instead of `id`

dantownsend:
@aminalaee This looks great so far.

Good catch with the `lastrowid` issue. Most of the time the value should be the same as the `id` column, but I guess there could be times when the values are different.

The SQLite version can be detected using `aiosqlite.sqlite_version`. So `RETURNING` could be used for newer versions, and an additional `SELECT` query for older versions.

I can understand why you removed the `PrimaryKey` column type, but it might be worth keeping it for now with a deprecation warning, to preserve backwards compatibility.

Just a reminder to myself to update the docs here too, because it currently says custom primary key columns can cause unexpected behaviour:

https://piccolo-orm.readthedocs.io/en/latest/piccolo/schema/defining.html?highlight=PrimaryKey#default-columns

Many thanks for this!

aminalaee:
@dantownsend I looked at sqlite version more, according to [this](https://bugs.python.org/issue43492) python thread, support for sqlite 3.35.X was added to Python and checking latest releases of Python I noticed it was only added in 3.9.
Again I checked that release of Python is not supported on many OS including mine Ubuntu 20.04.
I guess it'd be ok to leave a Todo or Task for it to pick it up later and do query to to get id for now.

dantownsend:
@aminalaee Yeah, it can be left for now.

Have you seen the dead snakes Python distributions for Ubuntu? You might be able to install Python 3.9 using that, if you need it.

https://launchpad.net/~deadsnakes/+archive/ubuntu/ppa

aminalaee:
@dantownsend That's cool. I hadn't seen that. Thanks

piccolo/engine/base.py
```diff
@@ -41,6 +42,10 @@ async def prep_database(self):
     async def batch(self, query: Query, batch_size: int = 100) -> Batch:
         pass
 
+    @abstractmethod
+    async def run_querystring(self, query_string: QueryString, in_pool: bool):
```
dantownsend, line 46: I think there's a minor typo here - should be `querystring`.

piccolo/engine/sqlite.py
```diff
@@ -10,6 +10,7 @@
 from decimal import Decimal
 
 from aiosqlite import Connection, Cursor, connect
+from packaging import version
```
dantownsend, line 13: Where does this dependency come from? Presumably it's a requirement from another dependency. Do you think it's worth adding it explicitly to requirements.txt, in case the third party removes it as a dependency?

piccolo/engine/sqlite.py
```diff
@@ -418,28 +419,49 @@ async def get_connection(self) -> Connection:
...
+        `ROWID` and `id` will return different types.
+        Need to query by `lastrowid` to get `id`s in SQLite prior to 3.35.0.
+        """
+        # TODO: Add RETURNING clause for > 3.35.0
+        if version.parse(sqlite3.sqlite_version) < version.parse("3.35.0"):
+            await cursor.execute(
+                f"SELECT id FROM {tablename} WHERE ROWID = {cursor.lastrowid}"
+            )
+            response = await cursor.fetchone()
+            return response["id"]
```
dantownsend, line 436: We'll have to be careful here. There are two options to prevent any Python 3.9 breakage when merging:

1. Remove the `if` statement
2. Add the RETURNING clause

piccolo/table.py
```diff
@@ -212,6 +211,14 @@ def __init_subclass__(
                 if isinstance(column, (JSON, JSONB)):
                     json_columns.append(column)
 
+        if not primary_key_column:
+            primary_key_column = cls._create_serial_primary_key()
+
+            columns.insert(0, primary_key_column)  # PK should be added first
+            default_columns.append(primary_key_column)
+
+        cls.id = primary_key_column
```
dantownsend, line 220: There could be a subtle issue here. Imagine someone defines a primary key which isn't called `id` (for example `pk`). We would then end up with two primary key columns (`id` and `pk`), which would fail.

A nice solution would be to add a new attribute to `TableMeta` called `primary_key`. In the example above, we would assign the `pk` column to `TableMeta.primary_key`.  So rather than us always assuming the primary key is called `id`, we can access it by doing `SomeTable._meta.primary_key`.

We would effectively be killing two birds with one stone - the user can define the column type used as the primary key, and also the name of the column.

I don't think there are two many places which assume the primary key is called 'id'. Here are some example:

* https://github.com/piccolo-orm/piccolo/blob/master/piccolo/utils/sql_values.py#L21
* https://github.com/piccolo-orm/piccolo/blob/master/piccolo/table.py#L319
* https://github.com/piccolo-orm/piccolo/blob/master/piccolo/table.py#L334
* https://github.com/piccolo-orm/piccolo/blob/master/piccolo/table.py#L376
* https://github.com/piccolo-orm/piccolo/blob/master/piccolo/table.py#L415

What do you think? I can help implement these changes.

tests/apps/migrations/commands/test_forwards_backwards.py
```diff
@@ -174,6 +174,8 @@ def test_forwards_fake(self):
                 "2020-12-17T18:44:30",
                 "2020-12-17T18:44:39",
                 "2020-12-17T18:44:44",
+                "2021-07-19T12:40:31:140762",
+                "2021-07-19T13:34:56:395354",
```
dantownsend, line 178: Presumably these new migrations are to test creating tables with the new metaclass behaviour?

dantownsend:
@aminalaee This is a great bit of work. I've reviewed it in detail, and have left comments for anything I can think of.

I'm happy to help with making these tweaks if you want.

piccolo/engine/sqlite.py
aminalaee: it's a dependency installed by `setuptools`. I guess it wouldn't harm to add it as dependency but that'd be very odd if that was removed `setuptools` since they're for the same ogranization.

piccolo/engine/sqlite.py
aminalaee: I agree with removing the `if` and keeping the comment for future referrences.

piccolo/table.py
aminalaee: I love that idea, I thought about that too. Let me give it a go tomorrow and if I see if it's too much I'll leave that to you.

tests/apps/migrations/commands/test_forwards_backwards.py
aminalaee: Shamefully I don't know why they were automatically generated when I ran the tests. I guess because I added new test tables?

dantownsend: No worries - they were probably just added by mistake when a file got moved. I've removed them, and added a new migration file, as I just realised some of the tables in the example app weren't covered by migrations.

piccolo/table.py
aminalaee: @dantownsend I've started this change but a dozen of tests are failing, mainly with `sqlite3.InterfaceError: Error binding parameter 0 - probably unsupported type.`
I've also added a test to define `pk` as `primary key` on table, it's passing but I have a feeling it's not right, it inserts 0 in `Serial`.
Any help with this would be great.

dantownsend: I think the  `sqlite3.InterfaceError: Error binding parameter 0 - probably unsupported type.` error is being caused by this:

```python
.where(MyTable._meta.primary_key == row._meta.primary_key)
```

The problem is `row._meta.primary_key`. It's returning the column instance (e.g. Sequence()), rather than the column value (e.g. 1).

It's a bit untidy, but this works:

```python
.where(MyTable._meta.primary_key == getattr(row, MyTable._meta.primary_key._meta.name)
```

aminalaee:
@dantownsend Nice, those were a bit tricky.

dantownsend:
@aminalaee Yeah, it's a tricky change. I think it's almost there now - I'm just going to test it with a bunch of different Piccolo apps to see if there are any edge cases we haven't thought of.

dantownsend:
@aminalaee I found a couple of edge cases, but I think that's all. They're fixed now, but need unit tests.

I'm excited to start using this new feature.

aminalaee:
@dantownsend I really like that too. I plan to use it in my next project. I guess the docs need to be updated too.

dantownsend:
@aminalaee I was writing some more unit tests, and realised there was one thing we overlooked.

When a foreign key is created, it has to have the same type as the primary key it references.

For example, with this:

```python
class Manager(Table):
    pk = UUID(primary=True, key=True)
    name = Varchar()


class Band(Table):
    pk = UUID(primary=True, key=True)
    name = Varchar()
    manager = ForeignKey(Manager)
        
```

The `Band.manager` foreign key needs to be of type `UUID`, so it can store a reference to `Manager.pk`. Before, foreign key columns were only ever created as integers.

I think I've fixed it, but will test it a bit more tomorrow.

piccolo/columns/base.py
```diff
@@ -318,6 +318,9 @@ def __init__(
             }
         )
 
+        if (primary is True) and (key is False):
```
aminalaee, line 321: Is there a case where `key` can be True and `primary` False? If they always should match then maybe we can turn them into a single field?

aminalaee:
@dantownsend That was a great find. I noticed we might be doubling the `primary key` tests. There's one in `columns/test_primary.py` and another `table/test_primary_key.py`. They have a bit of overlapping but maybe we can do it all in `columns`?

piccolo/columns/base.py
dantownsend: You're right, it would make more sense if there was just a `primary_key` argument, rather than `primary` and `key`. Honestly, I'm not sure why I did it like this originally.

I'll try changing it, but will have to tread carefully in case it effects migrations.

dantownsend:
I discovered a few quirks when using this with migrations - will try and fix them today.

aminalaee:
@dantownsend If there's something I can do, I'll be glad to pick that up.

dantownsend:
@aminalaee  Thanks - if you don't mind having a look, that would be good. It's a bit of a tricky one.

Basically, when you create a table with a custom primary key:

```python
class MyTable(Table):
    pk = UUID(primary_key=True)
```

When you create a migration for it, and run it:

```bash
piccolo migrations new my_app --auto
piccolo migrations forwards my_app
```

You'll notice that the custom primary key column doesn't get created.

The reason for this, is when the migrations are run, Piccolo first creates the table, then adds the columns:

https://github.com/piccolo-orm/piccolo/blob/982c6d0bf2f9cccdde7002afccd79c46e4ed281b/piccolo/apps/migrations/auto/migration_manager.py#L647-L649

So what you end up with is the default `id` column, which is a `Serial`, which is created as part of `_run_add_tables`. And the custom `pk` column doesn't get created, because it's added to `Table._meta.default_columns` and so `_run_add_columns` ignores it.

It just needs tweaking a bit, for example:

1. The migrations can be told to create an empty table, without the `id` column.
2. The `pk` column is added to `Table._meta.default_columns` instead of `Table._meta.columns`, which means it'll get added along with the other custom columns.

Or some other better solution.

aminalaee:
@dantownsend I thought about changing the `Create` method but I have a feeling that would break other places because we don't usually create tables without columns, it's only the case in migrations, is that right?
Maybe we can change this behavior and create table with columns in migrations?

dantownsend:
@aminalaee Yes, your approach makes more sense (creating the table and columns at the same time).

The columns for a given table can be gotten in the `MigrationManager` using `self.add_columns.for_table_class_name` . A `Table` can then be constructed using `create_table_class` in `tables.py`, and we just need to call `create_table` on it.

aminalaee:
@dantownsend Sorry this is taking forever to finish.
I tried what I proposed but it was difficult since we added columns in `_run_add_tables` and in the next step `_run_add_columns` they would be added again and that was an issue. I did create an empty table with no columns and then added columns next.
But I needed to change something, in `tables.py` I needed to add `pk` to `non_default_columns`. So now we're not using `default_columns` and `columns` is the same as `non_default_columns` so I think if there's no other use they can be removed?

dantownsend:
@aminalaee One of the reasons there's `default_columns` is for tooling which is outside of the main Piccolo repo. The main thing which uses it is `create_pydantic_model` in Piccolo API, which as the name suggests lets you create a Pydantic model from a Piccolo table:

https://github.com/piccolo-orm/piccolo_api/blob/019e3ae00564a8ffb8cd04ee539515c73bcaf89a/piccolo_api/crud/serializers.py#L40

When I'm building web apps, I usually want to omit the primary key column from the serialiser, so someone can't try and patch it in a HTTP request.

So we need to keep the `default_columns` attribute if possible.

As you mentioned, we need to make sure we don't then try and add the columns twice when running the migration. In `_run_add_columns` we can add an if statement like:

```python
if table_class_name in [i.table_class_name for i in self.add_tables]:
    continue
```

Underneath here:

https://github.com/piccolo-orm/piccolo/blob/982c6d0bf2f9cccdde7002afccd79c46e4ed281b/piccolo/apps/migrations/auto/migration_manager.py#L603

Hope that helps.

aminalaee:
@dantownsend I've fixed all tests except one `tests/apps/migrations/commands/test_new.py`. I think our latest change generated a new migration for `id` column of existing tables and if I don't add that migration this test fails. Otherwise many other `forwards/backwards` fail.
I guess our change isn't backward compatible and we have to delete and create migrations for `example_app` in tests. Is that so?

piccolo/table.py
```diff
@@ -212,11 +209,19 @@ def __init_subclass__(
                 if isinstance(column, (JSON, JSONB)):
                     json_columns.append(column)
 
+        if not primary_key:
+            primary_key = cls._create_serial_primary_key()
+            setattr(cls, "id", primary_key)
+
+            columns.insert(0, primary_key)  # PK should be added first
+            non_default_columns.append(primary_key)
```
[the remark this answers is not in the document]
aminalaee: @dantownsend Thanks. That works.

dantownsend:
@aminalaee I'm going to merge it in now and release it.

Piccolo admin needs a few tweaks to work with this change, but otherwise seems to work well.

Thanks for your hard work on this.

aminalaee:
@dantownsend I'll try and update the Admin if I can.
Thank you for the help.

PR #145, opened 2021-08-01
add uvloop optional dependency

### LINKED ISSUE #143: Configuring event loop implementation
As far as I know there's no way to set up any event loop other than Python's `asyncio`. And to my understanding if using `uvloop` performance will be improved.

I suggest we add `uvloop` as an extra dependency and maybe do one of the followings:
- Try using `uvloop` as default and if not available fallback to `asyncio`
- Pass the event loop to the engine confugarion
Or any other ways you can suggest.

aminalaee:
Closes #143 

- The docs can be improved in clarifying the optional dependencies.
- Maybe instead of ignoring `uvloop` for mypy, we can install `mypy --install-types uvloop` or for all packages? This needs upgrading mypy to have this option [here](https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-install-types).


dantownsend:
@aminalaee  This is great, thanks.

> Maybe instead of ignoring uvloop for mypy, we can install mypy --install-types uvloop or for all packages? This needs upgrading mypy to have this option here

Yes, I have no issue with upgrading MyPy to support this new feature.

requirements/extras/uvloop.txt
```diff
@@ -0,0 +1 @@
+uvloop==0.15.3
```
dantownsend, line 1: We might want to loosen the versioning on this (`uvloop>=0.15.3`) - or perhaps removing it all together so it's just `uvloop`.

Otherwise it could cause issues when installing it with other libraries which are dependent on uvloop.

I imagine the uvloop API will stay pretty stable, as far as we need it, so should be safe enough.

setup.py
```diff
@@ -9,16 +9,37 @@
...
 
 
+def parse_requirement(req_path):
+    """Parse requirement file.
+    Example:
+        parse_requirement('requirements.txt')       # requirements/requirements.txt
+        parse_requirement('extras/playground.txt')  # requirements/extras/playground.txt
+    Returns:
+        List[str]: list of requirements specified in the file.
+    """
```
dantownsend, line 26: Should we move the type annotations to the function body?

```python
def parse_requirement(req_path: str) -> List[str]:
```

In Piccolo the docstrings tend to start with a newline character - it might be worth changing this to make it consistent? Not a big deal though.

```python
"""
Parse requirement file.
```

requirements/extras/uvloop.txt
aminalaee: Thanks. That's a good point. I did a quick search and it seems `uvloop.install()` was introduced in 0.12.0 according to [this](https://github.com/MagicStack/uvloop/commit/a3d8d4011fe432d7c83b80ecac8e678257b384f0#diff-62ea7a112b3168cba4028fb6d7d3f8763d426e4fbb96bca64396541b6be1af92)

dantownsend:
@aminalaee This looks good to me - if there's nothing else you want to add to it, I'll merge it in.

aminalaee:
@dantownsend Thanks. I have nothing else to add.

PR #146, opened 2021-08-03
basic aggregate functions

sinisaos:
Based on question [#141](https://github.com/piccolo-orm/piccolo/discussions/141) in Discussions.


tests/table/test_aggregate.py
```diff
@@ -0,0 +1,54 @@
...
+from ..example_app.tables import Band
+
+
+class TestAggregate(DBTestCase):
+    def test_avg(self):
+        self.insert_rows()
+
+        response = Band.select(Avg(Band.popularity)).first().run_sync()
+
+        self.assertTrue(float(response["avg"]) == 1003.3333333333334)
```
aminalaee, line 13: This is great. I think it can also be useful to allow `as_alias` for aggregate functions the same way as `Column` does. What do you think?

dantownsend:
@sinisaos Thanks for this - looks great.

piccolo/query/methods/aggregate.py
```diff
@@ -0,0 +1,87 @@
...
+    """
+
+    def __init__(self, column=t.Type[Column]):
+        self.column = column
+
+    def get_select_string(self, engine_type: str, just_alias=False) -> str:
+        if self.column.column_type.split("(")[0] in NUMERIC_TYPES:
+            column_name = self.column._meta.get_full_name(
+                just_alias=just_alias
+            )
```
dantownsend, line 30: An alternative to this is to check the column's `value_type`.

```python
if self.column.value_type in (int, decimal.Decimal, float):
    column_name = self.column._meta.get_full_name(
        just_alias=just_alias
    )
```

It's slightly more flexible, as this way people can define their own column types and it will still work. What do you think?

It's probably worth moving the logic into a separate method called something like `is_number_column`, as it reduces the code repetition, and makes unit testing easier.

```python
def is_number_column(column: Column) -> bool:
    return column.value_type in (int, decimal.Decimal, float)
```

dantownsend: Also, if we move the check to the `__init__` method it might be better, as otherwise it will just detect the problem when the query runs. What do you think?

dantownsend (commented):
Just a few comments.

We'll have to remember to add docs for these.

As @aminalaee mentions, I think having an `as_alias` method would be useful.

Something roughly like this?

```python
class Max(Selectable):
    """
    MAX() SQL function.

    await Band.select(Max(Band.popularity)).run()
    """

    def __init__(self, column=t.Type[Column], alias: str='max'):
        self.column = column
        self.alias = alias
    
    def as_alias(self, alias: str):
        self.alias = alias

    def get_select_string(self, engine_type: str, just_alias=False) -> str:
        column_name = self.column._meta.get_full_name(just_alias=just_alias)
        return f"MAX({column_name}) AS {alias}"
```
Also, I was thinking about where best to store these aggregate classes. In `select.py` we have `Count`, so I was thinking whether the aggregate classes should also be in `select.py` or if we should move `Count` into `aggregate.py`? What do you think?

piccolo/query/methods/aggregate.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+class Min(Selectable):
+    """
+    MIN() SQL function.
+
+    await Band.select(Min(Band.popularity)).run()
+    """
+
+    def __init__(self, column=t.Type[Column]):
```
dantownsend, line 60: There are minor typos here in the `__init__` method. Should be `def __init__(self, column: t.Type[Column]):`.

tests/table/test_aggregate.py
sinisaos: @aminalaee Thanks for great suggestion.

piccolo/query/methods/aggregate.py
sinisaos: @dantownsend Thank you. These are great suggestions, I will try to implement it and write more tests.

piccolo/query/methods/aggregate.py
sinisaos: @dantownsend Sorry about that. I'll fix that.

sinisaos:
@dantownsend and @aminalaee Thank you both for your ideas and code samples. I'll do my best to fix this.

sinisaos:
@dantownsend I made some changes as suggested:

- Made `alias` for aggregate functions. The query looks like this `await Band.select(Avg(Band.popularity, alias="popularity_avg")).run()`
- The code has been moved from `aggregate.py` to `select.py` because it is a more natural place to be
- Aggregate tests have been moved to `test_select.py` and add test for aliases

Hope this looks much better. Cheers

dantownsend (commented):
@sinisaos It's looking good. I just have some minor suggestions.

 * Using `ValueError` instead of `Exception`
 * I think it's worth also adding an `as_alias` method, as it's used elsewhere in the codebase for creating aliases (even though it seems redundant, as you can pass the alias into `__init__`).

sinisaos:
@dantownsend I changed the code based on your suggestions and added docs for new features. What do you think?

piccolo/query/methods/select.py
```diff
@@ -46,6 +72,71 @@ def get_select_string(self, engine_type: str, just_alias=False) -> str:
...
+
+    await Band.select(Min(Band.popularity)).run() or with aliases
+    await Band.select(Min(Band.popularity, alias="popularity_min")).run()
+    """
+
+    def __init__(self, column: Column, alias: str = "min"):
+        self.column = column
+        self.alias = alias
+
+    def as_alias(self, alias: str):  # pragma: no cover
```
aminalaee, line 107: @sinisaos Sorry to jump in again, but I think you can add unit tests for the `as_alias` method so you don't have to exclude them from test coverage using `#  pragma: no cover`

dantownsend:
@sinisaos Thanks for making those changes, and adding the docs. What do you think about moving the content in `aggregate_functions.rst` into `select.rst`. It could go in a section above `Query clauses`?

piccolo/query/methods/select.py
sinisaos: @aminalaee I agree with you. I changed the code to actually use `as_alias` method and the test now don't need `# pragma: no cover`.

```python
class Sum(Selectable):
    """
    SUM() SQL function. Column type must be numeric to run the query.

    await Band.select(Sum(Band.popularity)).run() or with aliases
    await Band.select(Sum(Band.popularity).as_alias("popularity_sum")).run()
    """

    def __init__(self, column: Column, alias: str = "sum"):
        self.column = column
        self.alias = alias

    def as_alias(self, alias: str) -> Sum:
        column = copy.deepcopy(self)
        column.alias = alias
        return column

    def get_select_string(self, engine_type: str, just_alias=False) -> str:
        if self.column.value_type in (int, decimal.Decimal, float):
            column_name = self.column._meta.get_full_name(
                just_alias=just_alias
            )
        else:
            raise ValueError("Column type must be numeric to run the query.")
        return f"SUM({column_name}) AS {self.alias}"
```
The query with aliases is now like this `await Band.select(Sum(Band.popularity).as_alias("popularity_sum")).run()`.  What do you think? Is this better or are we going back to the previous code?

sinisaos:
@dantownsend I agree with you. I can move content in `aggregate_functions.rst` to `select.rst`. But I think that we should leave this in `Query Types` section because `Select` is already there. Can you please check my last comment to @aminalaee suggestion and new code for `as_alias` method. The query with aliases might look like this `await Band.select(Sum(Band.popularity).as_alias("popularity_sum")).run()`. Is it better or do we leave it as it is?

piccolo/query/methods/select.py
dantownsend: @sinisaos The `as_alias` method can be this (no need to do the copy):

```python
    def as_alias(self, alias: str) -> Sum:
        self.alias = alias
        return self
```

I think it's worth having the `as_alias` because that's how it works with other parts of the Piccolo code base, for example `await User.select(User.name.as_alias('title')).run()`.

I expect most people will use `Sum(MyTable.column, alias="foo")`, but having `as_alias` is a nice to have, to keep things consistent.

sinisaos: @dantownsend That's great. Thanks. Now we can make query in two ways `await Band.select(Sum(Band.popularity, alias="popularity_sum")).run()` and `await Band.select(Sum(Band.popularity).as_alias("popularity_sum")).run()` and coverage is fine. I'll add this to docs.

sinisaos:
@dantownsend I updated the docs. If this is not what you mean, feel free to fix them.

dantownsend:
@sinisaos Cool, thanks for this!

sinisaos:
@dantownsend  No problem. Cheers.

PR #147, opened 2021-08-03
LazyLoad database engines

### LINKED ISSUE #137: marking database engines as extras in setup.py
It would be handy to only install the specific drivers for use and not install unused ones:

- `pip install piccolo` install core piccolo
- `pip install piccolo[postgres]` install piccolo + asyncpg
- `pip install piccolo[sqlite]` install piccolo + aiosqlite
- ...

Right now we make use of `extra_requires` in `setup.py`. I noticed `celery` has done the following project structure which is clean:

```
src/
tests/
requirements/
  extras/
    postgres.txt
    sqlite.txt
  dev-requirements.txt
  requirements.txt
  test-requirements.txt
```

This should be easy. But importing required engines at runtime might need some further investigation.

aminalaee:
Fixes #137 

Right now piccolo installs both `asyncpg` and `aiosqlite`. This will make each database engine an optional dependency.

~~Running `pip install piccolo` wouldn't install any drivers. Maybe install a default database driver with `piccolo`? Or somehow specify that in README or docs so a driver has to be installed?~~

- [x] Install database engines as extras
- [x] LazyLoad database engines
- [x] Throw proper errors if using a database engine and driver is not installed
- [x] Update the docs to reflect this new change

Any feedback is appreciated.

@dantownsend Any feedback on this? My only concern was performance, which I've tested will only lazy load module once so we don't have to worry about that.


dantownsend:
@aminalaee I think it's a good idea.

Most other frameworks / ORMs make you install the drivers separately.

My main concern with removing asyncpg as a default install is it's a backwards incompatible change - if a project just has `piccolo` in it's requirements.txt file, it will break when installed. As Piccolo is still pre 1.0 it's probably OK though.

aminalaee:
@dantownsend I agree.
I also think it's a nice feature to allow multiple drivers, maybe keep it in the roadmap as low priority.

piccolo/utils/lazy_loader.py
```diff
@@ -0,0 +1,60 @@
...
+    """
+
+    # The lint error here is incorrect.
+    def __init__(
+        self, local_name, parent_module_globals, name
+    ):  # pylint: disable=super-on-old-class
+        self._local_name = local_name
+        self._parent_module_globals = parent_module_globals
+
+        super(LazyLoader, self).__init__(name)
```
dantownsend, line 23: @aminalaee Do we have to use the old style `super` here, or can we just use `super().__init__(name)`, and drop the `pylint: disable=super-on-old-class` comment?

aminalaee: @dantownsend You're right. That works. I just go that from the link in code I didn't give it much thought. Thanks

aminalaee:
@dantownsend Please consider squashing this as it has multiple commits.

aminalaee:
We can also add Codecov coverage badge if you like, but need to get token from it.

dantownsend:
@aminalaee Thanks for this - looks good to go. Anything else you think needs adding? If not, I'll merge it.

aminalaee:
@dantownsend Thanks. I think it's ready.

Issue #138, opened 2021-07-21
Modify the Blacksheep template to mount Piccolo admin


dantownsend:
When I created the BlackSheep template, there wasn't support for mounting ASGI apps, such as the Piccolo admin. This is no longer the case, so requires a tweak to the template.

PR #151, opened 2021-08-05
tweak blacksheep template

sinisaos:
Based on #138


dantownsend:
@sinisaos Thanks for this, it looks good.

Closes https://github.com/piccolo-orm/piccolo/issues/138

PR #154, opened 2021-08-06
implement abstract tables

### LINKED ISSUE #153: Feature: Abstract Tables
Implementing abstract tables to avoid repeating columns for the tables. Obviously the abstract tables are ignored in automatic migrations. It would be something like this:

```
class Base(Table, abstract=True):
    created_at = Date()


class Band(Base):
    name = Varchar()
```

aminalaee:
Closes #153

### DIFF
--- a/docs/src/piccolo/schema/advanced.rst
+++ b/docs/src/piccolo/schema/advanced.rst
@@ -86,6 +86,28 @@ use mixins to reduce the amount of repetition.
 
 -------------------------------------------------------------------------------
 
+Abstract
+--------
+
+Abstract tables also help to reduce the amount of repetition. The difference is
+that abstract tables are ignored in migrations so they're not created but
+mixin tables are created.
+
+.. code-block:: python
+
+    from piccolo.columns import Varchar, Boolean
+    from piccolo.table import Table
+
+
+    class Base(Table, abstract=True):
+        favourite = Boolean(default=False)
+
+
+    class Manager(Base):
+        name = Varchar()
+
+-------------------------------------------------------------------------------
+
 Choices
 -------
 
--- a/piccolo/apps/user/commands/change_permissions.py
+++ b/piccolo/apps/user/commands/change_permissions.py
@@ -3,7 +3,7 @@ import typing as t
 from piccolo.apps.user.tables import BaseUser
 from piccolo.utils.warnings import Level, colored_string
 
-if t.TYPE_CHECKING:
+if t.TYPE_CHECKING:  # pragma: no cover
     from piccolo.columns import Column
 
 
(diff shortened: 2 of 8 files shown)


aminalaee:
The issue is caused by Abstract class having an `id` assigned and child class would detect that as new column

dantownsend:
@aminalaee What happens if we change this line:

https://github.com/piccolo-orm/piccolo/blob/96965cad4bd9d2ba6a9ccd334b996d83d9577373/piccolo/apps/migrations/commands/new.py#L164

So instead it's:

```python
columns=i._meta.columns,
```

aminalaee:
@dantownsend  I thtink that helps. Let me create another PR for it track it there.

dantownsend:
An alternative to the abstract tables is to use mixins:

https://piccolo-orm.readthedocs.io/en/latest/piccolo/schema/advanced.html#mixins

The difference is the mixin doesn't inherit from `Table`:

```python
class FavouriteMixin:
    favourite = Boolean(default=False)


class Manager(FavouriteMixin, Table):
    name = Varchar()
```

I wonder if there's enough distinction between abstract tables and mixins. What do you think?

aminalaee:
@dantownsend Yeah I  think that's the same idea. Maybe we can close this off.

dantownsend:
@aminalaee I think abstract tables are nicer when you want to add custom methods. For example:

```python
class MyBase(Table):
    created_on = Timestamp()

    @classmethod
    def get_latest(cls):
        return cls.select().order_by(cls.created_on, ascending=False).first()
```

As it inherits from `Table`, there won't be any type warnings. With the current mixin approach, you can't do that without getting a type error.

```python
class MyBase():
    created_on = Timestamp()

    @classmethod
    def get_latest(cls):
        # Warning! Has no method called select ...
        return cls.select().order_by(cls.created_on, ascending=False).first()
```

So the abstract approach does have some merits, but we should be explicit in the docs, so it's clear why you would use this vs a mixin.

As an aside, you can kind of hack abstract tables at the moment, using tags.

```python
class MyBase(Table, tags=['abstract']):
    created_on = Timestamp()

    @classmethod
    def get_latest(cls):
        return cls.select().order_by(cls.created_on, ascending=False).first()
```

And using `table_finder` in `AppConfig` in `piccolo_app.py`.

```python
APP_CONFIG = AppConfig(
    app_name='test',
    table_classes=table_finder('my.module', exclude_tags=['abstract'],
)
```

In summary ... I do think what you've done is a good idea. If we can get it to work smoothly with migrations without any large refactors, we should do it. Otherwise we can park it for now.

aminalaee:
@dantownsend I agree there's no easy way to make this work smoothly with migrations.
Maybe we can close this and add your workaround to the docs?
Although I think the migrations will need some fixing as it will limit the functionalities we can have.

PR #156, opened 2021-08-06
fix-primary-key-migrations

aminalaee:
Migration files did not have auto `id` column in previous migrations. As far as I saw, having custom `PrimaryKey` couldn't be a backwards compatible change, so I had to remove all old migrations and create new ones.

Also adding tests to make sure migrations work for both auto and non-auto primary key fields.


tests/example_app/piccolo_migrations/2021-08-06T15-40-44-097377.py
```diff
@@ -0,0 +1,194 @@
...
+async def forwards():
+    manager = MigrationManager(migration_id=ID, app_name="example_app")
+
+    manager.add_table("Studio", tablename="studio")
+
+    manager.add_column(
+        table_class_name="Studio",
+        tablename="studio",
+        column_name="id",
+        column_class_name="Serial",
```
aminalaee, line 19: This auto `id` should not be added

aminalaee: There was an issue with the test itself

aminalaee:
@dantownsend Sorry the migrations are yet new for me.
Isn't this failure odd? Forward migrations fail with duplicate column.

aminalaee:
@dantownsend If you decide to merge this through please consider squashing it as some of my commit messages can be confusing.

brnosouza:
@aminalaee I may have found a different bug regarding the custom primary key feature: migrations doesn't drop old PK if they are different, for example, if the old one is `Serial` and second is `UUID`, also it doesn't trigger an alter FK command to replace the old column in the FK(s).

aminalaee:
@brnosouza Thank you. I'll take a look into it.

dantownsend:
@brnosouza @aminalaee Yeah, this will be a tricky thing to solve.

The question is what to do with existing foreign keys which reference the table. If the primary key is now a UUID, and the foreign keys referencing it are integers, they won't be able to store a UUID.

brnosouza:
@aminalaee I checked and django doesn't trigger a alter FK, but I think we should if the primary key name was updated. Maybe adding this information to the docs saying that you manually need to update the FK and maybe provide a function inside MigrationManager?

aminalaee:
@brnosouza I think the rename can be implemented, but type conversion can be very complicated and risky.

@dantownsend What do you think?

brnosouza:
@aminalaee I agree regarding the type conversion and, what I mean by providing a function is to not apply it automatically, but let the dev apply it and do the magic... I think it exists, I just didn't go deeper in piccolo's code

dantownsend:
Yeah, type conversion is tricky. Would require quite a bit of thought.

As for renaming primary keys, that should be possible. I've never tried it though, so I'm not sure if Postgres handles it gracefully or not.

brnosouza:
I did a little digging and it requires to drop the constraint, rename and readd the primary key:

```sql
-- Firstly, remove PRIMARY KEY attribute of former PRIMARY KEY
ALTER TABLE <table_name> DROP CONSTRAINT <table_name>_pkey;
```

```sql
-- Then change column name of  your PRIMARY KEY and PRIMARY KEY candidates properly.
ALTER TABLE <table_name> RENAME COLUMN <primary_key_candidate> TO id;
```

```sql
-- Lastly set your new PRIMARY KEY
ALTER TABLE <table_name> ADD PRIMARY KEY (id);
```

Issue #160, opened 2021-08-10
Is there a way to use all columns of a joined table?


brnosouza:
I was working in a feature and I had to list all the columns of a related table, is there a better way to select all columns without having to list all? Something like the `star` method in peewee.

dantownsend:
Using the example schema in the docs:

```python
class Manager(Table):
    id = Serial()
    name = Varchar()

class Band(Table):
    id = Serial()
    name = Varchar()
    manager = ForeignKey(references=Manager)
    popularity = Integer()
```

You can try something like:

```python
# To get the band name, and all manager columns
Band.select(
    Band.name,
    *[getattr(Band.manager, column._meta.name) for column in Manager._meta.columns]
).run_sync()
```

There could definitely be a nicer API for it - like you mention, something like:

```python
Band.select(
    Band.name,
    Band.manager.star
).run_sync()
```

I'll try and add this at some point.

PR #161, opened 2021-08-10
be able to select all related columns

dantownsend:
Fixes https://github.com/piccolo-orm/piccolo/issues/160

When joining on large tables, having to specify each column is tedious. This adds a convenience method called `all_columns`.

### DIFF
--- a/piccolo/columns/column_types.py
+++ b/piccolo/columns/column_types.py
@@ -1006,12 +1006,19 @@ class ForeignKey(Column):  # lgtm [py/missing-equals]
 
     **Joins**
 
-    Can also use it to perform joins:
+    You also use it to perform joins:
 
     .. code-block:: python
 
         >>> await Band.select(Band.name, Band.manager.name).first().run()
         {'name': 'Pythonistas', 'manager.name': 'Guido'}
+
+    To retrieve all of the columns in the related table:
+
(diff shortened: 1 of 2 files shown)


dantownsend:
@brnosouza Any thoughts on this new API?

piccolo/columns/column_types.py
```diff
@@ -1217,11 +1224,27 @@ def set_proxy_columns(self):
...
+
+            Band.select(Band.name, *Band.manager.all_columns()).run_sync()
+
+            """
+            return [
+                getattr(self, column._meta.name)
+                for column in _fk_meta.resolved_references._meta.columns
+            ]
+
+        setattr(self, "all_columns", all_columns)
```
sinisaos, line 1247: @dantownsend I think this looks great and reduces verbosity in queries.

sinisaos (commented):
@dantownsend This is great and I like it.

piccolo/columns/column_types.py
dantownsend: @sinisaos Awesome, thanks!

brnosouza:
> @brnosouza Any thoughts on this new API?

I think this is awesome, it really solves my issue and I like explict way of doing things, thanks a lot @dantownsend

PR #169, opened 2021-08-19
Random model builder

aminalaee:
random ModelBuilder class to ease testing like `blender` or `django model bakery` or `django dynamic fixtures`.


dantownsend:
@aminalaee This is amazing - thanks. Just a few thoughts.

## Async build

Would there ever be a situation where a user would want to run `build` asynchronously? Most unit tests are sync anyway, but just interested to hear your thoughts.

## Column references for args

We pass in args to specify the defaults, for example:

```python
band = ModelBuilder().build(Band, name='Guido', manager=manager)
```

We could potentially let the user pass in a dict instead, so direct column references could be used:

```python
band = ModelBuilder().build(Band, defaults={Band.name: 'Guido', Band.manager: manager})

# The user could still use strings if they like
band = ModelBuilder().build(Band, defaults={'name': 'Guido', 'manager': manager})
```

What do you think about this?

## Related object defaults

And this is probably way too hardcore for a V1, but it would cool if we could specify default values on related objects.

```python
band = ModelBuilder().build(Band, defaults={Band.name: 'Pythonistas', Band.manager.name: 'Guido'})

# Or
band = ModelBuilder().build(Band, defaults={'name': 'Pythonistas', 'manager.name': manager})
```

aminalaee:
@dantownsend 
For `async build` I think we can have a consistent API the same as Piccolo.
We can have `ModelBuilder().build()`  and`ModelBuilder().build_sync()`. The only drawback I can think is that right now we can tell ModelBuilder not to save the model and only return that.
In that case it doesn't matter if we call `.build()` or `build_sync()`.

For Column references for args I agree it's more convenient.
Let me give it try.

brnosouza:
@aminalaee this is amazing work, I recently started using FactoryBoy with piccolo for this same case and this is a great replacement, but I was thinking, wouldn't be better to just integrate it with [Faker](https://faker.readthedocs.io/) the same way FactoryBoy does instead of creating a randomizer, it could be a soft requirement?

piccolo/testing/model_builder.py
```diff
@@ -0,0 +1,155 @@
...
+
+        Examples:
+
+            manager = ModelBuilder.build_sync(Manager)
+            manager = ModelBuilder.build_sync(Manager, name='Guido')
+            manager = ModelBuilder(persist=False).build_sync(Manager)
+            manager = ModelBuilder(minimal=True).build_sync(Manager)
+            band = ModelBuilder.build_sync(Band, manager=manager)
+        """
+        return asyncio.run(
```
dantownsend, line 78: Can use `run_sync` from `piccolo.utils.sync` here. It's pretty similar, but slightly more robust.

aminalaee:
@brnosouza thank you. TBH that was my first thought too, but there are a couple of reasons I personally didn't go that way:

- they provide more features than we need so it's an overkill. For example they can build different string formats like email, country, etc but they're not useful in Piccolo since we don't have the formattings like django EmailField or PasswordField...
- the builder logic isn't really that complicated and I would be cautious to add new requiements to the project. That would cause people to reconsider using the package.

But anyway if you find any good reason we should use them, lets discuss

brnosouza:
@aminalaee I see your point on this, yeah, it's a little overkill. If the someone needs more, they could overwrite the 
`__DEFAULT_MAPPER` property in a subclass of ModelBuilder, thanks for the explanation.

Issue #170, opened 2021-08-20
Primary Key missing in auto-generated migration


adriangb:
Using Postgres & Piccolo 0.32.0, the auto-generated migrations fail to include the primary key.

Schema:

```python
from piccolo.table import Table
from piccolo.columns import ForeignKey, UUID

class Types(Table):
    id = UUID(primary_key=True)

class Things(Table):
    type = ForeignKey(references=Types)
```

When I try to run migrations via:

```
piccolo migrations new my_app --auto
piccolo migrations forwards all
```

I get:

```
Running MigrationManager ...
The command failed.
foreign key constraint "things_type_fkey" cannot be implemented
DETAIL:  Key columns "type" and "id" are of incompatible types: integer and uuid.
```

Looking at the generated migration, I see:

```python
class Types(Table, tablename="types"):
    pass
```

Adding the PK manually to the migration fixes the issue:

```python
class Types(Table, tablename="types"):
    id = UUID(primary_key=True)
```

aminalaee:
@adriangb Thank you for reporting this.
There's an open issue to fix this #156 . Unfortunately the fix is not backwards compatible (yet) but we're trying to fix that.

adriangb:
Nice! Feel free to close this if it's a duplicate

aminalaee:
@adriangb That's fine I think. We didn't really have an issue for that PR. Thanks for reporting it.

dantownsend:
I think the issue here is the `ForeignKey` should be created with type `UUID`, otherwise it can't store the reference to a `UUID` primary key.

This should work - I'll investigate what went wrong.

dantownsend:
I've been able to replicate the bug locally, and know what the issue is.

dantownsend:
@adriangb Thanks for reporting this, and @aminalaee for the assistance. Should be fixed on PyPI now (v0.33.0) - let me know if there are still issues.

PR #173, opened 2021-08-20
fix foreign keys in in auto migrations

dantownsend:
Fixes https://github.com/piccolo-orm/piccolo/issues/170

The issue was with auto migrations, and serialising Foreign Keys.

The old approach didn't include the custom primary key:

Screenshot 2021-08-20 at 08 45 00

The new approach does:

Screenshot 2021-08-20 at 08 44 49

This is important because when creating a `ForeignKey` the type needs to be the same as the primary key being referenced.

### DIFF
--- a/piccolo/apps/migrations/auto/serialisation.py
+++ b/piccolo/apps/migrations/auto/serialisation.py
@@ -335,6 +335,21 @@ def serialise_params(params: t.Dict[str, t.Any]) -> SerialisedParams:
             extra_imports.append(
                 Import(module=Table.__module__, target="Table")
             )
+
+            extra_imports.append(
+                Import(
+                    module=value._meta.primary_key.__class__.__module__,
+                    target=value._meta.primary_key.__class__.__name__,
+                )
+            )
+            # Include the extra imports and definitions required for the
+            # primary column params.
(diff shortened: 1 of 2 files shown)


piccolo/apps/migrations/auto/serialisation.py
```diff
@@ -99,7 +99,20 @@ def __eq__(self, other):
...
+        pk_column = self.table_type._meta.primary_key
+        pk_column_name = pk_column._meta.name
+        serialised_pk_column = SerialisedColumnInstance(
+            pk_column,
+            serialised_params=serialise_params(params=pk_column._meta.params),
+        )
+
+        return (
+            f'class {class_name}(Table, tablename="{tablename}"): '
+            f"{pk_column_name} = {serialised_pk_column}"
```
aminalaee, line 114: A quick question: It creates PK for both auto and custom primary key columns, right?

aminalaee (approved):

dantownsend:
@aminalaee thanks

piccolo/apps/migrations/auto/serialisation.py
dantownsend: that's right

PR #176, opened 2021-08-22
Update PrimaryKey deprecation warning

tonybaloney:
The `PrimaryKey` deprecation warning doesn't tell you what you should use instead.

This PR leads people toward the two most likely options.

### DIFF
--- a/piccolo/columns/column_types.py
+++ b/piccolo/columns/column_types.py
@@ -530,7 +530,8 @@ class PrimaryKey(Serial):
 
         colored_warning(
             "`PrimaryKey` is deprecated and "
-            "will be removed in future versions.",
+            "will be removed in future versions. Use `UUID(primary_key=True)`"
+            " or `BigInt(primary_key=True)` instead.",
             category=DeprecationWarning,
         )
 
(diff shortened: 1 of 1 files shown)


piccolo/columns/column_types.py
```diff
@@ -530,7 +530,8 @@ def __init__(self, **kwargs) -> None:
 
         colored_warning(
             "`PrimaryKey` is deprecated and "
-            "will be removed in future versions.",
+            "will be removed in future versions. Use `UUID(primary_key=True)`"
+            " or `BigInt(primary_key=True)` instead.",
```
aminalaee, line 534: As you mentioned, It can also be `Serial`, `Integer` or something else. Maybe we can point them to the docs as they can see how to specify a primary key?

dantownsend: Yeah, the recommended way currently is to just let Piccolo create the primary key for you, in which case you don't need to do anything.

Custom primary keys were added because a common feature request was UUID primary keys.

I agree that this deprecation warning could be more useful. Having a link to the docs could work.

PR #180, opened 2021-08-22
Update orjson.txt

hipertracker:
All versions below 3.5.1 have no binaries for M1 processors.


dantownsend:
@hipertracker Thanks for this

PR #181, opened 2021-08-22
`get_or_create` method

### LINKED ISSUE #165: Feature request: get_or_create
To simplify create or update it would be nice to have a `get_or_create` method.

So this:

```python
band = await Band.objects().where(name=="U2").first().run()
if not band:
    band = Band(name="U2")
band.popularity = 100
await band.save().run()
```

Can be turned into this:

```python
band = await Band.objects().get_or_create(name=="U2", defaults={"popularity": 100}).run()
```

aminalaee:
Closes #165 

This is a simple draft to get some feedback for `get_or_create` method and will need refactoring.

As the `get_or_create` needs to do one or two queries, it cannot return an `Objects` for user to call `run` method.
It has to run the `.run()` or `.run_sync()` methods itself.

### DIFF
--- a/docs/src/piccolo/query_types/objects.rst
+++ b/docs/src/piccolo/query_types/objects.rst
@@ -89,6 +89,23 @@ object, you can do so using ``get_related``.
     >>> print(manager.name)
     'Guido'
 
+get_or_create
+-------------
+
+With ``get_or_create`` you can get an existing record matching the criteria,
+or create a new one with the ``defaults`` arguments:
+
+.. code-block:: python
+
+    band = Band.objects().get_or_create(
+        Band.name == 'Pythonistas', defaults={Band.popularity: 100}
+    ).run_sync()
+
+    # Or using string column names
+    band = Band.objects().get_or_create(
+        Band.name == 'Pythonistas', defaults={'popularity': 100}
+    ).run_sync()
+
 Query clauses
 -------------
 
--- a/piccolo/query/methods/objects.py
+++ b/piccolo/query/methods/objects.py
@@ -14,12 +14,44 @@ from piccolo.query.mixins import (
     WhereDelegate,
 )
 from piccolo.querystring import QueryString
+from piccolo.utils.sync import run_sync
 
 from .select import Select
 
 if t.TYPE_CHECKING:  # pragma: no cover
     from piccolo.columns import Column
     from piccolo.table import Table
+
+
+@dataclass
+class GetOrCreate:
+    query: Objects
+    where: Combinable
+    defaults: t.Dict[t.Union[Column, str], t.Any]
+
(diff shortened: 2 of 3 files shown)


dantownsend:
@aminalaee Thanks a lot for this.

Good point with the `run` / `run_sync` issue.

Maybe we can return a class instance called something like `GetOrCreate`? It would just have `run` and `run_sync` methods on it.

aminalaee:
@dantownsend Thanks for the feedback. If I understand correctly, that would add an additional query to the database, if the record exists.

The flow will then be like this (always doing two queries):
- We have to check if record exists in the `get_or_create`
- We will return a query to the user
- User will call the query which will return the object or create it.

But if we don't return a query, and instead return the final object (if record exists, we've only done one query):
- We get the object
- If it doesn't exist, another query to create it

dantownsend:
@aminalaee Even though there are two queries, I think we can hide it from the users. Maybe something like this? It's very similar to what you've already done:

```python
class GetOrCreate:
    def __init__(self, query: Objects, where: Combinable, defaults: t.Dict[t.Union[Column, str], t.Any]):
        self.query = query
        self.where = where
        self.default = defaults

    async def run(self):
        instance = await self.query.where(self.where).first().run()
        if instance:
            return instance

        instance = self.query.table()
        setattr(instance, self.where.column._meta.name, self.where.value)  # type: ignore
        for column, value in self.defaults.items():
            column = (
                instance._meta.get_column_by_name(column)
                if isinstance(column, str)
                else column
            )
            setattr(instance, column._meta.name, value)
            await instance.save().run()
        return instance

    def run_sync(self):
        return run_sync(self.run())


class Objects(Query):
    async def get_or_create(
        self, where: Combinable, defaults: t.Dict[t.Union[Column, str], t.Any]
    ) -> GetOrCreate:
        return GetOrCreate(query=self, where=where, defaults=defaults)
```

dantownsend:
I don't want to scope creep, but it would be cool it we had a `get` method too. Maybe as part of this, or in the future.

```python
await MyTable.objects().get(MyTable.name == 'Bob').run()

# It would just be an alias of:
await MyTable.objects().where(MyTable.name == 'Bob').first().run()
```

aminalaee:
@dantownsend I'll create another PR to for `get` to go after this and move the discussion there.

dantownsend:
@aminalaee Cool, makes sense

dantownsend:
@aminalaee I think this is good to go. I just made two minor tweaks - allowing the query to be directly awaited (this isn't mentioned in the docs, but is there as a convenience with other queries). And add an empty dict as the default for `defaults`.

aminalaee:
@dantownsend Thank you for that.

Issue #178, opened 2021-08-22
Automigrations fails for models using ForeignKey


hipertracker:
I can't run migrations with `--auto` flag when a model uses a `ForeignKey`.
```
class Book(Table, tablename="books"):
    name = Varchar(length=32, unique=True)
    label = JSON(default=dict())
    sort = Integer(default=0, index=True)
    available = Boolean(default=True, index=True)
    tags = Array(base_column=Varchar(length=32), default=[])

class BookAlias(Table, tablename="book_aliases"):
    name = Varchar(length=32)
    book = ForeignKey(references="Book")
```
piccolo migrations new home --auto
Running Postgres version 13.4
Creating new migration ...
The command failed.
**unhashable type: 'list'**

hipertracker:
It's not related to the Postgres version. I tested it also on the supported Postgres 11.1 with the same effect

```
$ piccolo migrations new home --auto
Running Postgres version 11.1
Creating new migration ...
The command failed.
unhashable type: 'list'
```

dantownsend:
@hipertracker It will be related to the default argument on `tags` - seems like an edge case which isn't caught by the unit tests. I'll take a look at this.

PR #182, opened 2021-08-22
fix error with unhashable types and `compare_dicts`

dantownsend:
Fixes https://github.com/piccolo-orm/piccolo/issues/178

`compare_dicts` is used by migrations to work out if any `Column` arguments have changed (for example, `Varchar(length=255)` vs `Varchar(length=512)`.

It couldn't handle unhashable types properly (for examples, lists and dictionaries) because the previous implementation depended on comparing sets.

### DIFF
--- a/piccolo/apps/migrations/auto/diffable_table.py
+++ b/piccolo/apps/migrations/auto/diffable_table.py
@@ -20,8 +20,29 @@ def compare_dicts(dict_1, dict_2) -> t.Dict[str, t.Any]:
     """
     Returns a new dictionary which only contains key, value pairs which are in
     the first dictionary and not the second.
+
+    For example:
+        dict_1 = {'a': 1, 'b': 2}
+        dict_2 = {'a': 1}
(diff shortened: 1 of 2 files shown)


piccolo/apps/migrations/auto/diffable_table.py
```diff
@@ -20,8 +20,29 @@ def compare_dicts(dict_1, dict_2) -> t.Dict[str, t.Any]:
...
+
     """
-    return dict(set(dict_1.items()) - set(dict_2.items()))
+    output = {}
+
+    for key, value in dict_1.items():
+
+        dict_2_value = dict_2.get(key, ...)
+        if dict_2_value is ...:
+            output[key] = value
```
aminalaee, line 40: I'd personally prefer to `continue` after this as there's only one condition after this and avoid the next `else` but I think it's just a preferrence.

aminalaee (approved):
This is great.

PR #183, opened 2021-08-22
add objects `get` method

aminalaee:
Adding `objects().get()` method.

```
await MyTable.objects().get(MyTable.name == 'Bob').run()
```

It is just be an alias of:
```
await MyTable.objects().where(MyTable.name == 'Bob').first().run()
```


dantownsend:
@aminalaee This looks good to me - don't think it needs anything else.

aminalaee:
@dantownsend Sure, I'll do that.

dantownsend:
@aminalaee Thanks for this - have been meaning to add it for a long time!

Issue #200, opened 2021-08-27
Get dict from ORM objects?


wmshort:
Is it possible to get a dict of column names and values easily from ORM objects, as if you had originally done .select()? I mean, once you have some ORM objects, I'm finding I then sometimes want to treat these objects purely in terms of their values. I guess there is the brutish .__dict__, but that wouldn't offer the filtering or aliasing opportunities of .select(). I guess I am asking if it's possible to .select() from the result of a Table.objects() query.

dantownsend:
I agree, having a separate `to_dict` method would be better than overriding `__dict__`.

dantownsend:
@wmshort There's an open PR which adds the fix.

Let me know if this is what you expected, if so I'll release it later today.

wmshort:
That's brill, yes! I haven't tested it yet, is it possible to provide columns for inclusion, as well as aliasing? I had imagined something like this:

```
object.to_dict()       # all model columns, except of course the internal ones
object.to_dict(Table.column_name)    # include only the user specified column
object.to_dict(Table.column_name.as_alias('column_alias'))    # alias the specified column name
```

Or mix and match
``
object.to_dict(Table.column1_name, Table.column2_name.as_alias('column_alias'))
# include column1 and column2 (but not, say, column3) and simultaneously alias column2, leaving column1 as-is
```

dantownsend:
@wmshort Cool, makes sense - I can see that being useful. Have updated the PR.

PR #201, opened 2021-08-28
added `to_dict`

dantownsend:
Fixes https://github.com/piccolo-orm/piccolo/issues/200

Added a `to_dict` method.

If you just use `__dict__` , you get some non-column values. By using `to_dict` it's just the column values. Here's an example:

```python
class MyTable(Table):
    name = Varchar()

instance = MyTable.objects().first().run_sync()

>>> instance.__dict__
{'_exists_in_db': True, 'id': 1, 'name': 'foo'}

>>> instance.to_dict()
{'id': 1, 'name': 'foo'}
```

You can also use aliases to change the output keys:

```python
>>> instance.to_dict(MyTable.name.as_alias('title'))
{'id': 1, 'title': 'foo'}
```

### DIFF
--- a/docs/src/piccolo/query_types/objects.rst
+++ b/docs/src/piccolo/query_types/objects.rst
@@ -151,14 +151,15 @@ If you need to convert an object into a dictionary, you can do so using the
     >>> band.to_dict()
     {'id': 1, 'name': 'Pythonistas', 'manager': 1, 'popularity': 1000}
 
-You can override some of the column names if you like:
+If you only want a subset of the columns, or want to use aliases for some of
+the columns:
 
 .. code-block:: python
 
     band = Band.objects().first().run_sync()
 
-    >>> band.to_dict(Band.name.as_alias('title'))
-    {'id': 1, 'title': 'Pythonistas', 'manager': 1, 'popularity': 1000}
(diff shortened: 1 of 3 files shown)


aminalaee (commented):
Do we need tests for ForeignKey or Array columns? Just making sure we're covering all edge cases.

sinisaos:
@aminalaee No need. I tried both and everything is fine. If Daniel still wants to add tests something like this would be ok in `table/test_to_dict.py`

```python
def test_to_dict_fk(self):
    """
    Make sure that `to_dict` works correctly with FK.
    """
    self.insert_row()

    instance = Band.objects().first().run_sync()
    dictionary = instance.to_dict()
    self.assertEqual(
        dictionary,
        {"id": 1, "name": "Pythonistas", "manager": 1, "popularity": 1000},
    )
```
and in `columns/test_array.py`
```python
def test_to_dict(self):
    MyTable(value=[1, 2, 3]).save().run_sync()
    
    instance = MyTable.objects().first().run_sync()
    dictionary = instance.to_dict()
    self.assertEqual(
        dictionary,
        {'id': 1, 'value': [1, 2, 3]},
    )
```

dantownsend:
@aminalaee @sinisaos Thanks for looking into those edge case.

The devil is definitely in the details with these things!

Issue #171, opened 2021-08-20
Insert & get


adriangb:
Is there any way to do an INSERT and subsequent SELECT of the inserted object w/o a second SELECT query? It look like INSERT returns the PK, but there's no way to add additional columns, change the cardinality, etc.

adriangb:
Indeed, `RETURNING` is hardcoded: https://github.com/piccolo-orm/piccolo/blob/14e7c72634a8614edec87e8cbbd8eb2101d84f93/piccolo/query/methods/insert.py#L64

dantownsend:
You're right - this is a good idea. I'll have a think about how to tackle this.

PR #203, opened 2021-08-28
initial prototype for returning clause

dantownsend:
Fixes https://github.com/piccolo-orm/piccolo/issues/171

As pointed out in the above issue, there's no way to customise what the insert method returns.

We currently just return the primary key for the inserted rows.

```python
>>> Manager.insert(Manager(name="Bob"), Manager(name="Sally")).run_sync()
[{'id': 3}, {'id': 4}]
```

Being able to customise what gets returned is a nice usability improvement, and in some cases could save another SELECT query.

For example:

```python
>>> Manager.insert(Manager(name="Bob"), Manager(name="Sally"), returning=[Manager.id, Manager.name]).run_sync()
[{'id': 3, 'name': 'Bob'}, {'id': 4, 'name': 'Sally'}]
```

The implementation is fairly straightforward with Postgres, but SQLite only supports the RETURNING clause in fairly recent versions (>= 3.35.0), so needs more consideration:

https://www.sqlite.org/lang_returning.html

We could also potentially add a `returns_objects` argument, which would returns the new rows as Table instances instead of a list of dictionaries.

### DIFF
--- a/piccolo/query/methods/insert.py
+++ b/piccolo/query/methods/insert.py
@@ -4,24 +4,40 @@ import typing as t
 from dataclasses import dataclass
 
 from piccolo.query.base import Query
-from piccolo.query.mixins import AddDelegate
+from piccolo.query.mixins import AddDelegate, ReturningDelegate
 from piccolo.querystring import QueryString
 
 if t.TYPE_CHECKING:  # pragma: no cover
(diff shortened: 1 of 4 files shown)


dantownsend:
Replaced by https://github.com/piccolo-orm/piccolo/pull/564

PR #217, opened 2021-09-03
Nested objects prototype

dantownsend:
This is a prototype for nested objects.

Here's an example:

```python
# Without nested
band = await Band.objects().first().run()
>>> band.manager
1

# With nested
band = await Band.objects(Band.manager).first().run()
>>> band.manager
<Band 1>
>>> band.manager.name
'Guido'
```


wmshort:
Brill. Very excited for this feature.

Only in this example:
> # With nested
> band = await Band.objects(Band.manager).first().run()
> >>> band.manager
> <Band 1>
> >>> band.manager.name
> 'Guido'
> ```
the object accessed by ``band.manager`` should be ``<Manager>`` rather than <Band>``, shouldn't it? (But definitely not just the value ``1``!).

If the argument passed to ``objects()`` is the foreign-key tables to load as objects, would there also be an ``all()`` option, like ``select(Table.all_columns())``, which recursively returns all nested models as objects? Or maybe ``objects(nested=True)`` would do this across the board? Are there situations where you might want some nested models NOT as objects but others as?

Anyway, something like:
```
class Manager...
class Genre:
    name: str
    decade: str
class Band:
     manager: Manager
     genre: Genre

>>> b = Band.objects(nested=True).where(Band.mamanager.name == "Guido").run_sync()
>>> b.genre
<Genre ...>
```

or, alternatively, ``Band.objects(Band.all_related()).where(Band.manager.name == "Guido").run_sync()``

I actually like the option of being able to specify particular tables to be nested as objects.

dantownsend:
Thanks for the feedback.

When using the related functionality, `Band.manager` should return a `Manager` instance, rather than an integer.

I like the `all_related()` idea.

dantownsend:
Almost there - just missing support for `get_or_create`.

dantownsend:
@wmshort I think this is done now - will merge it in soon.

Do the API docs make sense?

https://github.com/piccolo-orm/piccolo/blob/28701e33b7d8cdae9f1ecbc0ea19a58162e541ba/docs/src/piccolo/query_types/objects.rst#prefetching-related-objects

wmshort:
@dantownsend amazing! Yes, the docs are looking really good on this -- clear, and thorough. Looking forward to this new funxtionality. Just one question came up, namely, in a situation like this:

```ticket = Ticket.objects(
    Ticket.concert,
    Ticket.concert.all_related()
).first().run_sync()```

(or the equivalent with ``prefetch`` -- nice!), what would it mean to omit ``Ticket.concert``? Is there ever a situation where you would want to load the FKs on a nested model (concert) but not the object itself? (Could that even work?) In that case, might make sense (and be less cluttery) to treat ``all_related`` as meanimg the model in question and its FKs (and theirs)?

**Comment by dantownsend** [comment]
@wmshort Good point.

This will work:

```python
ticket = Ticket.objects(
    Ticket.concert.all_related()
).first().run_sync()
```

But there's some weird behaviour - the intermediate `Concert` object exists, but some of it's values are missing (such as it's `id`).

I'll try and modify it, so all intermediate objects are loaded fully.

wmshort:
That's the behaviour I assumed. But I think it could catch some people out if they omit the intermediate model. It seems more intuitive to have all_related() load everything on the intermediates.

William Michael Short
Lecturer in Classics
Department of Classics & Ancient History
University of Exeter
________________________________

PR #219, opened 2021-09-06
Docker integration in tests

antonio-one:
A small docker integration for the postgres tests for early feedback in case you're interested in this kind of thing.

Make sure nothing occupies port 5432 and run the usual `/scripts/test-postgres.sh`

Let me know if this is useful at all or rubbish.

If you want to see this running outside the tests

```
python -m tests.utils.setup --db_engine=postgres --action=spin_up
```
OR 

```
python -m tests.utils.setup --db_engine=postgres --action=tear_down
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,5 @@ htmlcov/
 prof/
 .env/
 .venv/
+.idea/
+.python-version
--- a/docs/src/piccolo/contributing/index.rst
+++ b/docs/src/piccolo/contributing/index.rst
@@ -15,9 +15,9 @@ Get the tests running
  * Install default dependencies: ``pip install -r requirements/requirements.txt``
  * Install development dependencies: ``pip install -r requirements/dev-requirements.txt``
  * Install test dependencies: ``pip install -r requirements/test-requirements.txt``
- * Setup Postgres
  * Run the automated code linting/formatting tools: ``./scripts/lint.sh``
  * Run the test suite with Postgres: ``./scripts/test-postgres.sh``
+    * This will launch a docker postgres container with a piccolo database in port 5432
  * Run the test suite with Sqlite: ``./scripts/test-sqlite.sh``
 
 Contributing to the docs
--- a/requirements/dev-requirements.txt
+++ b/requirements/dev-requirements.txt
@@ -7,3 +7,4 @@ twine==3.1.1
 mypy==0.782
 pip-upgrader==1.4.15
 wheel==0.36.2
+pdbpp==0.10.3-
--- a/requirements/test-requirements.txt
+++ b/requirements/test-requirements.txt
@@ -2,3 +2,5 @@ coveralls==2.2.0
 pytest-cov==2.10.1
 pytest==6.2.1
 python-dateutil==2.8.1
+docker==5.0.2
+faker==8.12.1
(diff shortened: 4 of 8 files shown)


dantownsend:
@aminalaee Have you got any thoughts about this?

aminalaee:
@dantownsend To be honest I need to have more info about it.
Right now I don't see the point of all this, as it's not up to piccolo to setup dev environment for people. What happens if someone doesn't have docker installed or wants to do test on local/remote postgres instances.

And if we wanted to guide people towards the docker testing, we could have a simple `docker-compose` that could setup database(s) and we wouldn't need all this code.


@antonio-one thanks for the effort. Can you please clarify a bit?

antonio-one:
@dantownsend and @aminalaee thanks for getting back to me.

While I was going through the [contribution steps](https://piccolo-orm.readthedocs.io/en/latest/piccolo/contributing/index.html), one of them was "setup postgres," which stroke me as a bit out of place, because every other step was automated. For the sake of consistency and for maybe helping others onboard with less effort to a new ORM, it may be nice to at a minimum, offer a automated setup option. Every little helps when you want to increase adoption I guess.

If one chose to go down that route the could always use a dockerfile and/or docker-compose instead of a chattier version of the python docker sdk. There will always be more than one ways to skin a cat :-) 

I also appreciate the fact that someone may wish to use a local or even remote postgres. This is something that can be certainly built in to allow users to chose whether to use docker, a local or a remote host when testing.

Let me know if you require more clarifications.

dantownsend:
@antonio-one Thanks for that.

With most of our testing currently we rely on GitHub actions. Locally I just use Postgres.app, but I could see a Dockerised setup being useful for some people.

I had an idea though - rather than this being for testing purposes, what if it was a Piccolo app which read the contents of your `piccolo_conf.py` file, and setup the database with those credentials for you?

For example:

```python
# piccolo_conf.py
DB = PostgresEngine(
    config={
        "database": "my_db",
        "user": "my_user",
        "password": "abc123",
        "host": "localhost", # This could have to be localhost
        "port": 5432,
    }
)
```

It would then setup a `my_db` database with that username and password.

It could be bundled as a separate Python package, and we can link to it from the main docs. The advantage is it wouldn't add extra dependencies to the main repo.

There could be say 4 different commands: `piccolo dockerdb create`, `piccolo dockerdb start`, `piccolo dockerdb stop`, `piccolo dockerdb destroy`.

Just throwing it out there.

antonio-one:
If you think this will be useful to yourselves and your users I'm happy to repurpose this code to this pattern.

dantownsend:
@antonio-one Thanks. Would you like me to create you a repo as part of the Piccolo namespace? Like `piccolo_docker` or something? Or do you want to create one yourself?

antonio-one:
Hi @dantownsend. Happy to fork from your namespace as long as you're happy to review PRs etc.

dantownsend:
@antonio-one Cool - here you go: https://github.com/piccolo-orm/piccolo_docker

antonio-one:
Roger roger. Thank you very much @dantownsend.

dantownsend:
Going to close this for now - will pick up the conversation in the `piccolo_docker` repo.

https://github.com/piccolo-orm/piccolo_docker

Issue #10, opened 2020-09-20
Roadmap


dantownsend:
- [x] JSON fields
- [ ] More aggregation functions - e.g. SUM
- [x] Aliases in SELECT queries (e.g. `SELECT foo AS bar FROM my_table`)
- [ ] Improved support for data migrations
- [x] Fixtures
- [x] Schema visualisation tool
- [x] `piccolo sql_shell new` command - execute SQL directly in a shell, via the configured engine
- [x] `piccolo shell new` command - like the playground, but for your own tables
- [x] Shorter aliases for commands in CLI, or tab completion
- [ ] Improved documentation for how to test Piccolo apps
- [ ] Nested transactions
- [x] Move Pydantic integration from Piccolo API into Piccolo
- [x] Allow the `ForeignKey` `references` argument to accept a string such as `'my_app.SomeTable'` when circular imports are an issue.
- [ ] Support Postgres `Enum` types.
- [ ] Row level constraints.
- [ ] Making foreign key constraints optional on ``ForeignKey`` columns.
- [x] Allow UUIDs as primary keys?
- [ ] Subqueries

dantownsend:
Please feel free to suggest features for the roadmap.

99hats:
Nice project! 
How about row level security? 
Also, +1 for Allow the ForeignKey references argument to accept a string such as 'my_app.SomeTable' when circular imports are an issue.

dantownsend:
@99hats Thanks for the feedback. Accepting ``ForeignKey`` references as strings is in progress as we speak, and will be in the next release. How do you envisage row level security working? I could add a way to define the policies, and run them when creating tables / running migrations.

99hats:
@dantownsend That sounds great. It could also be added to the web admin. Here's how supabase.io (a firebase clone using postgres) implements it on web admin: https://supabase.io/docs/guides/auth#row-level-security

dantownsend:
@99hats Thanks for the link - I'll take a look at how they're doing it.

There's a fix for the circular dependency issue in Piccolo 0.14.10 - ``ForeignKey`` can now accept strings.

```python
class Band(Table):
    name = Varchar()
    manager = ForeignKey('Manager')

class Manager(Table):
    name = Varchar()

```

https://piccolo-orm.readthedocs.io/en/latest/piccolo/schema/column_types.html#foreignkey

99hats:
[...]

PR #229, opened 2021-09-13
Fixtures app

dantownsend:
Adding a fixtures app which can be used to dump data from the database to a JSON file, and then reload it again.

To create a fixture:

```bash
piccolo fixture dump --apps=blog --tables=Post,PostComment > fixture.json
```

To load a fixture:

```bash
piccolo fixture load fixture.json
```

# Reasons for adding it

## Sharing data with colleagues

A Piccolo project can contain a fixture with all of the data a colleague needs to setup a project with. For example, if it's an app about selling cars, the fixture could contain all of the makes and models of cars.

## Deploying to a new environment

When you want to deploy your application to production for the first time, you can load the fixture to populate the database with essential data.

## Why not use pg_dump and pg_restore?

Whilst these are great tools, and achieve similar goals to fixtures, they're locked to a single Postgres version. With fixtures, you should be able to dump from say Postgres 12 and load into Postgres 13. This means you don't have to rebuild your fixtures when you update your Postgres version.

Also, having your fixtures in JSON is better for version control, and makes them easier to modify by hand.

# Side effects

In order to implement this change, some fairly large changes had to be made to the wider codebase (though no breaking changes).

## Test refactoring

The tests were refactored slightly - in the past there was a single example Piccolo app which the tests use (containing the `Band`, `Manager` etc tables). There are now multiple example apps. The new one is called `mega`, and contains a table called `MegaTable` which has every column type, and is very useful for testing serialisation.

## Moving Pydantic into core

All of the Pydantic functionality is currently in Piccolo API. It was been discussed in the past about moving this into core (https://github.com/piccolo-orm/piccolo/issues/10#issuecomment-871776533).

The downside of doing so, is a tighter coupling to external dependencies. But on the other hand, it's so heavily used elsewhere in the Piccolo ecosystem, that practically it makes little difference.

So this PR moves the Pydantic code into the main Piccolo repo. In a future PR I'll remove the Pydantic code from Piccolo API, and will replace it with proxies to the code in Piccolo core.


dantownsend:
@sinisaos @aminalaee any thoughts on this change? In particular, moving Pydantic support into the main Piccolo repo.

sinisaos:
@dantownsend More or less all orm's (Tortoise, especially Ormar etc.) have a strong integration with Pydantic in their core. I think moving a serializers from Piccolo API to Piccolo ORM is a good idea that has more pros than cons. Unlike other libraries, the Piccolo API is specific because it allows you to easily create a web API-s, and I think there is no difference if we use Pydantic serialization from Piccolo ORM or from Piccolo API.

aminalaee:
I think that's ok. Pydantic is pretty stable now. My only concern is that there was a discussion that Python 3.10 will cause some issues with Pydantic, not sure how serious that is.

dantownsend:
@sinisaos @aminalaee Thanks for your input.

There's a giant thread here about Pydantic and Python 3.10: https://github.com/samuelcolvin/pydantic/issues/2678

It seems like they've solved the issue (at least temporarily) 🤞

PR #230, opened 2021-09-14
added `pip install piccolo[all]` option

dantownsend:
Piccolo has a bunch of optional requirements now - most of the time you just want to install them all.

You can now do so using:

```
pip install 'piccolo[all]'
```

For the fully featured version, with all bells and whistles.


aminalaee (approved):

dantownsend:
@aminalaee Thanks for the review

PR #234, opened 2021-09-16
Fix `fixtures` app name typos

hipertracker:
Fixed wrong command


dantownsend:
@hipertracker Thanks! Good catch

Issue #233, opened 2021-09-15
Add `create_tables` function


dantownsend:
If you want to create a table without using migrations, you can do this:

```python
Band.create_table().run_sync()
```

If you're creating a bunch of tables, you need to do something like:

```python
Manager.create_table().run_sync()
Band.create_table().run_sync()
```

You have to be careful to create them in the right order because of `ForeignKey` columns.

We have a useful utility which can automatically sort tables based on their foreign keys so they're created in the right order.

https://github.com/piccolo-orm/piccolo/blob/23a8c1f8cedee518288ecb702454f62521583c13/piccolo/apps/migrations/auto/migration_manager.py#L165

So we can quite easily create a ``create_tables`` function in tables.py, which accepts a list of tables, passes them to `sort_table_classes`, and then creates each of them in turn.

```python
create_tables(Band, Manager, if_not_exists=True)
```

We will be able to use this to cut out boilerplate from our tests, and will also be useful for people writing simple data science scripts using Piccolo, who don't need migrations.

AliSayyah:
This sounds very useful. I'll get onto it.

AliSayyah:
@dantownsend I'm not sure what the function should return. do you have any idea?

dantownsend:
@AliSayyah I think it's fine not to return anything - if something goes wrong, an exception will be raised anyway.

dantownsend:
Fixed by https://github.com/piccolo-orm/piccolo/pull/235

PR #235, opened 2021-09-16
create multiple tables at once 

AliSayyah:
This PR adds the function described in this issue #233 and a unit test to test it out.

### DIFF
--- a/docs/src/piccolo/query_types/create_table.rst
+++ b/docs/src/piccolo/query_types/create_table.rst
@@ -19,3 +19,13 @@ To prevent an error from being raised if the table already exists:
 
     >>> Band.create_table(if_not_exists=True).run_sync()
     []
+
+Also, you can create multiple tables at once.
+
+This function will automatically sort tables based on their foreign keys so they're created in the right order:
+
+.. code-block:: python
+    
+    >>> from piccolo.table import create_tables 
+    >>> create_tables(Band, Manager, if_not_exists=True)
+   
--- a/piccolo/apps/fixture/commands/dump.py
+++ b/piccolo/apps/fixture/commands/dump.py
@@ -6,8 +6,8 @@ from piccolo.apps.fixture.commands.shared import (
     FixtureConfig,
     create_pydantic_fixture_model,
 )
-from piccolo.apps.migrations.auto.migration_manager import sort_table_classes
 from piccolo.conf.apps import Finder
+from piccolo.table import sort_table_classes
 
 
 async def get_dump(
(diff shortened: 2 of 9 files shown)


AliSayyah:
`sqlite3.OperationalError: near "CASCADE": syntax error`
It seems that SQLite3 doesn't support CASCADE.

dantownsend:
@AliSayyah That might be a separate bug. For now, just leave out `cascade=True`, and it should be OK.

AliSayyah:
> @AliSayyah That might be a separate bug. For now, just leave out `cascade=True`, and it should be OK.

in that case, it fails the Postgres tests.
I think resolving this issue with try/except to cover both cases would be enough for now.

dantownsend:
It looks great.

The only thing I'm not sure about is where `create_tables` should live. I originally thought about it being in `piccolo/tables.py`, but not sure if that causes circular dependency issues (due to `sort_table_classes`).

AliSayyah:
> It looks great.
> 
> The only thing I'm not sure about is where `create_tables` should live. I originally thought about it being in `piccolo/tables.py`, but not sure if that causes circular dependency issues (due to `sort_table_classes`).

yes, I tried that and you are right. it causes circular import. the current module is not very related to the function but I couldn't find a better place.

dantownsend:
I don't mind moving `sort_table_classes` - as it's more broadly applicable than just migrations.

`sort_table_classes` could live in `tables.py` or a new file in `piccolo/utils`, if that fixes the circular import problem.

AliSayyah:
> I don't mind moving `sort_table_classes` - as it's more broadly applicable than just migrations.
> 
> `sort_table_classes` could live in `tables.py` or a new file in `piccolo/utils`, if that fixes the circular import problem.

ok, that's better. I'll see what works best.

AliSayyah:
I think it's cleaner now. if it's alright now, I can update the docs too.

dantownsend:
@AliSayyah Yeah, I like that - thanks.

For the docs, just a short section at the end of this page should be OK I guess:

https://piccolo-orm.readthedocs.io/en/latest/piccolo/query_types/create_table.html

dantownsend:
@AliSayyah Perfect, thanks!

AliSayyah:
@dantownsend 
Thank you for creating this amazing project.
I had a wonderful experience contributing and helping a little bit.
And I would like to help out more. we can have a conversation about how can I be more helpful if it's okay with you.

dantownsend:
@AliSayyah Thanks for the kind words, and your contributions.

You're welcome to contribute more - you can take a look at the issues, and I'll try and add some more with ideas I've had for improvements.

Also, if you have any ideas about how to improve the project I'd like to hear them.

Issue #232, opened 2021-09-15
Fix bug with Decimal / Numeric columns


dantownsend:
There's a subtle bug where if you have a `Decimal` / `Numeric` column, and don't specify a `digits` argument.

https://github.com/piccolo-orm/piccolo/blob/23a8c1f8cedee518288ecb702454f62521583c13/piccolo/columns/column_types.py#L908

For example:

```python
class MyTable(Table):
    # In this case Postgres just allows the maximum digits the database will support:
    value = Numeric()
```

The problem is when using `create_pydantic_model` it assumes the `precision` and `scale` properties of `Numeric` / `Decimal` return an integer - but they raise an exception.

https://github.com/piccolo-orm/piccolo/blob/23a8c1f8cedee518288ecb702454f62521583c13/piccolo/utils/pydantic.py#L120-L123

We just need to return something sensible from `precision` and `scale` if `digits` isn't set.

And if they're not available, we should use a standard decimal field in Pydantic rather than `pydantic.condecimal`.

AliSayyah:
@dantownsend pydantic's default value for max_digits and decimal_places is None, so if we return None when precision and scale are not defined, it should be okay. 

> And if they're not available, we should use a standard decimal field in Pydantic rather than pydantic.condecimal.

so is it necessary to use another field in this case?

dantownsend:
@AliSayyah That makes sense - just have `max_digits` and `decimal_places` set to `None`

dantownsend:
Fixed by https://github.com/piccolo-orm/piccolo/pull/236

PR #236, opened 2021-09-16
Fix numeric bug

AliSayyah:
fixed the bug discussed in #232. 
if precision and scale are not set for the Numeric Field, 'None'  will be returned.


dantownsend:
@AliSayyah This is great, thanks. I just added a type annotation for `precision` and `scale`, as for some reason I didn't add them when I originally wrote it.

Will merge it in once the tests pass.

Issue #237, opened 2021-09-16
Dynamic query builder


AliSayyah:
in one of my projects, I need to dynamically query the database based on user input.
consider this function:
```python
def dynamic_query(columns, where, group_by):

    data = Table.select(columns).where(where).group_by(group_by).run_sync()
```
and of course, sometimes there won't be any 'where' clause or 'group_by' clause from the user side.
I noticed piccolo doesn't handle empty clauses.
am I doing something wrong here?
Is there a way to dynamically add clauses too?
and if something like this can't be done, I would like to try and add it and I would appreciate your guidance.

dantownsend:
You're right - if you use the `where` method it currently requires an argument.

```python
# This will fail:
MyTable.select().where().run_sync()
```

You can work around this by composing your query as follows:

```python
def dynamic_query(columns, where, group_by):
    query = Table.select(columns)
    if where:
        query = query.where(where)
    if group_by:
        query = query.group_by(group_by)
    data = query.run_sync()
```

However, I've wanted to change the `where` method for a while, so it can accept any number of where arguments.

```python
MyTable.select().where(MyTable.foo == 'a', MyTable.bar == 'b').run_sync()

# This is equivalent to an AND, which can be achieved by multiple where clauses:
MyTable.select().where(MyTable.foo == 'a').where(MyTable.bar == 'b').run_sync()
```

It's just a bit more compact when you have loads of where clauses.

If this were to be implemented, it would also solve your issue with the empty where method.

AliSayyah:
Thank you for your workaround. 
That would be a nice change. 
I will try my best to implement multiple arguments.

dantownsend:
@AliSayyah Thanks.

Now you mention it, there's nothing in the docs about being able to build up a query iteratively like in the example I shared with you - that's something I should probably add at some point.

PR #244, opened 2021-09-18
multiple arguments in WHERE clause

AliSayyah:
WHERE clause now will support multiple arguments(and no argument) based on discussion #237
I think testing needs to be extended to other methods too.
just want to know your idea before moving on to that.


dantownsend:
@AliSayyah It looks great to me.

AliSayyah:
> @AliSayyah It looks great to me.

Ok, nice.

AliSayyah:
@dantownsend Updated the docs.

dantownsend:
@AliSayyah This is great - thanks!

Issue #242, opened 2021-09-16
Schema generate doesn't output placeholder columns for column types it doesn't recognise


dantownsend:
As reported here: https://github.com/piccolo-orm/piccolo/issues/193#issuecomment-921014890

PR #246, opened 2021-09-18
#242 bug-fix 

AliSayyah:
This fixes the bug reported in #242.
And I tested it myself, but if a unit test is necessary, I need a little help writing it.


dantownsend:
@AliSayyah Thanks for this. I added a test for unknown column types - it was a head scratcher.

PR #254, opened 2021-09-19
Update piccolo_apps.rst

hipertracker:
Add missing information about the --trace flag to the command line call


dantownsend:
@hipertracker Thanks for this

Issue #241, opened 2021-09-16
Support double precision column type


dantownsend:
As discussed here:

https://github.com/piccolo-orm/piccolo/issues/193#issuecomment-921014890

AliSayyah:
@dantownsend what do you think about using Decimal type with a fixed scale of 15 for `Double Precision`?
and I noticed that float has been used to represent the `Real` type with no decimal limitation.
This wouldn't be a problem while inserting data to the table?

dantownsend:
@AliSayyah For the types, I've gone off the asyncpg docs:

https://magicstack.github.io/asyncpg/current/usage.html?highlight=decimal#type-conversion

So looks like double precision should also be a float.

You're right, `Real` columns are also represented by Python `float` values.

What seems to happen is there's no limit on what floats you can insert, but you don't get 100% accurate values back from the database (i.e. if you insert 3.0913 you might get something like 3.0920001 back - these are totally made up numbers, just for illustration).

AliSayyah:
> @AliSayyah For the types, I've gone off the asyncpg docs:
> 
> https://magicstack.github.io/asyncpg/current/usage.html?highlight=decimal#type-conversion
> 
> So looks like double precision should also be a float.
> 
> You're right, `Real` columns are also represented by Python `float` values.
> 
> What seems to happen is there's no limit on what floats you can insert, but you don't get 100% accurate values back from the database (i.e. if you insert 3.0913 you might get something like 3.0920001 back - these are totally made up numbers, just for illustration).

My understanding was that if `Real` has, for example, 5 decimal digits, it should be precise.
That makes sense, thank you.

AliSayyah:
`asyncpg.exceptions.UndefinedObjectError: type "doubleprecision" does not exist`

dantownsend:
I think there has to be a space. So with the column type it will have to be something like:

```python
class DoublePrecision(Column):
    @property
    def column_type(self):
        return 'DOUBLE PRECISION'
```

Hopefully that works?

AliSayyah:
> I think there has to be a space. So with the column type it will have to be something like:
> 
> ```python
> class DoublePrecision(Column):
>     @property
>     def column_type(self):
>         return 'DOUBLE PRECISION'
> ```
> 
> Hopefully that works?

Yes, that's what I was looking for. Thank you.

dantownsend:
Fixed by https://github.com/piccolo-orm/piccolo/pull/255

PR #255, opened 2021-09-19
added `Double Precision` column type

AliSayyah:
added `Double Precision` column type based on #241.
`Double Precision` inherits from `Real`.
If this is the wrong approach, please correct me and I'll fix it.


dantownsend:
@AliSayyah This is great - thanks for adding it.

PR #258, opened 2021-09-20
Reflection

AliSayyah:
A simple implementation of reflection. It's far from done obviously. Just wanted to keep you updated on the process and use your guidance along the way.
`TableStorage` is a Singleton object so it can be accessed globally.
`ImmutableDict` is implemented to restrict access to write/delete operations on `tables` attribute of `TableStorage`. 

```python
>>> from piccolo.tablestorage import TableStorage
>>> storage = TableStorage()
>>> await storage.reflect(schema_name="music")
>>> storage.tables  # `Table` objects are accessible from `storage.tables`.

{"music.Band": <class 'Band'>, ... }

>>> Band = storage.tables["music.Band"]
>>> Band.select().run_sync()

```

to-do:

- ~~single table reflection ( needs changes in `schema` app.)~~ Done
- caching
- ~~performance improvements~~ Done

Known problems:

- ~~schema  name doesn't get included in tablename.~~ fixed


piccolo/tablestorage.py
```diff
@@ -0,0 +1,111 @@
...
+            )
+            for table in output_schema.tables
+        ]
+        await asyncio.gather(*add_tables)
+
+    async def _add_table(
+        self,
+        schema_name: str,
+        table: t.Type[Table],
+        if_not_exists: t.Optional[bool] = False,
```
dantownsend, line 92: Do you think we need `if_not_exists` here? If so, perhaps the type should just be `bool`?

piccolo/tablestorage.py
```diff
@@ -0,0 +1,111 @@
...
+
+
+class Immutable(object):
+    def _immutable(self, *arg, **kw) -> TypeError:
+        raise TypeError("%s object is immutable" % self.__class__.__name__)
+
+    __delitem__ = __setitem__ = __setattr__ = _immutable
+
+
+class ImmutableDict(Immutable, dict):
```
dantownsend, line 15: This is cool, but do you think we need it to be immutable, vs a standard dictionary?

If it for the user's safety, in case they somehow mangled this virtual schema, and then inserted garbage data into the database?

dantownsend:
@AliSayyah It's looking good. It's a very interesting feature.

piccolo/tablestorage.py
AliSayyah: I agree. This is not useful in here. This should be handled way before this point to have an at least a performance advantage. Some kind of caching mechanism could solve the problem too. 
The point is to decrease the overhead of creating all the objects from beginning when an update to the `TableStorage` is needed.
If explanation is unclear, please tell me and I'll try to explain in another way.

piccolo/tablestorage.py
AliSayyah: Yes, exactly. Overriding default data types are not a very good idea. But I couldn't find a better way to ensure data safety.
I'll gladly try alternative ways.

dantownsend: I can't think of an alternative. Python has frozen sets, and frozen data classes, but not sure how to make them applicable here.

piccolo/tablestorage.py
dantownsend: I see what you're saying. Maybe it's a useful feature - it depends if people will reflect the same database multiple times.

AbhijithGanesh (changes requested):
Please run linting tests and fix the patch, tests for Python version 3.8 and 3.7 have failed I suppose.

dantownsend:
@AbhijithGanesh This PR is on hold for now - it depends on https://github.com/piccolo-orm/piccolo/pull/260 which we should merge soon.

AbhijithGanesh:
All tests have been passed @dantownsend can this be merged?

AliSayyah:
> All tests have been passed @dantownsend can this be merged?

It's not done yet. I need to write tests for it and some functionalities are still missing. If you are interested in this feature, I'll gladly use your feedback and suggestions.

AbhijithGanesh:
Dan you want someone to write Django testcases or GitHub actions ymls? I haven't got a lot of experience when it comes to the former which'd make it a great Avenue for me to explore.

dantownsend:
@AbhijithGanesh Thanks for the offer, but I don't think we need them as part of this PR.

If you want to experiment with writing tests, you should look at our code coverage and look for areas which are missing:

https://app.codecov.io/gh/piccolo-orm/piccolo

dantownsend:
@AliSayyah It looks good 👍 

Only two things jump to mind:

1. Where to put the code. At the moment it's `piccolo/tablestorage.py`. What do you think about `piccolo/table_reflection.py`? Just trying to think what's the least confusing for someone with no context of the code base. We might want to add a docstring to the top of this file explaining it's intent, something like - `This is an advanced Piccolo feature, which allows runtime reflection of database tables`.
2. We need to add some docs. It might be worth putting these in an advanced section.

AliSayyah:
> @AliSayyah It looks good 
> 
> Only two things jump to mind:
> 
> 1. Where to put the code. At the moment it's `piccolo/tablestorage.py`. What do you think about `piccolo/table_reflection.py`? Just trying to think what's the least confusing for someone with no context of the code base. We might want to add a docstring to the top of this file explaining it's intent, something like - `This is an advanced Piccolo feature, which allows runtime reflection of database tables`.
> 2. We need to add some docs. It might be worth putting these in an advanced section.

That makes sense. I'll change it.
Yeah, I will add the documents later. haven't found the time yet.

piccolo/tablestorage.py
```diff
@@ -0,0 +1,218 @@
...
+        tablenamelist = tablename.split(".")
+        if len(tablenamelist) == 2:
+            return TableNameDetail(
+                name=tablenamelist[1], schema=tablenamelist[0]
+            )
+
+        elif len(tablenamelist) == 1:
+            return TableNameDetail(name=tablenamelist[0], schema="public")
+        else:
+            raise ValueError("Couldn't find schema name.")
```
AliSayyah, line 207: @dantownsend What do you think about this part?
can we handle this more gracefully?

dantownsend: It looks OK to me. I'd have have to test it on a proper schema to be sure, but in theory can't see any issues.

AliSayyah: I'll write a test for this to be more sure.
There is only one case that comes to my mind. If there is a `.` in the name of a table, this will fail.

dantownsend: I was 99% sure that `.` wasn't allowed in table names, but just tried it, and it seems like it's possible.

dantownsend: The same seems to be true of schema names, which complicates matters too.

AliSayyah: Yes, I tried them the other day and that worries me.
We can raise an informative exception when the pattern is different and put `schema_name` as an optional parameter for this case.
This probably won't fit well with the rest of the API though.
A better approach could be to implement schema support directly into the whole API. But I think there will be some problems regarding backward compatibility that needs to be handled to prevent breaking changes.
Was there any technical reason for not separating table_name and schema_name from each other when you started working on piccolo?

dantownsend: When I'm building a web app, I typically just use the default schema, so it never really occurred to me to add support for different schemas.

It might be possible to retrofit it to Piccolo without any breakage with a `schema` argument in `Table`.

We can probably assume that in practice nobody puts a `.` character in either their table names or schema names.

AliSayyah: Yeah, I think that would be a pretty safe bet, but still a bet.
So first we try to use `schema` if specified, else we fall back to splitting `tablename`, right?

dantownsend: Yeah, should work.

AliSayyah: Okay. So I'll leave this as it is and work on this matter on another PR.

dantownsend: Cool, makes sense - thanks

AliSayyah:
@dantownsend I added documentation. but I think there are some grammatical errors.

dantownsend:
@AliSayyah Thanks for adding the docs - don't worry about the grammar, I can tweak it.

dantownsend:
@AliSayyah Thanks for this - it's an interesting feature

PR #260, opened 2021-09-21
Better schema generation

AliSayyah:
1. references from other schemas are supported.

2. now `schema generate` will recursively create reference tables.
For example:
Table `Schema1.A` references Table `Schema2.B`, and Table `Schema2.B` references Table `Schema2.C`. 
now, Table `Schema2.C`  will be created too.

I'm going to add the tests later.

### DIFF
--- a/piccolo/apps/schema/commands/generate.py
+++ b/piccolo/apps/schema/commands/generate.py
@@ -61,6 +61,7 @@ class Constraint:
     constraint_type: Literal["PRIMARY KEY", "UNIQUE", "FOREIGN KEY", "CHECK"]
     constraint_name: str
     column_name: t.Optional[str] = None
+    constraint_schema: t.Optional[str] = None
 
 
 @dataclasses.dataclass
@@ -107,15 +108,16 @@ class TableConstraints:
                 return True
         return False
 
-    def get_foreign_key_constraint_name(self, column_name) -> str:
+    def get_foreign_key_constraint_name(
+        self, column_name
+    ) -> t.Tuple[str, str] | ValueError:  # type: ignore
         for i in self.foreign_key_constraints:
             if i.column_name == column_name:
-                return i.constraint_name
+                return i.constraint_name, i.constraint_schema
 
         raise ValueError("No matching constraint found")
 
 
-@dataclasses.dataclass
 class OutputSchema:
     """
     Represents the schema which will be printed out.
@@ -129,9 +131,10 @@ class OutputSchema:
 
     """
 
-    imports: t.List[str]
-    warnings: t.List[str]
-    tables: t.List[t.Type[Table]]
+    def __init__(self):
+        self.imports: t.List[str] = []
+        self.warnings: t.List[str] = []
+        self.tables: t.List[t.Type[Table]] = []
 
     def get_table_with_name(self, name: str) -> t.Optional[t.Type[Table]]:
         """
@@ -178,7 +181,7 @@ async def get_contraints(
     """
     constraints = await table_class.raw(
         (
-            "SELECT tc.constraint_name, tc.constraint_type, kcu.column_name "  # noqa: E501
+            "SELECT tc.constraint_name, tc.constraint_type, kcu.column_name, tc.constraint_schema "  # noqa: E501
             "FROM information_schema.table_constraints tc "
             "LEFT JOIN information_schema.key_column_usage kcu "
             "ON tc.constraint_name = kcu.constraint_name "
@@ -251,23 +254,115 @@ async def get_table_schema(
 
 
 async def get_foreign_key_reference(
-    table_class: t.Type[Table], constraint_name: str
-) -> t.Optional[str]:
+    table_class: t.Type[Table], constraint_name: str, constraint_schema: str
+) -> t.Tuple[str, str] | None:  # type: ignore
     """
     Retrieve the name of the table that a foreign key is referencing.
     """
     response = await table_class.raw(
         (
-            "SELECT table_name "
+            "SELECT table_name, table_schema "
             "FROM information_schema.constraint_column_usage "
-            "WHERE constraint_name = {};"
+            "WHERE constraint_name = {} AND constraint_schema  = {};"
         ),
         constraint_name,
+        constraint_schema,
     )
     if len(response) > 0:
-        return response[0]["table_name"]
+        return response[0]["table_name"], response[0]["table_schema"]
     else:
         return None
+
+
+def get_table_name(name: str, schema: str) -> str:
+    if schema == "public":
+        return name
+    return f"{schema}.{name}"
+
+
+async def create_table(
+    Schema: t.Type[Table],
+    tablename: str,
+    schema_name: str,
+    output_schema: OutputSchema,
+) -> t.Type[Table]:
+    constraints = await get_contraints(
+        table_class=Schema, tablename=tablename, schema_name=schema_name
+    )
+    table_schema = await get_table_schema(
+        table_class=Schema, tablename=tablename, schema_name=schema_name
+    )
+
+    columns: t.Dict[str, Column] = {}
+
+    for pg_row_meta in table_schema:
+        data_type = pg_row_meta.data_type
+        column_type = COLUMN_TYPE_MAP.get(data_type, None)
+        column_name = pg_row_meta.column_name
+        if not column_type:
+            output_schema.warnings.append(
+                f"{tablename}.{column_name} ['{data_type}']"
+            )
+            column_type = Column
+
+        kwargs: t.Dict[str, t.Any] = {
+            "null": pg_row_meta.is_nullable == "YES",
+            "unique": constraints.is_unique(column_name=column_name),
+        }
+
+        if constraints.is_primary_key(column_name=column_name):
+            kwargs["primary_key"] = True
+            if column_type == Integer:
+                column_type = Serial
+
+        if constraints.is_foreign_key(column_name=column_name):
+            (
+                fk_constraint_name,
+                fk_constraint_schema,
+            ) = constraints.get_foreign_key_constraint_name(
+                column_name=column_name
+            )
+            column_type = ForeignKey
+            (
+                referenced_tablename,
+                refrenced_schemaname,
+            ) = await get_foreign_key_reference(
+                table_class=Schema,
+                constraint_name=fk_constraint_name,
+                constraint_schema=fk_constraint_schema,
+            )
+            if referenced_tablename:
+                kwargs["references"] = await create_table(
+                    Schema,
+                    referenced_tablename,
+                    refrenced_schemaname,
+                    output_schema,
+                )
+            else:
+                kwargs["references"] = ForeignKeyPlaceholder
+
+        output_schema.imports.append(
+            "from piccolo.columns.column_types import " + column_type.__name__
+        )
+
+        if column_type is Varchar:
+            kwargs["length"] = pg_row_meta.character_maximum_length
+
+        column = column_type(**kwargs)
+
+        serialised_params = serialise_params(column._meta.params)
+        for extra_import in serialised_params.extra_imports:
+            output_schema.imports.append(extra_import.__repr__())
+
+        columns[column_name] = column
+
+    table = create_table_class(
+        class_name=_snake_to_camel(tablename),
+        class_kwargs={"tablename": get_table_name(tablename, schema_name)},
+        class_members=columns,
+    )
+    output_schema.tables.append(table)
+    return table
 
 
 async def get_output_schema(schema_name: str = "public") -> OutputSchema:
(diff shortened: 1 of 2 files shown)


piccolo/apps/schema/commands/generate.py
```diff
@@ -107,31 +108,25 @@ def is_foreign_key(self, column_name: str) -> bool:
                 return True
         return False
 
-    def get_foreign_key_constraint_name(self, column_name) -> str:
+    def get_foreign_key_constraint_name(
+        self, column_name
+    ) -> t.Tuple[str, str] | ValueError:  # type: ignore
```
dantownsend, line 113: We can't use the new `|` syntax for union unfortunately , as I don't think there's a back port to Python versions < 3.10.

I don't think exceptions can be put in type annotations - it would be nice if it did though. I think the only way is to put it in the docstring:

```python
:raises: ValueError
```

piccolo/apps/schema/commands/generate.py
```diff
@@ -107,31 +108,25 @@ def is_foreign_key(self, column_name: str) -> bool:
...
         for i in self.foreign_key_constraints:
             if i.column_name == column_name:
-                return i.constraint_name
+                return i.constraint_name, i.constraint_schema
```
dantownsend, line 116: How about returning a dataclass instead of a tuple. Something like:

```
@dataclass
class Constraint:
    name: str
    schema: str
```

I find them a bit nicer than tuples, but up to you.

dantownsend:
@AliSayyah Thanks for this - looks cool.

piccolo/apps/schema/commands/generate.py
AliSayyah: Yes, you are right. I forgot about that. But it's a cool thing though. I'll change it.
Ah, I didn't know that about exceptions. thank you.

piccolo/apps/schema/commands/generate.py
AliSayyah: Very good idea. I'll definitely change this.

piccolo/apps/schema/commands/generate.py
```diff
@@ -107,31 +114,23 @@ def is_foreign_key(self, column_name: str) -> bool:
...
-
     """
 
-    imports: t.List[str]
-    warnings: t.List[str]
-    tables: t.List[t.Type[Table]]
+    def __init__(self):
+        self.imports: t.List[str] = []
+        self.warnings: t.List[str] = []
+        self.tables: t.List[t.Type[Table]] = []
```
dantownsend, line 133: It's possible to keep it as a dataclass, if you just want to add the defaults as lists:

```python
imports: t.List[str] = dataclasses.field(default_factory=list)
```

piccolo/apps/schema/commands/generate.py
```diff
@@ -251,23 +257,111 @@ async def get_table_schema(
...
-            "WHERE constraint_name = {};"
+            "WHERE constraint_name = {} AND constraint_schema  = {};"
         ),
         constraint_name,
+        constraint_schema,
     )
     if len(response) > 0:
-        return response[0]["table_name"]
+        return ConstraintTable(
+            response[0]["table_name"], response[0]["table_schema"]
```
dantownsend, line 276: Might be better to pass these in as kwargs rather than args, so it's clearer.

piccolo/apps/schema/commands/generate.py
```diff
@@ -175,10 +174,14 @@ async def get_contraints(
     :param table_class:
         Any Table subclass - just used to execute raw queries on the database.
 
+    :param tablename - name of the table
+
+    :param schema_name name of the schema
```
dantownsend, line 179: Should be:

```
:param tablename:
    Name of the table.
:param schema_name:
    Name of the schema.
```

dantownsend:
@AliSayyah It's looking good

piccolo/apps/schema/commands/generate.py
AliSayyah: That's right. Thank you.
It's ok now.

AliSayyah:
The latest commit is necessary for the Reflection feature.
This will give an ability to reflect needed tables or exclude ones that are not necessary.

AliSayyah:
https://github.com/piccolo-orm/piccolo/pull/260/commits/561f1c5e554bed8f6388d404b07f662cf5889c59 :
previous result:

> real    0m44.608s
user    0m1.718s
sys     0m0.131s

current result:

> real    0m7.398s
user    0m1.695s
sys     0m0.119s

Tests have been done on a schema with 50 tables.

dantownsend:
@AliSayyah Wow, amazing - great job!

AliSayyah:
Thank you @dantownsend.
I think this PR is done now.
Unless there is something you want me to change.

piccolo/apps/schema/commands/generate.py
```diff
@@ -251,28 +267,134 @@ async def get_table_schema(
...
+
+
+def get_table_name(name: str, schema: str) -> str:
+    if schema == "public":
+        return name
+    return f"{schema}.{name}"
 
 
-async def get_output_schema(schema_name: str = "public") -> OutputSchema:
+async def create_table(
```
dantownsend, line 298: I think the naming of this could be confusing - people might think it actually creates a table in the database.

Perhaps it can be called `create_table_class_from_db`?

piccolo/apps/schema/commands/generate.py
```diff
@@ -251,28 +267,134 @@ async def get_table_schema(
...
+    Schema: t.Type[Table],
```
dantownsend, line 299: I'm not sure about having a capitalised argument - it looks unusual to me.

Maybe `table_class` instead?

piccolo/apps/schema/commands/generate.py
```diff
@@ -251,28 +267,134 @@ async def get_table_schema(
...
+    tablename: str,
+    schema_name: str,
+    output_schema: OutputSchema,
```
dantownsend, line 302: Passing in an object an manipulating can sometimes be troublesome.

How about we don't pass in the `output_schema`, but this function returns a dataclass or tuple containing the table class, any warnings and any import strings. The calling function then updates their `output_schema` from the response.

piccolo/apps/schema/commands/generate.py
```diff
@@ -251,28 +267,134 @@ async def get_table_schema(
...
+    """
+
+    :param schema_name:
+    name of the schema
+    :param tablenames:
+    optional list of table names. Only creates the specifed tables.
+    :param exclude:
+    optional list of table names. excludes the specified tables.
+    :return:
+    OutputSchema
```
dantownsend, line 391: We should try and indent these to make them more readable and consistent with the rest of the code base.

Something like:

```python
    """
    :param schema_name:
        Name of the schema
    :param tablenames:
        Optional list of table names. Only creates the specifed tables.
    :param exclude:
        Optional list of table names. Excludes the specified tables.
    :returns:
        OutputSchema
    """
```

piccolo/apps/schema/commands/generate.py
```diff
@@ -291,91 +413,37 @@ class Schema(Table, db=engine):
...
-    warnings: t.List[str] = []
+    output_schema = OutputSchema()
 
-    for tablename in tablenames:
-        constraints = await get_contraints(
-            table_class=Schema, tablename=tablename, schema_name=schema_name
-        )
-        table_schema = await get_table_schema(
-            table_class=Schema, tablename=tablename, schema_name=schema_name
+    tablecoroutines = (
```
dantownsend, line 421: Maybe `table_coroutines` instead to make it more readable?

piccolo/apps/schema/commands/generate.py
```diff
@@ -291,91 +413,37 @@ class Schema(Table, db=engine):
...
-
-            columns[column_name] = column
-
-        table = create_table_class(
-            class_name=_snake_to_camel(tablename),
-            class_kwargs={"tablename": tablename},
-            class_members=columns,
-        )
-        tables.append(table)
+        for table in tablenames
```
dantownsend, line 428: maybe `tablename` instead of `table` - just to be more explicit.

tests/apps/schema/commands/test_generate.py
```diff
@@ -108,3 +124,65 @@ class Box(Column):
...
+        publication = output_schema.tables[0]
+        writer = output_schema.tables[1]
+        book = output_schema.tables[2]
+        # make sure refrerenced tables have been created
+        self.assertEqual(
+            Publication._meta.tablename, publication._meta.tablename
+        )
+        self.assertEqual(Writer._meta.tablename, writer._meta.tablename)
+
+        # make sure foregin key values are correct.
```
dantownsend, line 186: foreign

dantownsend:
@AliSayyah I've left feedback - mostly just minor style preferences. I'm happy to update these if you like.

AliSayyah:
@dantownsend  I appreciate your feedback. I learned lots of things from your feedbacks.
I know I have a problem with meaningful namings:))
I'll fix all of them. Thank you.

piccolo/apps/schema/commands/generate.py
AliSayyah: @dantownsend I didn't know it can be troublesome. but I can't really understand the reason. If it's not a trouble for you, can you please explain it to me a little bit?
I thought this should be pretty safe because we only use append operations on the lists and not actually manipulating the data inside of them.

dantownsend: Yeah, maybe troublesome is the wrong word. It's more of a preference. It's easier to reason about a function if it's pure - and just returns stuff without any side effects.

I'm happy to keep it as it is.

AliSayyah: Ok now I understand your point and I agree with you. That's more natural. 
I'll change this.

AliSayyah: @dantownsend I'm sorry but I couldn't find a clean way to make this change. I just kept adding to the complexity. I would appreciate your help in this matter.

dantownsend: @AliSayyah how about this - rather than `create_table` returning a table class, have it return an `OutputSchema`.

Then modify `OutputSchema` so it has a custom add method like this:

```python
class OutputSchema:
    ...
    def __add__(self, value) -> OutputSchema:
        self.imports.extend(value.imports)
        self.warnings.extend(value.warnings)
        self.tables.extend(value.tables)
```

So then you can just add the returned `OutputSchema` to the existing one, rather than passing it in as an argument and mutating it within the `create_table`.

With the part  where you're using `asyncio.gather` you can then do:

```python
output_schemas = await asyncio.gather(*tablecoroutines)

output_schema = sum(output_schemas, start=OutputSchema())
```

Does that make sense? It's the cleanest approach I can think of.

AliSayyah: That's a very good idea, thanks.

AliSayyah: The start parameter of the sum function has been added in python 3.8.

So this is my workaround:
```python
    def __radd__(self, value) -> OutputSchema:
        if isinstance(value, int):
            return self
        value.imports.extend(self.imports)
        value.warnings.extend(self.warnings)
        value.tables.extend(self.tables)
        return value

    def __add__(self, value) -> OutputSchema:
        self.imports.extend(value.imports)
        self.warnings.extend(value.warnings)
        self.tables.extend(value.tables)
        return self
```

Now I'm stuck with a bunch of lint errors:))

> piccolo/apps/schema/commands/generate.py:453: error: Item "int" of "Union[Any, int]" has no attribute "tables"
piccolo/apps/schema/commands/generate.py:454: error: Item "int" of "Union[Any, int]" has no attribute "tables"
piccolo/apps/schema/commands/generate.py:456: error: Item "int" of "Union[Any, int]" has no attribute "imports"
piccolo/apps/schema/commands/generate.py:460: error: Item "int" of "Union[Any, int]" has no attribute "imports"
piccolo/apps/schema/commands/generate.py:464: error: Incompatible return value type (got "Union[Any, int]", expected "OutputSchema")


is it ok to ignore these with `# type: ignore`?

dantownsend: Yeah, you can ignore them if you like. Maybe add a type annotation to value, so it's `value: OutputSchema` helps?

It's a shame about the `start` parameter.

AliSayyah: unfortunately, It doesn't help.

dantownsend: `type: ignore` is fine then

AliSayyah:
@dantownsend some interesting insights:

> total time: 7.277s
process time: 1.328s
write time: 0.0002s

the rest of it is the processes(piccolo app) before the `generate` function.

AliSayyah:
> @AliSayyah Thanks a lot for this.

Thanks for your help.

Issue #193, opened 2021-08-25
Schema generation enhancements


dantownsend:
Piccolo now has a command for generating a schema from an existing database.

```bash
piccolo schema generate
```

The current implementation doesn't cover every possible edge case and database feature, so there's room for improvement. For example:

## Column defaults

Getting the defaults for a column, and reflecting them in the Piccolo column. For example, if the column has a default of `1` in the database, we want that to be reflected in the Piccolo schema as:

```python
class MyTable(Table):
    my_column = Integer(default=1)
```

## On Delete / On Update

The database values aren't currently reflected in the ``ForeignKey`` column definitions.

```python
class MyTable(Table):
    my_column = ForeignKey(on_delete=OnDelete.cascade)
```

## Decimal precision

The precision of a decimal column in the database currently isn't reflected in the ``Decimal`` column definitions.

```python
class MyTable(Table):
    my_column = Decimal(digits=(5,2))
```

--------

It's a fairly beginner friendly feature to work on, because even though the code is fairly advanced, it's completely self contained, and doesn't require extensive knowledge of the rest of Piccolo.

wmshort:
I would be willing to take a crack at this, though I can't guarantee a quick turn-around.

dantownsend:
@wmshort That would be great, thanks

AliSayyah:
> WARNING: Unrecognised column types, added `Column` as a placeholder:
  test_table.value_updated ['double precision']

I think this problem can be added to this issue. unless it's the same thing as your `Decimal precision` part of the explanation.
and it says `Column` added as e placeholder. but these columns haven't been included in the generated code. they have been skipped.

dantownsend:
@AliSayyah I'll create a separate issue for that - I think it's because Piccolo doesn't have a `double precision` column type, only `real`.

You're right though, even without supporting that column type, the schema generate command should still add a placeholder value. I'll create a separate issue for that too.

AliSayyah:
> @AliSayyah I'll create a separate issue for that - I think it's because Piccolo doesn't have a `double precision` column type, only `real`.
> 
> You're right though, even without supporting that column type, the schema generate command should still add a placeholder value. I'll create a separate issue for that too.

That's good, thank you.
What is your opinion about adding support for views and materialized views?

dantownsend:
@AliSayyah I think having support for views and materialised views would be a nice feature.

I haven't given much thought into what the API would look like though.

AliSayyah:
[...]

PR #263, opened 2021-09-22
Add defaults in automatic schema generation

wmshort:
Partially addresses #193 .

Initial attempt at default parsing and generation. This PR is mainly intended to kick off review, and changes will almost certainly need to be made. But I need another set of eyeballs on it, to see if I am even on the right track!


wmshort:
Thanks. One open question is handling defaults for Interval columns; how do we get from the postgres string to IntervalCustom?

dantownsend:
@wmshort Yeah, that one looks really tricky. It doesn't look like the easiest thing to parse.

Screenshot 2021-09-22 at 21 26 32

I'm not the best with regex - it might be possible ... but a Python function might be better. Something like:

```python
date, time = default.rsplit(' ', 1)
hour, minute, second = [int(i) for i in time.split(':')]
```

Once we've parsed it, it's easy enough to create the `IntervalCustom` - we just need to instantiate it with the relevant days / hours etc.

wmshort:
In the postgres documentation I am not seeing the version where digital times are mixed with phrasal years / months / days. The examples in the docs are all like:

5 weeks 4 days 3 hours 2 minutes 1 second

I think I have a workable regex now for this fully phrasal type, but does it need to handle both kinds? Thus equivalent:

5 weeks 4 days 3 hours ago
5 weeks 4 days 03:00:00 ago

I guess it can't hurt...

piccolo/apps/schema/commands/generate.py
```diff
@@ -174,15 +174,17 @@ def get_table_with_name(self, name: str) -> t.Optional[t.Type[Table]]:
...
     Bytea: re.compile(r"'(.*)'::bytea$"),
-    DoublePrecision: re.compile(r"[+-]?([0-9]*[.])?[0-9]+"),
+    DoublePrecision: re.compile(r"([+-]?(?:[0-9]*[.])?[0-9]+)"),
     Varchar: re.compile(r"^'(.*)'::character varying$"),
     Date: re.compile(r"^((?:\d{4}-\d{2}-\d{2})|CURRENT_DATE)$"),
     Integer: re.compile(r"^(-?\d+)$"),
-    Interval: re.compile(r"^'(.*)'::interval"),
-    JSON: re.compile(r"^'(.*)'::json$"),
+    Interval: re.compile(
+        r"^(?:')?(?:(?:(?P<years>\d+) y(?:ear(?:s)?)?\b)|(?:(?P<months>\d+) m(?:onth(?:s)?)?\b)|(?:(?P<weeks>\d+) w(?:eek(?:s)?)?\b)|(?:(?P<days>\d+) d(?:ay(?:s)?)?\b)|(?:(?:(?:(?P<hours>\d+) h(?:our(?:s)?)?\b)|(?:(?P<minutes>\d+) m(?:inute(?:s)?)?\b)|(?:(?P<seconds>\d+) s(?:econd(?:s)?)?\b))|(?:(?P<digits>-?\d\d:\d\d:\d\d))?\b))+(?P<direction>ago)?(?:'::interval)?$"
```
dantownsend, line 182: That's an impressive regex!

wmshort: It might be overkill, but it works...

wmshort:
One more thing: callables as defaults. The Serial column utilizes an internal postgres function, nextval, to evaluate the next in the sequence. This is not callable in the same way as, say, UUID4. Is it even meant to be settable by the user?

dantownsend:
That's a good point - if the user hardcoded the default for a Serial column to something like `5`, I'm not sure if Postgres would reject that, but wouldn't be good.

dantownsend:
With the default values in Piccolo, it tries to mirror them as much as possible, so it exists in Python and the database.

So with UUID, if someone bypassed the ORM and inserted straight into the database, it would still behave itself.

There are some situations where the default can only exist in the database (e.g. Serial), or in Python (some custom user function).

wmshort:
> With the default values in Piccolo, it tries to mirror them as much as possible, so it exists in Python and the database.
> 
> So with UUID, if someone bypassed the ORM and inserted straight into the database, it would still behave itself.
> 
> There are some situations where the default can only exist in the database (e.g. Serial), or in Python (some custom user function).

So for Serial we probably just want to omit the reflection of the default (nextval etc) in the generated schema -- i.e. just output column_name = Serial() -- as the mechanism is all handled internally. Or...?

dantownsend:
@wmshort That's right, yeah - if it's `Serial` either omit the default argument, or set it to `None` or `...`.

dantownsend:
@wmshort It's looking great - the tests failed due to a type error. It's because when I originally wrote the code I didn't add a type annotation for `COLUMN_TYPE_MAP`, so it's my mistake.

This should fix it:

```python
COLUMN_TYPE_MAP: t.Dict[str, t.Type[Column]] = {
```

wmshort:
Sorry about that last one, I had moved on to the decimal precision and forgot to remove extraneous code :) should be all set now, though.

Issue #252, opened 2021-09-19
Add an API for upserting


dantownsend:
Based on this discussion: https://github.com/piccolo-orm/piccolo/discussions/248

We don't currently have an API for upserting.

This would be based around the `ON CONFLICT` clause you can add to `INSERT` queries.

https://www.postgresql.org/docs/current/sql-insert.html#SQL-ON-CONFLICT

There will probably be two APIs needed. Something for `insert` queries, and a convenience method for `objects` queries.

For example:

```python
# For insert
Band.insert(Band(name='Pythonistas')).on_conflict(...).run_sync()

# For objects
Band.objects().upsert(Band.name == 'Pythonistas', {Band.popularity: 5000}).run_sync()
```

AbhijithGanesh:
I would like to work on this if It is open and unassigned.

dantownsend:
@AbhijithGanesh Great, thanks. This might be a tricky issue though, just so you're aware!

AbhijithGanesh:
Thanks for the heads up , I'll read upon docs and get this reviewed before it gets merged

AbhijithGanesh:
Clarification: The upserting method/process is like HTTP PUT right?

AbhijithGanesh:
This requires more interaction with the maintainer than I thought, where can I reach out to you? @dantownsend

AbhijithGanesh:
> 
> 
> Based on this discussion: #248
> 
> We don't currently have an API for upserting.
> 
> This would be based around the `ON CONFLICT` clause you can add to `INSERT` queries.
> 
> https://www.postgresql.org/docs/current/sql-insert.html#SQL-ON-CONFLICT
> 
> There will probably be two APIs needed. Something for `insert` queries, and a convenience method for `objects` queries.
> 
> For example:
> 
> ```python
> # For insert
> Band.insert(Band(name='Pythonistas')).on_conflict(...).run_sync()
> 
> # For objects
> Band.objects().upsert(Band.name == 'Pythonistas', {Band.popularity: 5000}).run_sync()
> ```

So I am implementing the insert part at query>insert.py and upsert as a new file.
Would that be okay?

dantownsend:
You're best reaching me here - I don't use much else besides email.. You can start a separate discussion thread if you prefer.

That's right - an UPSERT is a bit like a PUT. When an insert fails due to a unique constraint, it will update the existing row instead of raising an error.

Here are the docs:

[...]

PR #269, opened 2021-09-23
Upserting

AbhijithGanesh:
Attempts to create an upserting method #252 . I have tried using update, insert and exist method. Please review @dantownsend

### DIFF
--- a/.github/ISSUE_TEMPLATES/features.yml
+++ b/.github/ISSUE_TEMPLATES/features.yml
@@ -1,0 +1,34 @@
+name: Features💡
+description: Suggest us a new feature
+title: "[Feature]: Describe your Feature/Suggestion here"
+labels: [goal:addition]
+body:
+  - type: markdown
+    attributes:
+      value: |
+        Thanks for taking the time to fill out this suggestion form!
+  - type: input
+    attributes:
+      label: Suggestion
+      description: Can you please elaborate about your suggestion?
+      placeholder: I would like to see...
+    validations:
+      required: true
+  
+  - type: textarea
+    id: what-happened
+    attributes:
+      label: What is expected?
+      description: Also tell us, what did you wish to see ?
+      placeholder: Give us specific details .
+      value: "Details!"
+    validations:
+      required: true
+  
+  - type: textarea
+    id: multimedia
+    attributes:
+      label: Relevant log output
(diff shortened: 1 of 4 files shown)


AbhijithGanesh:
I committed the template files by mistakes, please excuse.

AbhijithGanesh:
please review @dantownsend

dantownsend:
@AbhijithGanesh It's a nice start. You should focus on creating the `OnConflict` delegate next.

Rather than having an `Upsert` class, we should try and encapsulate it into the existing `Insert` class.

So we just do:

```python
MyTable.insert(MyTable(name="bob"), on_conflict=Conflict.do_nothing).run_sync()
```

AbhijithGanesh:
Thank you, I misinterpreted on how to implement this.I shall implement an onconflict delegeate

AbhijithGanesh:
I shall send in a delegate with On_Conflict, would that solve the entire issue (as a standalone) or would it implement upsert class?

dantownsend:
So we need something like this in `mixins.py`:

```python
class Conflict(Enum):
    do_nothing = 'DO NOTHING'
    do_update = 'DO UPDATE'
    

@dataclass
class OnConflict:

    _on_conflict: t.Optional[Conflict] = None

    def on_conflict(self, conflict: Conflict):
        self._on_conflict = conflict

```

And then in `Insert`:

```python
class Insert(Query):
    __slots__ = ("add_delegate", "on_conflict_delegate")

    def __init__(self, table: t.Type[Table], *instances: Table, on_conflict: t.Optional[Conflict] = None, **kwargs):
        super().__init__(table, **kwargs)
        self.add_delegate = AddDelegate()
        self.on_conflict_delegate = OnConflictDelegate()
        self.on_conflict(conflict)
        self.add(*instances)

    def on_conflict(self, conflict: Conflict) -> Insert:
        self.on_conflict_delgate.on_conflict(conflict)
        return self
```

That gives the user the ability to tell us what they want to do when a conflict occurs.

And then in the `postgres_querystrings` and `sqlite_querystrings` methods of `Insert`, we can alter the query accordingly.

AbhijithGanesh:
I have attempted to create it again, I know it is kind of slow progress but I am really grateful for the mentoring @dantownsend I will make changes if there are any to be made. Also I'd like to know if there's specific documentation about features related to methods.

PR #275, opened 2021-09-25
capture decimal precision for schema generation

wmshort:
Partially addresses #193 by adding decimal precision to schema generation

### DIFF
--- a/piccolo/apps/schema/commands/generate.py
+++ b/piccolo/apps/schema/commands/generate.py
@@ -63,7 +63,10 @@ class RowMeta:
     table_name: str
     character_maximum_length: t.Optional[int]
     data_type: str
-
+    numeric_precision: t.Optional[t.Union[int, str]]
+    numeric_scale: t.Optional[t.Union[int, str]]
+    numeric_precision_radix: t.Optional[Literal[2, 10]]
+    
     @classmethod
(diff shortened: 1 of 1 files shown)


piccolo/apps/schema/commands/generate.py
```diff
@@ -63,7 +63,10 @@ class RowMeta:
     table_name: str
     character_maximum_length: t.Optional[int]
     data_type: str
-
+    numeric_precision: t.Optional[t.Union[int, str]]
+    numeric_scale: t.Optional[t.Union[int, str]]
```
dantownsend, line 67: Just out of interest, under what situations do we get a string back instead of an int?

I'm just trying to understand what edge cases there are.

wmshort: @dantownsend Good question. Oddly, I cannot trace this now in the docs, but I was operating under the impression that the values for ``precision`` and ``scale`` could sometimes be stored internally as a binary string -- which accounts for that extra cast to ``int``. Happy to remove . . .

dantownsend: No, I'm happy to keep it - was just curious.

dantownsend:
Looks good to me 👍

PR #279, opened 2021-09-27
Inline Variables that immediately returned & Refactor Multipart of Codes ✨

yezz123:
During this week, I work with Piccolo to build a new Project, I found the code that is great & Friendly, also Why not helping the team to build a clean Code 👨‍💻 , so I start reading the code and refactor it.

- I found that the multiple things are the none inlined Variable: in the case when the variable being declared is immediately returned.
- Also the issue relates to Moves statements or the *Hoist statements out of if*.
- Other issues relate to using `any()` and other ones ...

### DIFF
--- a/piccolo/apps/asgi/commands/new.py
+++ b/piccolo/apps/asgi/commands/new.py
@@ -18,9 +18,7 @@ def print_instruction(message: str):
 
 
 def get_options_string(options: t.List[str]):
-    return ", ".join(
-        [f"{name} [{index}]" for index, name in enumerate(options)]
-    )
+    return ", ".join(f"{name} [{index}]" for index, name in enumerate(options))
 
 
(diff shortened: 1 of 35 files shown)


dantownsend:
@yezz123 Thanks for this - took me a while to go through it, but it all makes sense, and looks great.

You must have a super linter, or else you have super linting skills!

yezz123:
> @yezz123 Thanks for this - took me a while to go through it, but it all makes sense and looks great.
> 
> You must have a super linter, or else you have super linting skills!

@dantownsend Just Documentation & some previous experience in Linting and Clean Code 😅

AliSayyah:
@yezz123 wow good job!

pinchXOXO:
@yezz123 This is a Great PR! 🚀

PR #280, opened 2021-09-27
sample basic migrations ui updates

wmshort:


dantownsend:
@wmshort Thanks for this - it looks so much better.

Here's a before and after:

Screenshot 2021-09-28 at 08 46 50
Screenshot 2021-09-28 at 08 46 19

wmshort:
More can be done with this, but this provides a first coat of new paint...

AliSayyah:
@wmshort This is much better. Thank you.

wmshort:
@dantownsend Take a took at the UI when there are forward migrations to run or when migrations are reversed. If anyone has suggestions for further refinements please make them. I have tried to keep things simple yet informative

dantownsend:
@wmshort  I like it. On my machine the warning symbol (exclamation symbol in triangle) is rendered quite small. I'm not sure why Mac OS renders some emoji more plainly than others. I don't know if there's any solution.

Screenshot 2021-09-28 at 16 54 37

dantownsend:
@wmshort Thanks - they look better now.

Screenshot 2021-09-29 at 20 37 05

dantownsend:
@wmshort I have one suggestion - can we change the information symbol to the one shown below?

In the weird world of emoji the information symbol is two characters wide, vs one character wide for the others, so they don't line up.

Screenshot 2021-09-29 at 21 06 49

If you're happy with this, I can push the change.

wmshort:
Definitely, that does look better! I thought it was just an anomaly with that circle-i 'information' symbol.

dantownsend:
@wmshort I was confused by it - when you do `len('ℹ️')` in Python the answer is 2 ...

I'll just wait for the build to pass then will merge it in. Thanks for this - it's so much better than before!

PR #278, opened 2021-09-26
add index reflection in schema generation

wmshort:

PR #282, opened 2021-09-30
include index info in schema generation (replaces #278)

wmshort:
@dantownsend Somehow I managed to royally botch a commit and this has screwed with the tests. Starting the PR from scratch seemed easier.

### DIFF
--- a/piccolo/apps/schema/commands/generate.py
+++ b/piccolo/apps/schema/commands/generate.py
@@ -36,6 +36,7 @@ from piccolo.columns.column_types import (
     Varchar,
 )
 from piccolo.columns.defaults.interval import IntervalCustom
+from piccolo.columns.indexes import IndexMethod
 from piccolo.engine.finder import engine_finder
 from piccolo.engine.postgres import PostgresEngine
 from piccolo.table import Table, create_table_class, sort_table_classes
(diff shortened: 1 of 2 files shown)


dantownsend:
@wmshort No worries - if there's anything I can help with let me know.

wmshort:
> @wmshort No worries - if there's anything I can help with let me know.

Can you check the test logs? I can't for the life of me figure out what I have done to screw things up. Sorry :/

dantownsend:
@wmshort No problem, I'll investigate.

dantownsend:
@wmshort  This one is a bit of a head scratcher. Sorry it took so long. I worked out why the tests were failing.

The unique value for the column was being set in two separate places, and they were conflicting.

https://github.com/piccolo-orm/piccolo/blob/7c086af5fab955b6c633833c5e3b93503d49f697/piccolo/apps/schema/commands/generate.py#L651

https://github.com/piccolo-orm/piccolo/blob/7c086af5fab955b6c633833c5e3b93503d49f697/piccolo/apps/schema/commands/generate.py#L657

It's because there are unique indexes, and unique constraints - it's possible for a column to have one and not the other.

I let the unique constraint take precedence for now, but might merit more investigation.

wmshort:
@dantownsend good catch and I think your solution is the right one. The StackOverflow answers didn't really clarify things for me but the docs seem clear enough: the unique index is consequent on the unique constraint -- it is created automatically for the unique constraint. I guess this means you could have a unique index that is independent of any unique constaint but if you have a unique constraint or PK then you will have a unique index on it as the enforcement mechanism.

dantownsend:
@wmshort Cool, I'm happy with it then. I just added a quick unit test.

@AliSayyah Do you have any thoughts on this, as you did a lot of the reflection stuff?

PR #287, opened 2021-10-03
Changing Piccolo apps example to reflect its name/docstring

gpkc:


dantownsend:
@gpkc Thanks for this - good catch!

Issue #299, opened 2021-10-11
Update GitHub actions for Postgres 14, and Python 3.10


dantownsend:
Python 3.10 and Postgres 14 have both been released.

PR #300, opened 2021-10-11
Add Postgres 14

dantownsend:
Runs OK locally - just depends if GitHub Actions supports it.


dantownsend:
Partially implements https://github.com/piccolo-orm/piccolo/issues/299

Issue #297, opened 2021-10-11
Reflection on a table with self-referencing ForeignKey does not complete


knguyen5:
When running a reflection on a table that includes a self-referencing ForeignKey, such as the example below, the process hangs or never completes:
```python
class Musician(Table):
    name = Varchar(length=100)
    instructor = ForeignKey(references='self')
```

After removing the foreign key or replacing with another table reference, the reflection completes as expected.

knguyen5:
Thanks @AliSayyah! The scenario that I'm trying to capture in this data modeling is when there's an unknown number of parent entities and child entities. To keep the schema flexible, it's kept in a long table format. 

For example, this could be useful in a growing organization tree and mapping out which employee reports to another employee (whom might report to someone else). 

I'm no data modeling expert, so I'm very open to other models that might cover this type of scenario.

dantownsend:
I've used recursive foreign keys a few times in the past.

Things like this are quite common:

```python
class Employee(Table):
    name = Varchar(length=100)
    manager = ForeignKey(references='self')
```

It should be possible to support this in `piccolo schema generate`. It's just an edge case we overlooked.

PR #305, opened 2021-10-14
correctly reflect self-referencing FKs

wmshort:
Addresses issue #297 . I haven't been able to test this yet but this is the idea!

### DIFF
--- a/piccolo/apps/schema/commands/generate.py
+++ b/piccolo/apps/schema/commands/generate.py
@@ -687,7 +687,7 @@ async def create_table_class_from_db(
             )
             if constraint_table.name:
                 referenced_table: t.Union[str, t.Optional[t.Type[Table]]]
-                
+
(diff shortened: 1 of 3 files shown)


wmshort:
> @wmshort Thanks for this - the solution is also what I had in mind.

@dantownsend this checks out OK now, but does still need specialized testing

wmshort:
This is turning out to be a bit more complicated than I originally thought. Since ``generate`` relies on the Table and Column constructors the reference to ``self`` after reading the schema from the DB is automatically converted back to a table name, whereas in fact we want to preserve it as a string. I'll keep trying to untangle things.

wmshort:
@dantownsend This appears to work correctly now in the generated schema reflection, but I'm not sure the right way to test it. The unittest I added isn't right: maybe ``alter().add_foreign_key_constraint``?

Issue #306, opened 2021-10-15
[BUG]: Decimal import conflict in auto generated migration


0scarB:
Auto-generating migrations for tables with columns of type `Decimal` causes conflicting imports <br>
```python
from piccolo.columns.column_types import Decimal
```
and 
```python
from decimal import Decimal
```
<br>

This means that the following code in the migration fails
```python
manager.add_column(
        ...
        column_class_name="Decimal",
        column_class=Decimal,  # Wrong `Decimal` - is `decimal.Decimal`, should be `piccolo.columns.column_types.Decimal`!!!
        params={
            "default": Decimal("0"),  # Correct `Decimal` - is `decimal.Decimal`
            "digits": None,
            "null": False,
            "primary_key": False,
            "unique": False,
            "index": False,
            "index_method": IndexMethod.btree,
            "choices": None,
            "db_column_name": None,
        },
    )
```

because `Decimal` has type `decimal.Decimal`, not `piccolo.columns.column_types.Decimal`. <br><br>

This probably hasn't been discovered yet because most people use `Numeric` which does not have this problem.

## Quick Fix
Use an absolute import or import alias in generation.

## Extensive Fix
Although this certainly is an edge case, similar import conflicts may occur in the future. To future-proof new code one could, for example, check for conflicts during generation and fall back on absolute imports.

**Pro:** Future-proof
**Con:** Extra code-complexity

---
**piccolo version** == 0.57.0

## Full example
**broken_decimal/tables.py**
```python
from piccolo.table import Table
from piccolo.columns import Decimal, Numeric


class BrokenDecimalAutoMigration(Table):
    # This breaks migration auto generation because of `Decimal` import conflict.
    dec_col = Decimal()

    # This works
    # dec_col = Numeric()
```
<br>

`piccolo migrations new broken_decimal --auto` generates **broken_decimal/piccolo_migrations/2021-10-15T19-10-02-685456.py**
```python
from piccolo.apps.migrations.auto import MigrationManager
from decimal import Decimal
from piccolo.columns.column_types import Decimal
from piccolo.columns.indexes import IndexMethod


ID = "2021-10-15T19:10:02:685456"
VERSION = "0.57.0"
DESCRIPTION = ""


async def forwards():
    manager = MigrationManager(
        migration_id=ID, app_name="broken_decimal", description=DESCRIPTION
    )

    manager.add_table(
        "BrokenDecimalAutoMigration", tablename="broken_decimal_auto_migration"
    )

[...]

PR #307, opened 2021-10-19
Fix auto migration import conflict2

0scarB:
This PR attempts to fix conflicting/clashing globals in migration auto-generation (#306).

To achieve this a the `UniqueGlobalNames` class has been added to [serialisation.py](https://github.com/0scarB/piccolo/blob/3ef88822edcd4c631006a90ea49c52415c04b845/piccolo/apps/migrations/auto/serialisation.py#L127-L179). This class is meant to store global names as class attributes. The serialisation logic has been rewritten to check global names against the class and raise errors if there is a conflict.

This approach should help future developers to write conflict free serialisation code. This however does come at the cost of readability is some places:

**Pros:** 
- It is much harder to add conflicts to serialisation logic
- Globals for column types are automatically discovered (in `UniqueGlobalNamesMeta`) and added to `UniqueGlobalNames` meaning that future conflicts may found without any developer intervention.

**Cons:**
- Developer must add new globals to `UniqueGlobalNames` class
- Accessing `UniqueGlobalNames.COLUMN_VARCHAR` is less readable that simply using string `"Varchar"`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -5,9 +5,3 @@ target-version = ['py37', 'py38', 'py39']
 [tool.isort]
 profile = "black"
 line_length = 79
-
-[tool.pytest.ini_options]
-filterwarnings = [
-    "error::piccolo.apps.migrations.auto.serialisation.UniqueGlobalNameConflictWarning",
-]
-


dantownsend:
Thanks for making those changes.

The tests are failing for some reason - it says the minimum coverage % hasn't been reached, but when I run them locally it's fine ... weird.

0scarB:
Sorry for not getting back to you on the code coverage issue - been a little busy lately. Removing the `filterwarnings` is totally fine : ). You'd see the warning when running pytest anyway

dantownsend:
@0scarB No worries - thanks!

Issue #321, opened 2021-11-03
Add `include_columns` option to `create_pydantic_model`


dantownsend:
`create_pydantic_model` has an `exclude_columns` argument, but it would be very useful to also have an `include_columns` argument.

If neither are specified, all columns are returned. If both are specified, then an exception is raised.

```python
class Band(Table):
    name = Varchar()
    popularity = Integer()
    # Imagine it has lots more columns ...

model = create_pydantic_model(include_columns=(Band.name,))
```

dantownsend:
Fixed by https://github.com/piccolo-orm/piccolo/pull/324

PR #324, opened 2021-11-04
added include columns

sinisaos:
Related to #321


dantownsend:
@sinisaos Thanks for this

PR #340, opened 2021-11-12
Better tests for asgi new

AliSayyah:
this seems to do the trick but I have no idea If this is a good solution. 
I would appreciate your ideas.
Also, I think it would be better to run the tests with different combinations of `ROUTERS` and `SERVERS`.


yezz123 (approved):
I like this one LGTM+1 ✨🚀

yezz123 (approved):

tests/apps/asgi/commands/test_new.py
```diff
@@ -1,28 +1,92 @@
...
+
+                    venv_path = os.path.join(root, "venv")
+                    os.mkdir(venv_path)
+                    os.system(
+                        f"#!/bin/bash &&"
+                        f"cd {root} && virtualenv venv &&"
+                        f" source {venv_path}/bin/activate &&"
+                        f" pip install -r requirements.txt &&"
+                        f" python main.py"
+                    )
```
dantownsend, line 92: This looks great, thanks!

I wonder what happens here with `os.system`? Does the sub process just get killed off when the tests finish? I wonder if there's some way of getting a status code back from it?

AliSayyah: yes, the subprocess ends and I think it's possible to get the status code.
when everything is okay, it returns `0`. 
but try/except doesn't work with `os.system`.
using `subprocess` module gives more flexibility.
if you have anything in particular in your mind, I can try it out.

yezz123: @dantownsend for me looks good cause here if we use the try/except we could get multi issues when we debug, the idea that Ali provide using a subprocess is amazing.

AliSayyah:
I'm trying another way that's more testable. I'll share it with you.

dantownsend:
@AliSayyah Thanks - it's a tricky one. `python main.py` launches a web server, so won't immediately return a response code. We might need to do something like 'did it return a response code of `1` in under 1 second', otherwise treat it as a pass.

AliSayyah:
@dantownsend 
I was planning on going with this exact approach. but I didn't quite like the timeout part. It doesn't seem very concrete. but if I can't come up with anything else, that's the plan.

dantownsend:
@AliSayyah Sorry if you are getting a bunch of email spam about this PR - took me ages to get the build passing!

I took what you did and made some minor changes:

 * Using `subprocess.run` instead of `os.system`. It is a bit easier to get the result code from the sub process.
 * Added a dummy ASGI server. Uvicorn was a pain - I couldn't figure out how to test it.

I also separated this test out as an integration test, because it takes a while to run.

What do you think?

yezz123 (approved):
Even I got more than 10 emails 😂 but no problem, its look like a good fix and this gonna improve the tests for the ASGI Server 🪨 

LGTM 🚀

dantownsend:
@yezz123 Haha, sorry about that!

Thankfully I got there on the 10th attempt, and not the 100th!

AliSayyah:
@dantownsend Thank you for finishing this. It looks great!
And I'm sorry that I couldn't finish it.
I was busy with midterm exams:))

dantownsend:
@AliSayyah No problem - hope your exams go well!

PR #346, opened 2021-11-13
CI: Add Pre-commit configuration 🐰 

yezz123:
based on this discussion https://github.com/piccolo-orm/piccolo/discussions/343


dantownsend:
@yezz123 Thanks for adding this.

Black recently added a feature where having a trailing comma forced it onto a new line:

```python
data = (1,2,3,)

# Is formatted to:
data = (
    1,
    2,
    3,
)
```

Which version of Black are you on? I think if you use the latest version, the formatting will be different.

We encountered this issue in Piccolo API recently, and were confused why Black was giving inconsistent results (it was down to this trailing comma change).

yezz123:
> @yezz123 Thanks for adding this.
> 
> Black recently added a feature where having a trailing comma forced it onto a new line:
> 
> ```python
> data = (1,2,3,)
> 
> # Is formatted to:
> data = (
>     1,
>     2,
>     3,
> )
> ```
> 
> Which version of Black are you on? I think if you use the latest version, the formatting will be different.
> 
> We encountered this issue in Piccolo API recently and were confused why Black was giving inconsistent results (it was down to this trailing comma change).

I use the `19.10b0` version and the py version was `py3.8`

dantownsend:
@yezz123 I think the change was in around version 20. If you use try the latest version, the diff will be smaller.

yezz123:
> @yezz123 I think the change was in around version 20. If you use to try the latest version, the diff will be smaller.

okay I will fix it

yezz123:
@dantownsend could you check now, I pull to the last version of black

.pre-commit-config.yaml
```diff
@@ -0,0 +1,39 @@
...
+      hooks:
+          - id: check-merge-conflict
+          - id: check-added-large-files
+          - id: check-ast
+          - id: check-symlinks
+          - id: trailing-whitespace
+          - id: check-json
+          - id: debug-statements
+          - id: pretty-format-json
+            args: ["--autofix", "--allow-missing-credentials"]
```
dantownsend, line 13: Thanks for this - I need to read up a bit more on pre-commit to work out which we need. We might be able to drop the JSON ones.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,39 @@
...
+            exclude: docs/
+            args:
+                - --select=F403,F406,F821,T003
+    - repo: https://github.com/humitos/mirrors-autoflake
+      rev: v1.3
+      hooks:
+          - id: autoflake
+            files: '\.py$'
+            exclude: '^\..*'
+            args: ["--in-place"]
```
dantownsend, line 34: How useful do you find this one? I assume the other linters would complain if there are unused imports, and it only seems to detect unused standard library imports, right?

.pre-commit-config.yaml
yezz123: Okay, i will drop the JSON one

.pre-commit-config.yaml
yezz123: the `mirrrors-autoflake` work on deleting the standards and also if we import another package that its not used in the code source

PR #349, opened 2021-11-14
added `Column._equals` method

dantownsend:
Added an `_equals` method to `Column`.

There wasn't a centralised piece of logic for comparing columns, which is why it was added.


yezz123 (approved):
overall LGTM+1, but the test need more improvement, that's gonna help with adding some cases of methods fails 👌🏻

dantownsend:
@yezz123 You're totally right - I added a few more tests.

PR #366, opened 2021-11-30
add sqlite connection pool warning to Engine

gmos:
Partly solves: https://github.com/piccolo-orm/pymdb/issues/1


dantownsend:
@gmos Thanks a lot of this.

I made two minor changes, because I'm an obsessive ...

 * Added an assertion to the test
 * Used the temp directory for the sqlite file, so it doesn't clutter up the file system

PR #375, opened 2021-12-17
M2M prototype

dantownsend:
An initial prototype for M2M (many to many). It still needs a bunch of testing, but it's a start.

```python
class Band(Table):
    name = Varchar()
    genres = M2M(
        LazyTableReference("GenreToBand", module_path=__name__)
    )


class Genre(Table):
    name = Varchar()


# The joining table:
class GenreToBand(Table):
    band = ForeignKey(Band)
    genre = ForeignKey(Genre)
```

We have a new class called `M2M`. You reference a joining table. By explicitly defining the joining table it has the following benefits:

 * Simplifies migrations, as we don't have to magically create new tables.
 * Makes it easier to retrofit Piccolo onto an existing schema.
 * Can add additional columns to the joining table if we want (see below).

```python
class GenreToBand(Table):
    band = ForeignKey(Band)
    genre = ForeignKey(Genre)
    created_on = Timestamp()
    reasoning = Text()
```

With `M2M` you can then do queries like this:

```python
>>> await Band.select(Band.name, Band.genres(Genre.name))
[
    {"name": "Pythonistas", "genres": ["Rock", "Folk"]},
    {"name": "Rustaceans", "genres": ["Folk"]},
    {"name": "C-Sharps", "genres": ["Rock", "Classical"]},
]
```


sinisaos:
@dantownsend This looks interesting and promising. I like explicitly define a joining table and all the benefits you described that come with it. Queries also look great and they are easy to understand. I haven't had a chance to try your new code yet, but I'll play around with it as soon as possible.

dantownsend:
@sinisaos Thanks. There's probably still lots of bugs, but it's getting there.

piccolo/table.py
```diff
@@ -407,6 +416,52 @@ def get_related(self, foreign_key: t.Union[ForeignKey, str]) -> Objects:
...
+    # TODO - I might merge this with ``get_related``.
+    def get_m2m(self, m2m: M2M) -> M2MGetRelated:
+        """
+        Get all matching rows via the join table.
+
+        .. code-block:: python
+
+            >>> band = await Band.objects().get(name="Pythonistas")
+            >>> await band.get_m2m(Band.genres)
+            [<Genre: 1>, <Genre: 2>]
```
sinisaos, line 428: Minor typo here it should be `await Band.objects().get(Band.name == "Pythonistas")` and for `line 428` I get result like this
`[<Genre: "id" SERIAL PRIMARY KEY NOT NULL>, <Genre: "id" SERIAL PRIMARY KEY NOT NULL>]`

piccolo/table.py
```diff
@@ -407,6 +416,52 @@ def get_related(self, foreign_key: t.Union[ForeignKey, str]) -> Objects:
...
+        an entry into the joining table.
+
+        .. code-block:: python
+
+            >>> band = await Band.objects().get(name="Pythonistas")
+            >>> await band.add_m2m(
+            >>>     Genre(name="Punk rock"),
+            >>>     m2m=Band.genres
+            >>> )
+            <Genre: 5>
```
sinisaos, line 450: Same here. Minor typo here it should be `await Band.objects().get(Band.name == "Pythonistas")` and for `line 450` I get result like this `[{'id': 5}]`

sinisaos (commented):
@dantownsend I gave two comments on minor typos in the query. I also get different query results. Maybe I'm doing something wrong?

piccolo/table.py
dantownsend: Good catch. I've fixed the typo.

The response should actually be `[<Genre: 1>, <Genre: 2>]` - this is a bug. I'll fix it.

piccolo/table.py
dantownsend: Thanks, fixed.

dantownsend:
> @dantownsend I gave two comments on minor typos in the query. I also get different query results. Maybe I'm doing something wrong?

@sinisaos Thanks for testing it. When you get different query results, was it just that `get_m2m` was returning something like `[{"id": 1}]`, or some other issue?

piccolo/table.py
sinisaos: Great. When you changed `__repr__` method in `table.py` I get correct result `[<Genre: 1>, <Genre: 2>]`

dantownsend: Cool, thanks

sinisaos:
@dantownsend No issue. Just a different result. I get `["id": 1}]`, not `<Genre: 1>` which I think is nice because it's the same as adding a normal record.

dantownsend:
@sinisaos OK, cool - thanks.

Next I'm going to see if I can get it working for SQLite.

Are there any useful features you can think of which need adding for this initial M2M version?

sinisaos:
@dantownsend No. This is great and I think you’ve covered everything you need for `M2M` to work properly.

piccolo/columns/m2m.py
```diff
@@ -198,7 +198,43 @@ async def run(self):
...
+
+            row_id = getattr(row, row._meta.primary_key._meta.name)
+            if row_id:
+                row_ids.append(row_id)
+
+        if row_ids:
+            joining_table = self.m2m._meta.table
+            return await joining_table.delete().where(
+                joining_table._meta.primary_key.is_in(row_ids)
+            )
```
sinisaos, line 235: This doesn't look right. We should delete a row from `Genre` table like this.
```python
return await related_table.delete().where(
    related_table._meta.primary_key.is_in(row_ids)
)
```
but the problem is that we will delete every `genre` E.g `punk rock` from every band that contains that genre in the M2M list and that's not right. Sorry if i'm wrong

dantownsend: @sinisaos Wow, yes - you're right! I'll fix that. 🤦

dantownsend: Should be fixed now

piccolo/columns/m2m.py
```diff
@@ -0,0 +1,336 @@
...
+                "LazyTableReference instance."
+            )
+
+    @property
+    def foreign_key_columns(self) -> t.List[ForeignKey]:
+        if not self._foreign_key_columns:
+            self._foreign_key_columns = (
+                self.resolved_joining_table._meta.foreign_key_columns[0:2]
+            )
+        return self._foreign_key_columns
```
yezz123, line 132:
```suggestion
            self._foreign_key_columns = (
                self.resolved_joining_table._meta.foreign_key_columns[:2]
            )
        return self._foreign_key_columns
```

The default starting value for a slice is 0, so it is unnecessary to explicitly define it.

piccolo/columns/m2m.py
```diff
@@ -0,0 +1,336 @@
...
+        :param foreign_key_columns:
+            If for some reason your joining table has more than two foreign key
+            columns, you can explicitly specify which two are relevant.
+
+        """
+        if foreign_key_columns:
+            if len(foreign_key_columns) != 2 or not all(
+                isinstance(i, ForeignKey) for i in foreign_key_columns
+            ):
+                raise ValueError("You must specify two ForeignKey columns.")
```
yezz123, line 323:
```suggestion
        if foreign_key_columns and (
            len(foreign_key_columns) != 2
            or not all(isinstance(i, ForeignKey) for i in foreign_key_columns)
        ):
            raise ValueError("You must specify two ForeignKey columns.")
```

Too much nesting can make code difficult to understand, and this is especially true in Python, where there are no brackets to help out with the delineation of different nesting levels.

Reading deeply nested code is confusing since you have to keep track of which conditions relate to which levels.

tests/columns/test_m2m.py
```diff
@@ -0,0 +1,328 @@
...
+
+    def test_get_m2m(self):
+        """
+        Make sure we can get related items via the joining table.
+        """
+        band: Band = Band.objects().get(Band.name == "Pythonistas").run_sync()
+
+        genres = band.get_m2m(Band.genres).run_sync()
+
+        self.assertTrue(all([isinstance(i, Table) for i in genres]))
```
yezz123, line 158:
```suggestion
        self.assertTrue(all(isinstance(i, Table) for i in genres))
```

Functions like `any`, `all` and `sum` allow you to pass in a generator rather than a collection. Doing so removes a pair of brackets, making the intent slightly clearer. It will also return immediately if a hat is found, rather than having to build the whole list.

tests/columns/test_m2m.py
```diff
@@ -0,0 +1,328 @@
...
+        """
+        Make sure we can get related items via the joining table.
+        """
+        customer: Customer = (
+            Customer.objects().get(Customer.name == "Bob").run_sync()
+        )
+
+        concerts = customer.get_m2m(Customer.concerts).run_sync()
+
+        self.assertTrue(all([isinstance(i, Table) for i in concerts]))
```
yezz123, line 324:
```suggestion
        self.assertTrue(all(isinstance(i, Table) for i in concerts))
```
Same here

yezz123 (commented):
Overall adding the M2M Prototype could help and fit the case that piccolo would like to describe and give to the developer, I like this addon 🚀

dantownsend:
@yezz123 Cool, thanks - I'll implement those suggestions.

dantownsend:
I think it's in a good state now. There is room for improvement in the future, but I think this puts a decent foundation in place.

If anyone has any suggestions let me know, otherwise will release it tomorrow. Thanks!

yezz123 (approved):

sinisaos (approved):

dantownsend:
I made one final tweak ... you can now request nested objects:

```python
>>> await Band.select(Band.name, Band.genres(Genre.id, Genre.name))
    [
        {
            'name': 'Pythonistas',
            'genres': [
                {'id': 1, 'name': 'Rock'},
                {'id': 2, 'name': 'Folk'}
            ]
        },
        ...
    ]
```

If you just want a single column in a list, you can still do that like this:

```python
>>> await Band.select(Band.name, Band.genres(Genre.name, as_list=True))
    [
        {
            'name': 'Pythonistas',
            'genres': ['Rock', 'Folk']
        },
        ...
    ]
```

yezz123:
Looks Clean @dantownsend 🚀

sinisaos:
@dantownsend These nested objects look and work great.

dantownsend:
Thanks again for all the input @sinisaos and @yezz123 👍

PR #382, opened 2021-12-24
max password length

dantownsend:
Added a max password length check to `BaseUser`.

### DIFF
--- a/docs/src/piccolo/authentication/baseuser.rst
+++ b/docs/src/piccolo/authentication/baseuser.rst
@@ -111,3 +111,11 @@ To change a user's password:
 
 .. warning:: Don't use bulk updates for passwords - use ``update_password`` /
    ``update_password_sync``, and they'll correctly hash the password.
+
+-------------------------------------------------------------------------------
+
+Limits
+------
+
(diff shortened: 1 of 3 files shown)


sinisaos:
@dantownsend Could you also add `min_length` password checker?

yezz123 (approved):
Good Ad-dons, its better to documented the use case of it 🚀

dantownsend:
@sinisaos Good idea - I'll add it.

@yezz123 Thanks, I'll add docs for it. The reason I added it is if someone uses a huge password, it takes longer to hash.

dantownsend:
@sinisaos I'll add min password length in a future PR.

sinisaos:
@dantownsend I later realized that this may not be so necessary because we cannot pass a password shorter than 4 characters when creating a user. https://github.com/piccolo-orm/piccolo/blob/2d207d8f8a1ea501016337aea692e59461133301/piccolo/apps/user/commands/create.py#L79

You can do a `min_length` password checker in the `login` method or you don't have to, as you wish.

dantownsend:
@sinisaos Cool, thanks.

PR #389, opened 2022-01-10
Nested model fix

dantownsend:
Fixes https://github.com/piccolo-orm/piccolo_api/issues/129

`create_pydantic_model` has the ability to create nested models. For example:

```python
create_pydantic_model(Band, nested=(Band.manager,))
```

In the above example, `Band` has a sub model called `Manager`.

It was discovered by @wmshort that there's a bug where Pydantic sometimes can't uniquely identify these sub models. The solution is to set the `__qualname__` of the sub model. In the above example, the `Manager` sub model now has a  `__qualname__` of `Band.manager`.

To learn more about `__qualname__`, see [PEP 3155](https://www.python.org/dev/peps/pep-3155/).

## Design decisions

One decision I wasn't sure about was what to use for the `__name__` attribute of these submodels.

If `__qualname__` is `Band.manager`, then `__name__` should be `manager`. However, if we do this we can end up with loads of models called 'manager', and in the Swagger docs of FastAPI / BlackSheep etc it's not very user friendly.

Screenshot 2022-01-10 at 12 51 52

So at the moment, I've set `__name__` to be the same as `__qualname__`, so they are uniquely identifiable in the Swagger docs (i.e. it will show `Band.manager` instead of `Manager`.

### DIFF
--- a/piccolo/utils/pydantic.py
+++ b/piccolo/utils/pydantic.py
@@ -198,6 +198,8 @@ def create_pydantic_model(
             )
         )
 
+    model_name = model_name or table.__name__
+
     for column in piccolo_columns:
         column_name = column._meta.name
 
@@ -255,6 +257,7 @@ def create_pydantic_model(
                     )
                 )
             ):
+                nested_model_name = f"{model_name}.{column._meta.name}"
                 _type = create_pydantic_model(
                     table=column._foreign_key_meta.resolved_references,
                     nested=nested,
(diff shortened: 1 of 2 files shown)


wmshort (approved):
This solution nicely accommodates multiply-nested models and maintains readability in the Swagger / OpenAPI UI.

dantownsend:
@wmshort Thanks. I think this solution makes the most sense for now.

dantownsend:
@wmshort The fix is on PyPI now - version 0.62.3.

wmshort:
All good now! Brilliant!

dantownsend:
@wmshort Awesome

Issue #390, opened 2022-01-11
Problem when add BaseUser to existing table


sinisaos:
I ran to a problem when I want to add `BaseUser` to an existing table (e.g. in the default Task app) after using `piccolo migrations  forwards user`. If I want to add new colum like this
```python
from piccolo.apps.user.tables import BaseUser
from piccolo.table import Table
from piccolo.columns import Varchar, Boolean, ForeignKey


class Task(Table):
    """
    An example table.
    """

    name = Varchar()
    completed = Boolean(default=False)
    task_user = ForeignKey(BaseUser) <= new column
```
I got the error `relation "piccolo_user" already exists`. If I don't run `piccolo migrations forwards user`, everything is fine and I can add BaseUser to the existing table. If I create a new application with `piccolo asgi new` and run only `piccolo migrations forwards session_auth` without `piccolo migrations forwards user`, everything is fine and I can add `BaseUser` to the existing table, can create a user (there is a warning that 3 migrations have not started - the app may not behave as expected) and the `piccolo_user` table is in admin.
I don't remember how I did it before, but not running a user app migration solves the problem. If there is another way, it would be good to add it to the docs to make users aware of it. Thanks.

dantownsend:
@sinisaos I think the problem is when you look at the `piccolo_app.py` file generated by `piccolo asgi new`, it uses `table_finder`.

It will be importing all `Table` classes from `table.py`, including `BaseUser`. When you create a migration file to add the `task_user` column, it is probably also trying to recreate `BaseUser`.

The way I get around this is using tags.

```python
class Task(Table, tags=['tasks']):
    ...
```
Then `table_finder` becomes: `table_finder(['home.tables'], include_tags=['tasks'])`.

Another solution is to use `LazyTableReference` for the `ForeignKey`, rather than directly importing `BaseUser`:

```python
task_user = ForeignKey(LazyTableReference("BaseUser", module_path="piccolo.apps.user.tables"))
```
I could add a new option to `table_finder` which automatically excludes any `Table` classes not defined in that file. Something like this maybe:

```python
table_finder(['home.tables'], exclude_imported=True)
```

What do you think?

PR #391, opened 2022-01-11
added `exclude_imported` option to `table_finder`

dantownsend:
Fixes https://github.com/piccolo-orm/piccolo/issues/390

When using `table_finder` to register `Table` subclasses with `AppConfig` (for auto migrations), you would sometimes encounter situations like this, where you import ``Table`` subclasses from other files:

```python
from piccolo.apps.user.tables import BaseUser # <- This
from piccolo.columns.column_types import ForeignKey, Varchar


class Musician(Table):
    name = Varchar()
    user = ForeignKey(BaseUser)

```

If we just wanted to import ``Musician`` and not ``BaseUser``, then the existing solution was to use tags:

```python
class Musician(Table, tags=['music']):
    name = Varchar()
    user = ForeignKey(BaseUser)
```

And then in ``piccolo_app.py``:

```python
APP_CONFIG = AppConfig(
    ...
    table_classes=table_finder(['music.tables'], include_tags=['music']),
)
```

It works OK, but means adding tags to all of your ``Table`` subclasses. Tags are still supported, but the new solution is to use ``exclude_imported``:

```python
APP_CONFIG = AppConfig(
    ...
    table_classes=table_finder(['music.tables'], exclude_imported=True),
)
```

It's set to ``False`` by default, to maintain backwards compatibility. But it's set to ``True`` in the template generated by ``piccolo asgi new`` so new users will have it switched on.

### DIFF
--- a/piccolo/apps/asgi/commands/templates/app/home/piccolo_app.py.jinja
+++ b/piccolo/apps/asgi/commands/templates/app/home/piccolo_app.py.jinja
@@ -16,7 +16,7 @@ APP_CONFIG = AppConfig(
     migrations_folder_path=os.path.join(
         CURRENT_DIRECTORY, "piccolo_migrations"
     ),
-    table_classes=table_finder(modules=["home.tables"]),
+    table_classes=table_finder(modules=["home.tables"], exclude_imported=True),
     migration_dependencies=[],
     commands=[],
 )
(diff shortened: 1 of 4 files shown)


dantownsend:
@sinisaos This is a potential solution to you issue.

The integration test is failing because it installs the version of Piccolo currently on PyPI, which doesn't support this new feature, so it's nothing to worry about.

As always, naming is probably the biggest challenge. `exclude_imported` is descriptive, but not super memorable. `allow_imports` or something instead?

sinisaos:
@dantownsend I like name `exclude_imported` (maybe can be `exclude_imported_tables`) . Can you add your description from the previous comment to the docs. It’s really nice and explains how to use this and more.

dantownsend:
@sinisaos Yeah, I need to rework the `table_finder` docs - it probably deserves it's own page. I'll do that in a separate PR.

PR #392, opened 2022-01-12
make the `value_type` dynamic for `ForeignKey`

dantownsend:
Each `Column` has a `value_type` which is basically the Python type it maps to.

It used to always be `int` for `ForeignKey`, but this is wrong, as it depends on the primary key being referred to (i.e. if it's referencing a table where a `UUID` is the primary key, then the column must also be of type `uuid`).


sinisaos (approved):
You are a genius. At least for me this  dynamic `value_type` for `ForeignKey` solves the problem with FK relationships with custom PK.

Issue #394, opened 2022-01-13
Foreign Keys that reference non pk columns


dantownsend:
### Discussed in https://github.com/piccolo-orm/piccolo/discussions/393

<div type='discussions-op-text'>

<sup>Originally posted by **theelderbeever** January 13, 2022</sup>
Is it possible to create a foreign key on a table that references a column other than the primary key on a table? The documentation seems to imply that you can only pass a table to `ForeignKey` and it picks up the pk. I would like to do something like the following...

```python
class Contract(BaseTable):

    address = Varchar(index=True, required=True, null=False, unique=True)

class Log(BaseTable):

    contract_address = ForeignKey(Contract.address)
```

Is there a way to accomplish this?</div>

dantownsend:
Reopening, as there are still changes required to Piccolo Admin and Piccolo API.

sinisaos:
@dantownsend Fix for Piccolo Api and Piccolo Admin is [here](https://github.com/piccolo-orm/piccolo_api/pull/134)

PR #395, opened 2022-01-13
allow setting target column for ForeignKey

dantownsend:
Fixes https://github.com/piccolo-orm/piccolo/issues/394

When you create a `ForeignKey` with Piccolo it automatically points to the primary key on the related table. With this PR you can specify a different column on the related table if you want (as long as it's unique).

Currently a proof of concept - seems to work fine, but needs some more tests.


theelderbeever:
I pulled the branch down to play around with it and do some testing on my own. I maybe making a mistake but when I setup my tables and then autogenerate a migration and apply it the foreign keys ignore the `target_column` field and end up using the pk from the table instead.


```python
# How it is in my table def
contract_address = ForeignKey(references=Contract, target_column=Contract.address)

# How it shows up in the migration file
    manager.add_column(
        table_class_name="Log",
        tablename="log",
        column_name="contract_address",
        db_column_name="contract_address",
        column_class_name="ForeignKey",
        column_class=ForeignKey,
        params={
            "references": Contract,
            "on_delete": OnDelete.cascade,
            "on_update": OnUpdate.cascade,
            "null": True,
            "primary_key": False,
            "unique": False,
            "index": False,
            "index_method": IndexMethod.btree,
            "choices": None,
            "db_column_name": None,
            "secret": False,
        },
    )
```

dantownsend:
@theelderbeever Thanks for taking a look at it. You're right - there's a bug. The `target_column` value should be pulling through to the migration. I'll push the fix now.

I need to write some more tests for migrations - as you say, the `address` column should be created before the foreign key.

dantownsend:
It seems to work now.

Piccolo Admin will require some changes to support this new feature.

theelderbeever:
Awesome! Would the piccolo_admin issue show up as the tables not displaying values properly by chance?

dantownsend:
@theelderbeever Yes, I think the list view and foreign key selector will be a bit broken.

I'll release this, then will try and sort out the admin.

PR #396, opened 2022-01-14
Add from_fixture to BaseUser

### LINKED ISSUE #265: Add a `from_fixture` method to `BaseUser`
If you pass a password into the `BaseUser` constructor, it hashes it.

```python
user = BaseUser(username="bob", password="bob123")
>>> user.password
pbkdf2_sha256$10000$bb353d7fc92...
```

This is the expected behaviour, but it's problematic if you're passing in an already hashed password (it will get hashed again).

An example use case is fixtures - you dump the users table, and end up with data like this:

```json
{
    "user": {
        "BaseUser": [
            {
                "id": 1,
                "username": "bob",
                "password": "pbkdf2_sha256$10000$abc123...",
                "first_name": "",
                "last_name": "",
                "email": "bob@bob.com",
                "active": true,
                "admin": true,
                "superuser": true,
                "last_login": null
            },
        ]
    },
```

We need a class method on `BaseUser`, called something like `from_fixture` (or some other, better name).

```python
class BaseUser(Table):
    @classmethod
    def from_fixture(self, data: t.Dict[str, t.Any]) -> BaseUser:
       # Create a BaseUser instance, but bypasses the password hashing, as we can assume the password is already hashed.
        ...
```

We should also raise a `ValueError` in the `Baseuser` constructor if someone tries to pass in an already hashed password.

mrbazzan:
Closes #265

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,7 +7,7 @@ build/
 .doctrees/
 .vscode/
 piccolo.egg-info/
-build/
+.idea/
 dist/
 piccolo.sqlite
 _build/
--- a/piccolo/apps/user/tables.py
+++ b/piccolo/apps/user/tables.py
@@ -50,8 +50,20 @@ class BaseUser(Table, tablename="piccolo_user"):
         # Generating passwords upfront is expensive, so might need reworking.
         password = kwargs.get("password", None)
         if password:
+            if password.startswith("pbkdf2_sha256"):
+                logger.warning("Hashed password passed to the constructor.")
+                raise ValueError("Do not pass hashed password.")
+
             kwargs["password"] = self.__class__.hash_password(password)
         super().__init__(**kwargs)
+
+    @classmethod
+    def from_dict(cls, data: t.Dict[str, t.Any]) -> BaseUser:
+        """Create a BaseUser instance from fixture"""
+        password = data.get("password")
+        user = cls(**{k: v for k, v in data.items() if k != "password"})
+        user.__setattr__("password", password)
+        return user
 
     @classmethod
     def get_salt(cls):
--- a/piccolo/apps/fixtures/commands/load.py
+++ b/piccolo/apps/fixtures/commands/load.py
@@ -56,7 +56,7 @@ async def load_json_string(json_string: str):
 
                 await table_class.insert(
                     *[
-                        table_class(**row.__dict__)
+                        table_class.from_dict(row.__dict__)
                         for row in model_instance_list
                     ]
                 ).run()
--- a/piccolo/table.py
+++ b/piccolo/table.py
@@ -310,6 +310,10 @@ class Table(metaclass=TableMetaclass):
 
         return pk
 
+    @classmethod
+    def from_dict(cls, data: t.Dict[str, t.Any]) -> Table:
+        return cls(**data)
+
     ###########################################################################
 
     def save(
--- a/tests/apps/user/test_tables.py
+++ b/tests/apps/user/test_tables.py
@@ -42,6 +42,18 @@ class TestInstantiateUser(TestCase):
             manager.exception.__str__(), "The password is too long."
         )
 
+    @patch("piccolo.apps.user.tables.logger")
+    def test_hashed_password(self, logger: MagicMock):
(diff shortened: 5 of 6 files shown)


piccolo/apps/user/tables.py
```diff
@@ -50,9 +50,24 @@ def __init__(self, **kwargs):
         # Generating passwords upfront is expensive, so might need reworking.
         password = kwargs.get("password", None)
         if password:
+            if password.startswith("pbkdf2_sha256") and len(password) == 117:
```
sinisaos, line 53: I think that only first check `if password.startswith("pbkdf2_sha256")` will be enough.

piccolo/apps/user/tables.py
```diff
@@ -50,9 +50,24 @@ def __init__(self, **kwargs):
...
 
+    @classmethod
+    def from_fixture(cls, data: t.Dict[str, t.Any]):
+        """Create a BaseUser instance from fixture"""
+        for kwargs in data["user"][cls.__name__]:
+            password = kwargs.get("password")
+            user = cls(**{k: v for k, v in kwargs.items() if k != "password"})
+            user.password = password
+
+            return user
```
sinisaos, line 69: With the new format given by Daniel this method can be simpler. Something like this
```python
@classmethod
def from_fixture(cls, data: t.Dict[str, t.Any]) -> BaseUser:
    """Create a BaseUser instance from fixture"""
    password = data.get("password")
    user = cls(**{k: v for k, v in data.items() if k != "password"})
    user.password = password
    return user
```

sinisaos:
@mrbazzan I think this is great, but we have to wait to see what Daniel will say about this. Can you add tests for new code to maintain coverage. Thanks.

piccolo/apps/user/tables.py
mrbazzan: What if the user intentionally enters a password that starts with "pbkdf2_sha256"? The `len(password)` is to differentiate between if a value was hashed(`hash_password` returns a password of length 117) or just a value entered by the user that starts with "pbkdf2_sha256"?

Should I still remove the condition?

dantownsend: I think both have valid arguments. On the one hand checking the length provides extra confidence. But we can't guarantee that the length will always be 117 (we might increase the default salt size and / or number of iterations in the future). So I think it's OK to remove it.

.gitignore
```diff
@@ -7,7 +7,7 @@ build/
 .doctrees/
 .vscode/
 piccolo.egg-info/
-build/
```
dantownsend, line 10: Looks like this was removed by mistake.

mrbazzan: I intentionally removed it. There is another one on line 6

dantownsend:
I had an idea - we could call this method `from_dict` instead. As we already have a `to_dict` method, it kind of makes sense. Even though it will be used primarily for fixtures, it is more generally useful.

We should implement a generic `from_dict` class method on `Table` too.

```python
class Table(metaclass=TableMetaclass):
    @classmethod
    def from_dict(cls, dict: t.Dict[str, t.Any]) -> Table:
        return cls(**dict)
```

We just happen to be overriding this default implementation for ``BaseUser``.

piccolo/apps/user/tables.py
mrbazzan: Okay.

dantownsend:
@mrbazzan This looks great, thanks.

We just have to modify this, so it uses `from_dict`:

https://github.com/piccolo-orm/piccolo/blob/6b23206ac1b870a9f4bf4fd290cdec28e425867f/piccolo/apps/fixtures/commands/load.py#L57-L62

mrbazzan:
> @mrbazzan This looks great, thanks.
> 
> We just have to modify this, so it uses `from_dict`:
> 
> https://github.com/piccolo-orm/piccolo/blob/6b23206ac1b870a9f4bf4fd290cdec28e425867f/piccolo/apps/fixtures/commands/load.py#L57-L62

@dantownsend 
Done! Please review.

dantownsend:
@mrbazzan Looks great - thanks again for this 👍

Issue #404, opened 2022-01-25
str representation of query not compatible with Table.raw when using uuid/date columns


theelderbeever:
### Discussed in https://github.com/piccolo-orm/piccolo/discussions/403

<div type='discussions-op-text'>

<sup>Originally posted by **theelderbeever** January 25, 2022</sup>
Version info:
piccolo = "0.64.0"

I am attempting to use some of the query building tools to generate custom complex queries that aren't directly possible through the ORM at the moment. During this I ran into a case where Attempting to run the `str(MyTable.insert(MyTable()))` doesn't seem to work with uuid or date columns as they aren't wrapped in `''`.

I believe the following minimal example should work

```python
from datetime import datetime

from piccolo.columns import UUID, Timestamptz
from piccolo.table import Table


class Table(Table):

    uuid = UUID(primary_key=True)
    created_at = Timestamptz(default=datetime.now, index=True)
    updated_at = Timestamptz(default=datetime.now, index=True)


# Instantiate an instance
table = Table()
Table.raw(str(Table.insert(table))).run_sync()
```
It is contrived but the query fails with a SQL error midway through the UUID string that gets output. The uuid value _should_ be wrapped in `''` I believe.

Example Error
```python
PostgresSyntaxError: syntax error at or near "fa"
```

This seems to come down to how the `Table.querystring` is formed. The values don't get wrapped and the datetime doesn't replace the `' '` with a `T`. [Relevant Code](https://github.com/piccolo-orm/piccolo/blob/master/piccolo/querystring.py#L89)
```python
>>> print(table.querystring)
(3a47b208-4834-4d0d-9213-db3511a52886,2022-01-25 16:58:44.238463+00:00,2022-01-25 16:58:44.238470+00:00)
```

Ultimate goal here is that I am writing some custom on conflict statements and was taking inspiration from the `Insert.postgres_querystrings` method to construct them when I ran into this.
</div>

PR #405, opened 2022-01-25
Add UUID type conversion and fix datetime comparison

theelderbeever:
Reference Issue #404 

## Motivation
The querystring builder is valuable when trying to write custom complex queries with `VALUES` however, certain values are not converted to sql compatible values properly. This results in `Table.raw(query).run()` failing. Changes here aspire to address this issue.

## Explanation
A couple of type check fixes. The `datetime` comparison in the `querystring` was comparing to the module not the class and wouldn't be picked up. Additionally, the `.replace("T", " ")` doesn't work with postgres as far as I can tell and I removed it.

There also needs to be a check on `UUID`. When being written to a query string the value should be wrapped in `''`.

I imagine something similar may need to be done with JSONB, Enum, etc. Maybe there could be some merging of usage with the `piccolo.utils.sql_values.convert_to_sql_value` function.

### DIFF
--- a/piccolo/querystring.py
+++ b/piccolo/querystring.py
@@ -1,10 +1,10 @@
 from __future__ import annotations
 
-from datetime import datetime
 import typing as t
 from dataclasses import dataclass
-from string import Formatter
+from datetime import datetime
 from importlib.util import find_spec
+from string import Formatter
 
 if t.TYPE_CHECKING:
     from piccolo.table import Table
@@ -13,6 +13,7 @@ if find_spec("asyncpg"):
     from asyncpg.pgproto.pgproto import UUID
 else:
     from uuid import UUID
+
 
 @dataclass
 class Unquoted:
--- a/piccolo/querystring.py
+++ b/piccolo/querystring.py
@@ -9,10 +9,11 @@ from string import Formatter
 if t.TYPE_CHECKING:
     from piccolo.table import Table
 
+from uuid import UUID
 if find_spec("asyncpg"):
-    from asyncpg.pgproto.pgproto import UUID
+    from asyncpg.pgproto.pgproto import UUID as apgUUID
 else:
-    from uuid import UUID
+    apgUUID = UUID
 
 
 @dataclass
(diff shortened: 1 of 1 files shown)


piccolo/querystring.py
```diff
@@ -91,8 +97,10 @@ def __str__(self):
             if _type == str:
                 converted_args.append(f"'{arg}'")
             elif _type == datetime:
-                dt_string = arg.isoformat().replace("T", " ")
```
dantownsend, line 94: I checked with and without the `'T'`, and both seem to work. So you're right, the `replace` is redundant.

piccolo/querystring.py
```diff
@@ -1,13 +1,19 @@
...
+from importlib.util import find_spec
 from string import Formatter
 
 if t.TYPE_CHECKING:
     from piccolo.table import Table
 
+if find_spec("asyncpg"):
+    from asyncpg.pgproto.pgproto import UUID
+else:
+    from uuid import UUID
```
dantownsend, line 15: Ideally it would check both of these, because for Postgres, a value could be a `uuid.UUID` or `asyncpg.pgproto.pgproto.UUID`.

We pass `uuid.UUID` into queries like INSERTS, but asyncpg returns `asyncpg.pgproto.pgproto.UUID` from SELECT. We're most likely to be dealing with `uuid.UUID`.

theelderbeever: `asyncpg` only gets installed if the `piccolo[postgres]` is installed, correct? I am not sure I have a good idea on how the case where it hasn't been installed other than moving the `find_spec("asyncpg")` into the conversion `if-elif-else`  scope. Is that what you would recommend?

dantownsend:
@theelderbeever Thanks for this. The changes are definitely useful.

The string representation of a query is more of a convenience, so you can do something like `print(Band.select())` and immediately see the SQL that would be generated. That's why there might be bugs with it, because Piccolo doesn't use it directly in queries.

I think this issue has also made it clear that an API is needed for composing custom queries with `Querystring`, like `await Manager.run_querystring(querystring)`. I'll create a separate issue for that.

piccolo/querystring.py
theelderbeever: or something like

```python
if find_spec("asyncpg"):
    from asyncpg.pgproto.pgproto import UUID as apgUUID
else:
    from uuid import UUID
    apgUUID = UUID
 ```
Then do the appropriate typecheck on line 102

dantownsend: Yes, that works.

dantownsend:
@theelderbeever Thanks, looks good 👍

PR #407, opened 2022-01-27
make docs async first

dantownsend:
Most of the examples in the docs use the sync version:

```python
Band.select().run_sync()
```

The reason for this was so people could copy and paste any query from the docs into the playground, and it would work.

However, the playground has supported top level await for a while. You can paste this in and it will work:

```python
await Band.select()
```

Also, Piccolo was designed to be async first, and is generally used within async web frameworks.

Long story short ... this PR changes the docs so examples use `async` by default instead of `run_sync`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,23 +33,23 @@ await Band.select(
     Band.name
 ).where(
     Band.popularity > 100
-).run()
+)
 
 # Join:
 await Band.select(
     Band.name,
     Band.manager.name
-).run()
+)
 
 # Delete:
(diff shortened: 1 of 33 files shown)


sinisaos (approved):
You’ve well noticed that the playground has supported top level await for a while. This looks great.

dantownsend:
@sinisaos Thanks. I think it makes more sense this way.

sinisaos:
@dantownsend I absolutely agree with you.

Issue #411, opened 2022-01-29
Add error to Table.update() without where clause


dantownsend:
### Discussed in https://github.com/piccolo-orm/piccolo/discussions/409

<div type='discussions-op-text'>

<sup>Originally posted by **theelderbeever** January 28, 2022</sup>
As a point of discussion running `await Table.update({"column": "value"})` without a `.where` will cause an update on an entire table. While this is a poor developer mistake to make I wonder if it would be wise to have an arg in the the update to dictate behavior of the ORM when performing such a query. Maybe something like 

```python
>>> await Table.update({"column": "value"}, without_where="error")

UpdateWithoutWhereError("An update on the entire table was attempted please explicitly allow this behavior")
```
Have multiple options such as `error`, `warn`, `ignore`.  And just have it defaulted. Having made this mistake in production it can make for a very bad evening.

Maybe this could exist on `delete` as well?</div>

PR #412, opened 2022-01-29
add `force` argument to `Update` queries

dantownsend:
Fixes https://github.com/piccolo-orm/piccolo/issues/411

As observed in this discussion https://github.com/piccolo-orm/piccolo/discussions/409, an accidental mass update can be almost as disruptive as a mass deletion.

```python
# This would be bad if done by accident, as every band would have its name updated to 'Mega Band':
await Band.update({Band.name: 'Mega Band'})
```

This PR makes it so an update query requires a where clause, or specifying `force=True`.

```python
>>> await Band.update({Band.name: 'Mega Band'})
UpdateError!

# Works OK, because it has a where clause:
>>> await Band.update({Band.name: 'Mega Band'}).where(Band.name == 'Good Band')

 # You can still update every row, but you must specify `force=True`
>>> await Band.update({Band.name: 'Mega Band'}, force=True)
```

This is a somewhat breaking change, which we try to avoid. However, the impact is likely to be small, as I imagine most users only use update with a where clause. On balance I think it's definitely worthwhile.

### DIFF
--- a/docs/src/piccolo/query_types/update.rst
+++ b/docs/src/piccolo/query_types/update.rst
@@ -14,13 +14,30 @@ This is used to update any rows in the table which match the criteria.
     >>> )
     []
 
-As well as replacing values with new ones, you can also modify existing values, for
-instance by adding to an integer.
+-------------------------------------------------------------------------------
+
+force
+-----
+
+Piccolo won't let you run an update query without a where clause, unless you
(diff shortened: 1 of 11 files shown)


sinisaos (approved):
I think this is great and adds additional security if someone forgot to put `where()` clause to update.

dantownsend:
@sinisaos Thanks for reviewing

Issue #410, opened 2022-01-29
Where JSONB column is_null produces incorrect query statement


theelderbeever:
python 3.10.1
piccolo: 0.66.0
db: Postgres

When doing a select or otherwise and comparing a `JSONB` columns to `None` or `.is_null()` the query string adds in a python object causing an error in the Postgres prepare statement.

```python
Table.select(Table.id).where(Table.jsonb_column == None).run_sync()
Table.select(Table.id).where(Table.jsonb_column.is_null()).run_sync()
```

Emitted sql
```sql
SELECT "table"."id" FROM table WHERE "table"."jsonb_column" IS NULL'"<piccolo.columns.combination.Undefined object at 0x103eded10>"'
```

Error
```python
asyncpg.exceptions.PostgresSyntaxError: syntax error at or near "NULL$1"
```

dantownsend:
You're right - I've been able to replicate this.

It's pretty strange. I'll look into it.

theelderbeever:
Not sure if relevant or already addressed by this but when doing an `Table.insert(Table(jsonb_column=None))` Piccolo doesn't appear to encode the value as a SQL `NULL` value which makes `NOT NULL` and `IS NULL` comparison not behave as expected. The value will show up as `null` in the database but it will be a jsonb serialized `null` instead of a SQL `NULL`.

theelderbeever:
Maybe a `SQL_NULL` delegate so the ORM knows to add it to the query as `NULL` without quoting it?

theelderbeever:
Tested with setting the column default to `None` and that value is appropriately set to a SQL `NULL` so this only occurs when doing an `INSERT` with a `None`.

dantownsend:
@theelderbeever Yes, you're right - I only discovered that yesterday. I added a fix for it as part of the same PR:

https://github.com/piccolo-orm/piccolo/pull/413

What I was thinking was, if the column is nullable:

 * Specify `None` for SQL null.
 * Specify. `'null'` for JSON null

If the column isn't nullable then both `None` and `'null'` give JSON null.

What do you think?

theelderbeever:
With the risk of having too many configurable arguments... I think `None` *always* being SQL `NULL` is a bit more consistent. That way devs can rely on the  `NOT NULL` constraint to reject values.

dantownsend:
@theelderbeever I think you're right - making `None` always mean SQL null is more straight forward and easier to understand.

theelderbeever:
[...]

PR #413, opened 2022-01-29
fix `JSONB` column where null

dantownsend:
Related to https://github.com/piccolo-orm/piccolo/issues/410

For example:

```python
class RecordingStudio(Table):
    facilities = JSONB(null=True)

# This was causing an error:
await RecordingStudio.select().where(
    RecordingStudio.facilities.is_null()
)
```

The problem was due to a bug with how Piccolo serialises JSON values.

### DIFF
--- a/requirements/dev-requirements.txt
+++ b/requirements/dev-requirements.txt
@@ -1,9 +1,9 @@
-black>=21.7b0
+black==22.1.0
 ipdb==0.13.9
-ipython==7.31.1
+ipython==8.0.1
 flake8==4.0.1
 isort==5.10.1
 twine==3.7.1
-mypy==0.930
+mypy==0.931
 pip-upgrader==1.4.15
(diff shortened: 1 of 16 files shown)


requirements/dev-requirements.txt
```diff
@@ -1,9 +1,9 @@
-black>=21.7b0
+black==22.1.0
 ipdb==0.13.9
-ipython==7.31.1
+ipython==8.0.1
```
sinisaos, line 3: I think you should put dependencies like this `ipython>=7.31.1` to pass the Python3.7 linter because `ipython==8.0.1` cannot be installed in Python3.7

dantownsend: Good point. I'll change that.

dantownsend: Done

Issue #414, opened 2022-02-02
Asgi template for Xpresso framework


sinisaos:
We could add an asgi template for the new [Xpresso](https://xpresso-api.dev/0.9.2/) framework. We practically need to add two files and change several others because Xpresso shares a lot of things with Starlette and we can [mount](https://xpresso-api.dev/0.9.2/tutorial/routing/) Piccolo Admin at a route. I can do PR because I’ve already tried that and everything works well. What do you think about that?

dantownsend:
I think it's a good idea. If we were to support another ASGI framework, then Xpresso would be top of my list, because it's shaping up to be a nice framework, and as you mention we can easily mount Piccolo Admin.

It's still quite early days with Xpresso though, and the API could potentially change. @adriangb - do you think the API is stable enough for this?

@sinisaos If you already have it working locally, then definitely open a PR for it.

sinisaos:
@dantownsend Done.

adriangb:
> We can quite easily extend it to test Xpresso too, so we'll know if something breaks:

I think that'd be a good idea. It will help both of us. I'm happy to fix any issues that come up as long as we can detect them.

PR #415, opened 2022-02-02
asgi template for Xpresso framework

sinisaos:
Related to #414

### DIFF
--- a/docs/src/piccolo/asgi/index.rst
+++ b/docs/src/piccolo/asgi/index.rst
@@ -21,7 +21,7 @@ Routing frameworks
 ******************
 
 Currently, `Starlette <https://www.starlette.io/>`_, `FastAPI <https://fastapi.tiangolo.com/>`_,
-`BlackSheep <https://www.neoteroi.dev/blacksheep/>`_ and `Xpresso <https://xpresso-api.dev/0.9.2/>`_
+`BlackSheep <https://www.neoteroi.dev/blacksheep/>`_ and `Xpresso <https://xpresso-api.dev/>`_
 are supported.
 
 Other great ASGI routing frameworks exist, and may be supported in the future
--- a/docs/src/piccolo/asgi/index.rst
+++ b/docs/src/piccolo/asgi/index.rst
@@ -21,7 +21,8 @@ Routing frameworks
 ******************
 
 Currently, `Starlette <https://www.starlette.io/>`_, `FastAPI <https://fastapi.tiangolo.com/>`_,
-and `BlackSheep <https://www.neoteroi.dev/blacksheep/>`_ are supported.
+`BlackSheep <https://www.neoteroi.dev/blacksheep/>`_ and `Xpresso <https://xpresso-api.dev/0.9.2/>`_
+are supported.
 
 Other great ASGI routing frameworks exist, and may be supported in the future
 (`Quart <https://pgjones.gitlab.io/quart/>`_ ,
--- a/piccolo/apps/asgi/commands/templates/app/_xpresso_app.py.jinja
+++ b/piccolo/apps/asgi/commands/templates/app/_xpresso_app.py.jinja
@@ -23,7 +23,7 @@ async def tasks() -> t.List[TaskModelOut]:
 
 
 async def create_task(task_model: FromJson[TaskModelIn]) -> TaskModelOut:
-    task = Task(**task_model.__dict__)
+    task = Task(**task_model.dict())
     await task.save()
     return task.to_dict()
 
@@ -35,7 +35,7 @@ async def update_task(
     if not task:
         raise HTTPException(status_code=404)
 
-    for key, value in task_model.__dict__.items():
+    for key, value in task_model.dict().items():
         setattr(task, key, value)
 
     await task.save()
@@ -56,8 +56,10 @@ async def delete_task(task_id: FromPath[int]):
 @asynccontextmanager
 async def lifespan():
     await open_database_connection_pool()
-    yield
-    await close_database_connection_pool()
+    try:
+        yield
+    finally:
+        await close_database_connection_pool()
 
 
 app = App(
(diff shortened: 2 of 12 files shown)


piccolo/apps/asgi/commands/templates/app/_xpresso_app.py.jinja
```diff
@@ -0,0 +1,104 @@
...
+    except Exception:
+        print("Unable to connect to the database")
+
+
+async def close_database_connection_pool():
+    try:
+        engine = engine_finder()
+        await engine.close_connection_pool()
+    except Exception:
+        print("Unable to connect to the database")
```
dantownsend, line 104: These need tying into the Xpresso app's lifecycle.

I think it should be something like:

```python
from contextlib import asynccontextmanager

@asynccontextmanager
async def lifecycle():
    await open_database_connection_pool()
    yield
    await close_database_connection_pool()

# And then pass this into `App`.
app = App(lifecycle=lifecycle, ...)
```

piccolo/apps/asgi/commands/templates/app/_xpresso_app.py.jinja
```diff
@@ -0,0 +1,104 @@
...
+TaskModelIn: t.Any = create_pydantic_model(
+    table=Task, model_name="TaskModelIn"
+)
+TaskModelOut: t.Any = create_pydantic_model(
+    table=Task, include_default_columns=True, model_name="TaskModelOut"
+)
+
+
+async def tasks() -> t.List[TaskModelOut]:
+    return await Task.select().order_by(Task.id).run()
```
dantownsend, line 23: Can we change these to drop the `run`? So it's just `await Task.select().order_by(Task.id)`.

I know the other ASGI templates use `run`, but I'd like to update them so they're consistent with the docs.

dantownsend:
@sinisaos Thanks for this - just some minor suggestions.

piccolo/apps/asgi/commands/templates/app/_xpresso_app.py.jinja
sinisaos: @dantownsend You are right. It’s just a `lifespan`, not a `lifecycle`. I'll change that.

dantownsend: Oops, good catch!

docs/src/piccolo/asgi/index.rst
```diff
@@ -21,7 +21,8 @@ Routing frameworks
 ******************
 
 Currently, `Starlette <https://www.starlette.io/>`_, `FastAPI <https://fastapi.tiangolo.com/>`_,
-and `BlackSheep <https://www.neoteroi.dev/blacksheep/>`_ are supported.
+`BlackSheep <https://www.neoteroi.dev/blacksheep/>`_ and `Xpresso <https://xpresso-api.dev/0.9.2/>`_
```
dantownsend, line 24: Sorry, can we just change this link to be `https://xpresso-api.dev/`.

sinisaos: Of course. I just pasted the link and didn't pay attention to the url, but I can change that only later in the evening.

dantownsend: No worries. Other than that, I think it's good to go - thanks!

piccolo/apps/asgi/commands/templates/app/home/_xpresso_endpoints.py.jinja
```diff
@@ -0,0 +1,21 @@
+import os
+
+import jinja2
+from starlette.endpoints import HTTPEndpoint
+from starlette.responses import HTMLResponse
```
adriangb, line 5:
```suggestion
from xpresso.responses import HTMLResponse
```

I did the same re-export as most other Starlette based frameworks do.
I figure more than for convenience I'd do it so that if at some point I need to subclass `Request` or something I can maybe do it without breaking people's code.

piccolo/apps/asgi/commands/templates/app/_xpresso_app.py.jinja
```diff
@@ -0,0 +1,111 @@
...
+    await task.remove()
+
+    return {}
+
+
+@asynccontextmanager
+async def lifespan():
+    await open_database_connection_pool()
+    yield
+    await close_database_connection_pool()
```
adriangb, line 60:
```suggestion
    await open_database_connection_pool()
    try:
        yield
    finally:
        await close_database_connection_pool()
```

piccolo/apps/asgi/commands/templates/app/_xpresso_app.py.jinja
```diff
@@ -0,0 +1,111 @@
...
+    table=Task, include_default_columns=True, model_name="TaskModelOut"
+)
+
+
+async def tasks() -> t.List[TaskModelOut]:
+    return await Task.select().order_by(Task.id)
+
+
+async def create_task(task_model: FromJson[TaskModelIn]) -> TaskModelOut:
+    task = Task(**task_model.__dict__)
```
adriangb, line 26: Just for my knowledge, is there any particular reason to do `task_model.__dict__` instead of `task_model.dict()`? Same with `for key, value in task_model.__dict__.items()` [below](https://github.com/piccolo-orm/piccolo/blob/6566deb7a7c5d8868893291a2547084091c65b6a/piccolo/apps/asgi/commands/templates/app/_xpresso_app.py.jinja#L38).

piccolo/apps/asgi/commands/templates/app/home/_xpresso_endpoints.py.jinja
```diff
@@ -0,0 +1,21 @@
...
+
+
+async def home():
+    template = ENVIRONMENT.get_template("home.html.jinja")
+
+    content = template.render(
+        title="Piccolo + ASGI",
+    )
+
+    return HTMLResponse(content)
```
adriangb, line 21:
```suggestion
    return HTMLResponse(content)

```

piccolo/apps/asgi/commands/templates/app/_xpresso_app.py.jinja
dantownsend: You're right - it should be `dict()`.

sinisaos:
@adriangb Thanks for the suggestions @dantownsend I think that should be it.

Issue #421, opened 2022-02-03
Document extending `BaseUser`


dantownsend:
### Discussed in https://github.com/piccolo-orm/piccolo/discussions/420

<div type='discussions-op-text'>

<sup>Originally posted by **rogersjw** February  3, 2022</sup>
Hopefully the question is self-explanatory enough. I'm new to piccolo, coming from mainly using sqlalchemy previously. I love the 'battery included' BaseUser class and 'user' app. However, I'd like to build in multifactor authentication and would like to add a phone number column to the BaseUser class. What is the most appropriate way to do that while also still taking advantage of the nice 'user' app commands? I can't think of an elegant approach to make sure the migration get updated correctly and I don't break any of the user' app features. I've seen a suggestion to just create a new table with a foreign key but would rather modify the initial table instead for this other column that I will always consider mandatory. Thanks for any help!</div>

sinisaos:
@dantownsend If you want, I can add docs for extending `BaseUser`.

dantownsend:
@sinisaos If you don't mind, that would be great. I think we just document the `Profile` solution for now, until we figure put something better.

sinisaos:
@dantownsend OK. I will omit the part where the user can create a custom `BaseUser` from the default Piccolo `user` app. Although it is also useful.

sinisaos:
@dantownsend You can close this because it is done.

PR #422, opened 2022-02-03
add docs for a custom BaseUser

sinisaos:
Related to #421


dantownsend:
@sinisaos Thanks for this.

I decided to document that copying the user app was a solution, as you suggested.

PR #426, opened 2022-02-09
Fix overlapping, missing slots

ariebovenberg:
There are some misconfigured `__slots__`:

```
$ slotscheck -m piccolo
ERROR: 'piccolo.columns.combination:Combination' has slots but superclass does not.
ERROR: 'piccolo.columns.combination:Where' has slots but superclass does not.
ERROR: 'piccolo.query.methods.alter:AddColumn' defines overlapping slots.
ERROR: 'piccolo.query.methods.alter:SetDefault' defines overlapping slots.
Oh no, found some problems!
Scanned 145 module(s), 261 class(es).
```

Thankfully the fix was easy.

I discovered the slots issues with [slotscheck](https://github.com/ariebovenberg/slotscheck), a tool I maintain. Of course, If you like, I can add it to CI as I've done for [instagram/LibCST](https://github.com/Instagram/LibCST/pull/615), [sqlalchemy/sqlalchemy](https://github.com/sqlalchemy/sqlalchemy/pull/7670), and [aio-libs/aiohttp](https://github.com/aio-libs/aiohttp/pull/6547).


dantownsend:
@ariebovenberg This is great, thanks.

Slots are tricky to work with, and error prone, as your tool illustrates. `slotscheck` is impressively fast.

I think it's worthwhile adding `slotscheck` to Piccolo's test suite. If you're happy to do it, it just needs adding [here](https://github.com/piccolo-orm/piccolo/blob/master/scripts/lint.sh) and [here](https://github.com/piccolo-orm/piccolo/blob/master/requirements/dev-requirements.txt):

One feature you could potentially add in the future is emitting a warning for this situation:

```python
class Animal:
    __slots__ = ('colour',)

class Dog(Animal):
    pass  # Because it doesn't define slots, Python still assigns a dictionary, should be `__slots__ = ()`
```

This might already be supported, and I'm not aware.

ariebovenberg:
I added it to the lint script, but chose not to run it on `tests/`. This because tests are probably not meant to be imported/run outside of test context. `slotscheck` does support ignoring certain modules, but I felt excluding tests altogether was probably best. Let me know if you'd like to include it anyway, it should be easy to add an ignore for specific test modules that cannot be imported.

Regarding your idea about warnings: I'm very interested to hear any thoughts about slots 'enforcement' policies! Slotscheck does have a superstrict `require-subclass` mode, which does what you describe, but also counts `object` as a slotted class. Perhaps a 'softer' version of this rule could be useful 🤔 . I've opened https://github.com/ariebovenberg/slotscheck/issues/59 in case you have any thoughts about this.

dantownsend:
@ariebovenberg Thanks for this. I agree, there's no need to check the `tests/` folder.

I left a comment on the issue you opened.

dantownsend:
@ariebovenberg The only other suggestion I have for `slotscheck` is it feels to me like this should work:

```
slotscheck piccolo
```

But instead it's:

```
python -m slotscheck piccolo
```

Which is fine, but I think it would be more intuitive if both worked.

dantownsend:
@ariebovenberg Interesting - it makes a lot more sense now I've read your docs about it.

I can see your reasoning for doing it that way.

Issue #447, opened 2022-03-01
Feature request: make schema generate more verbose.


gmos:
It would be quite helpful when schema generate would let us know on which table and column it was working when it bails out.

dantownsend:
It runs a bunch of coroutines at the same time (one for each table in the database):

https://github.com/piccolo-orm/piccolo/blob/5aea0fdc4f40a99615f534ec49a38b242a5f5cb0/piccolo/apps/schema/commands/generate.py#L816

Originally it was sequential, but it took too long.

It could be modified so it's `asyncio.gather(*table_coroutines, return_exceptions=True)`. We can then work out which ones failed, and tell the user.

AliSayyah:
@dantownsend Hey there!
I would like to work on this, if possible.

gmos:
You already get the exception text printed.  Perhaps as a first step just put the table name in the text of the exceptions that Piccolo explicitly raises.

dantownsend:
@AliSayyah That would be great, thanks.

@gmos OK, makes sense.

AliSayyah:
@dantownsend Should we actually raise the exceptions gathered, or just handle it like warnings are handled?
maybe we can add an `exception` block?

dantownsend:
@AliSayyah Things get tricky with `asyncio.gather` with `return_exceptions=True`, because you can end up with 10 exceptions returned.

We could print out each exception, and then raise a single generic exception at the end like `Exception('Schema generation failed')`. Whatever you think is best.

AliSayyah:
> @AliSayyah Things get tricky with `asyncio.gather` with `return_exceptions=True`, because you can end up with 10 exceptions returned.
> 
> We could print out each exception, and then raise a single generic exception at the end like `Exception('Schema generation failed')`. Whatever you think is best.

Yeah, because of that, I raised a single exception that includes all the exceptions gathered with `asyncio.gather` .

```python
raise Exception([Exception("Error"), ...])
```

sinisaos:
Completed in #449

Issue #363, opened 2021-11-29
piccolo schema generate crash on index parse. 


gmos:
The following index DDL causes `piccolo schema generate` to crash.

```
CREATE UNIQUE INDEX sensor_uniq_in_time ON public.sensors USING btree (production_start_date, sensor_code) INCLUDE (production_end_date)
```
The regular expression used doesn't match more than one column name in the index and a failing match is not handled.  This resultsd in a: 'NoneType' object has no attribute 'groupdict'.

Moreover the INCLUDE clause is completely ignored.

dantownsend:
@gmos Thanks for reporting this issue.

I'm most concerned about it failing ungracefully. The schema generation isn't perfect, but as long as the user is alerted to issues, they can hand modify the schema when required.

gmos:
@dantownsend  Composite unique indices are quite common.  Fixing the recognition would not be a problem, but all index creation in Piccolo now seems to be tried to Column.  Won't work for composite indices.  Think a warning would do.
The INCLUDE clause is probably quite rare.  Warning would be ok too here.

dantownsend:
The situation has been improved by https://github.com/piccolo-orm/piccolo/pull/449

It won't crash on index parsing anymore, and will print out any problematic indexes which can't be parsed.

These are examples of problematic indexes I've encountered (as already mentioned, composite indexes are most common):

```sql
-- An index which references multiple columns
CREATE UNIQUE INDEX some_index ON some_table USING btree (column_1, column_2);

-- When `operator classes` are used
CREATE INDEX some_index ON some_table USING btree (column_1 varchar_pattern_ops);
```

At some point the regex needs improving, and Piccolo needs to support composite unique keys, and operator classes.

PR #449, opened 2022-03-04
get_output_schema exception with table name

AliSayyah:
get_output_schema will raise an exception with the name of the corresponding table if something goes wrong while generating the tables.
based on #447 

Another alternative would be to add the `errors` attribute to the  `OutputSchema` class and handle it like how warnings are being addressed.


AliSayyah:
The test works in python 3.10 (and 3.9 in local) but not with 3.7.
I have to figure out what functionality changed in these versions.

dantownsend:
I think something changed with mocks in > Python 3.7.

I've been caught out by this many times.

AliSayyah:
https://docs.python.org/3/library/unittest.mock.html#unittest.mock.AsyncMock

I think this is it. `create_table_class_from_db` is an async function and mock handles it in `>=3.8` as a coroutine.

dantownsend:
@AliSayyah If you search for `set_mock_return_value` in the Piccolo codebase, there's something there which should help.

AliSayyah:
@dantownsend I think this is ready for review,

dantownsend:
@AliSayyah Cool, thanks

dantownsend:
@AliSayyah This looks great thanks.

I pushed a few extra commits:

 * I decided to use your `AsyncMock` implementation everywhere, and removed `set_mock_return_value`, as I think your solution is more intuitive.
 * I took this opportunity to improve error handling if indexes couldn't be parsed (related to https://github.com/piccolo-orm/piccolo/issues/363)

AliSayyah:
Thank you this is great!

dantownsend:
@AliSayyah Yeah, it could do with a unit test really. I'm happy to release as is, but if you wouldn't mind doing a separate PR to add a test, that would be great.

AliSayyah:
> @AliSayyah Yeah, it could do with a unit test really. I'm happy to release as is, but if you wouldn't mind doing a separate PR to add a test, that would be great.

Yeah, I'll open another PR👍.

dantownsend:
@AliSayyah thanks

PR #455, opened 2022-03-07
comprehensions and literals

AliSayyah:
I found out that these cases can be optimized easily:

- `list(dictionary.items())` is much faster than `[i for i in dictionary.items()]`
- `{"key": value}` is much much faster than `dict(key=value)`
- using literals is a faster way of setting up empty data structures. for e.g: `dict() -> {}`, `tuple() -> ()` and so  on.

then I found out this awesome [flake8 plugin](https://github.com/adamchainz/flake8-comprehensions)  and changed them in piccolo's codebase. I'm not sure how much this would benefit the performance but it can't hurt.

sorry about the messed-up commits.

### DIFF
--- a/piccolo/apps/fixtures/commands/load.py
+++ b/piccolo/apps/fixtures/commands/load.py
@@ -24,9 +24,7 @@ async def load_json_string(json_string: str):
     fixture_configs = [
         FixtureConfig(
             app_name=app_name,
-            table_class_names=[
-                i for i in deserialised_contents[app_name].keys()
-            ],
+            table_class_names=list(deserialised_contents[app_name].keys()),
         )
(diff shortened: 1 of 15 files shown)


dantownsend:
@AliSayyah This looks very promising - thanks!

dantownsend:
@AliSayyah This all looks good to me 👍 

I'll merge it in and release it. Do you think it's worth adding this plugin somehow to our test suite?

AliSayyah:
> @AliSayyah This all looks good to me 👍 
> 
> I'll merge it in and release it. Do you think it's worth adding this plugin somehow to our test suite?

Thank you.
Yeah, I think it's pretty useful to have it in the stack.

PR #456, opened 2022-03-08
🪄  Chore(refactor): improve and Refactor Piccolo Functionality

yezz123:
hello @dantownsend Here is a little Pr while later, I try to read the changes in the latest versions of piccolo so, right now here are some of the updates to minify and enhance the code 🚀 

## What's Changed

- Inline variables that immediately returned.
- use the new pre-builtin `next`.
- back to use the `f-String` functions in some of the returns.
- the swap of if/else

- I guess Most of the tests gonna failed while the new way of refactoring is supporting the new built-in function that we have in py3.8


piccolo/apps/migrations/auto/schema_differ.py
```diff
@@ -291,21 +284,17 @@ def rename_tables(self) -> AlterStatements:
...
-                    table_class_name
-                )
-                snapshot_table = self.schema_snapshot_map.get(class_name)
-                if snapshot_table:
-                    snapshot_table.class_name = table_class_name
-                    return snapshot_table
+            )
+            if snapshot_table := self.schema_snapshot_map.get(class_name):
+                snapshot_table.class_name = table_class_name
+                return snapshot_table
```
dantownsend, line 297: This looks great, but we can't use the walrus operator at the moment because of Python 3.7 😞

yezz123: Yes, that was the issue, I was thinking to propose the drop of py3.7 but I guess it's better to wait until the new release of py3.11 and see if it's way much to drop the support on py3.7

dantownsend:
@yezz123 Thanks for this, it looks great. 👍 

There's just the issue with the walrus operator sadly.

I learned some new things from this PR. I didn't know the `raise exception from` syntax.

I also saw you replaced most default values which are mutable (lists and dicts) with `None`. I've been trying to do this more in my own code, as mutable defaults are an easy source of bugs.

piccolo/apps/migrations/auto/schema_differ.py
dantownsend: It has just over a year of life left: https://endoflife.date/python

There's an argument to remove it, but I can imagine some people being locked to Python 3.7 for weird corporate reasons.

yezz123:
> 

Amazing, yeah most of the time surfing the docs and seeing the last updates while trying them in some open sources projects!

I will fix the issue of the walrus operator and lets move on this way 🚀

piccolo/apps/migrations/auto/schema_differ.py
yezz123: > It has just over a year of life left: https://endoflife.date/python
> 
> There's an argument to remove it, but I can imagine some people being locked to Python 3.7 for weird corporate reasons.

Yeah same happen to me in some projects I work in with my old company, we use py3.7 and py3.6 in the testing, and its caused much of issues while refactoring the code and fix some bugs 🐛

yezz123:
@dantownsend i guess everything now its fine in this PR we could merge it

dantownsend:
@yezz123 This is great, thanks. Sorry for all the hassle with the walrus operator. 👍

yezz123:
> @yezz123 This is great, thanks. Sorry for all the hassle with the walrus operator. 👍 

No problem 🌶🎊

PR #459, opened 2022-03-11
add array column support to modelbuilder

backwardspy:
i have a model with an array field on it, and the model builder is crashing with the following error:
```python
>           random_value = cls.__DEFAULT_MAPPER[column.value_type]()
E           KeyError: <class 'list'>
```

this change adds basic support for array fields by generating a list of values of the appropriate types.

### DIFF
--- a/piccolo/testing/model_builder.py
+++ b/piccolo/testing/model_builder.py
@@ -4,7 +4,7 @@ from datetime import date, datetime, time, timedelta
 from decimal import Decimal
 from uuid import UUID
 
-from piccolo.columns.base import Column
+from piccolo.columns import Array, Column
 from piccolo.table import Table
 from piccolo.testing.random_builder import RandomBuilder
 from piccolo.utils.sync import run_sync
@@ -139,11 +139,18 @@ class ModelBuilder:
             Column class to randomize.
 
         """
+        random_value: t.Any
         if column.value_type == Decimal:
             precision, scale = column._meta.params["digits"]
             random_value = RandomBuilder.next_float(
                 maximum=10 ** (precision - scale), scale=scale
             )
+        elif column.value_type == list:
+            length = RandomBuilder.next_int(maximum=10)
+            base_type = t.cast(Array, column).base_column.value_type
+            random_value = [
+                cls.__DEFAULT_MAPPER[base_type]() for _ in range(length)
+            ]
         elif column._meta.choices:
             random_value = RandomBuilder.next_enum(column._meta.choices)
         else:
--- a/tests/testing/test_model_builder.py
+++ b/tests/testing/test_model_builder.py
@@ -1,6 +1,8 @@
 import asyncio
 import unittest
 
+from piccolo.columns import Array, Boolean, Integer, Varchar
+from piccolo.table import Table
 from piccolo.testing.model_builder import ModelBuilder
 from tests.example_apps.music.tables import (
     Band,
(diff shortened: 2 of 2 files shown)


piccolo/testing/model_builder.py
```diff
@@ -139,11 +139,18 @@ def _randomize_attribute(cls, column: Column) -> t.Any:
             Column class to randomize.
 
         """
+        random_value: t.Any
```
backwardspy, line 142: this stops mypy complaining on the following few lines where it thinks `random_value` is a float because of the first if-body, but then i set it to a list in the following one.

piccolo/testing/model_builder.py
```diff
@@ -139,11 +139,18 @@ def _randomize_attribute(cls, column: Column) -> t.Any:
...
         if column.value_type == Decimal:
             precision, scale = column._meta.params["digits"]
             random_value = RandomBuilder.next_float(
                 maximum=10 ** (precision - scale), scale=scale
             )
+        elif column.value_type == list:
+            length = RandomBuilder.next_int(maximum=10)
```
backwardspy, line 149: is 10 enough here? i don't think there's a way to limit array field lengths, otherwise i'd use that as the maximum.

piccolo/testing/model_builder.py
```diff
@@ -139,11 +139,18 @@ def _randomize_attribute(cls, column: Column) -> t.Any:
...
+            base_type = t.cast(Array, column).base_column.value_type
```
backwardspy, line 150: is `Array` the only column type based on lists?

tests/testing/test_model_builder.py
```diff
@@ -14,6 +16,12 @@
 )
 
 
+class TableWithArrayField(Table):
+    strings = Array(Varchar(30))
+    numbers = Array(Integer())
+    booleans = Array(Boolean())
```
backwardspy, line 22: i have constructed this here as i didn't necessarily want to edit the example tables and invoke a bunch of documentation changes et cetera! let me know if you'd prefer an array field on an existing example table.

tests/testing/test_model_builder.py
```diff
@@ -14,6 +16,12 @@
 )
 
 
+class TableWithArrayField(Table):
+    strings = Array(Varchar(30))
+    integers = Array(Integer())
+    floats = Array(Real())
```
backwardspy, line 22: i have constructed this here as i didn't necessarily want to edit the example tables and invoke a bunch of documentation changes et cetera! let me know if you'd prefer an array field on an existing example table.

piccolo/testing/model_builder.py
backwardspy, line 150: is `Array` the only column type based on lists?

piccolo/testing/model_builder.py
backwardspy, line 149: is 10 enough here? i don't think there's a way to limit array field lengths, otherwise i'd use that as the maximum.

piccolo/testing/model_builder.py
backwardspy, line 142: this stops mypy complaining on the following few lines where it thinks random_value is a float because of the first if-body, but then i set it to a list in the following one.

tests/testing/test_model_builder.py
dantownsend: That's fine 👍

piccolo/testing/model_builder.py
dantownsend: Yep, Array is the only one with `value_type = list`.

piccolo/testing/model_builder.py
dantownsend: I think is 10 is enough

piccolo/testing/model_builder.py
dantownsend: Cool, that's fine 👍

dantownsend:
The tests are just failing because PyPI is timing out for some reason. Probably a temporary issue - will run them again later on.

backwardspy:
awesome, thanks!

PR #468, opened 2022-03-24
Changed occurances "of" contraint in the repository to "constraint"

smythp:
I saw the typo contraint > constraint in the column documentation, and grepped on the project for other occurrences. There were a few in piccolo/query/methods/alter.py, plus one other occurrence in the docs. I ran the test suite, and it ran without errors after changing the _drop_contraint method and the code that uses it.

### DIFF
--- a/piccolo/apps/migrations/auto/migration_manager.py
+++ b/piccolo/apps/migrations/auto/migration_manager.py
@@ -457,7 +457,7 @@ class MigrationManager:
 
                 unique = params.get("unique")
                 if unique is not None:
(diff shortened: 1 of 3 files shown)


dantownsend:
@smythp Thanks a lot for this. I'll blame all these typos on my Mac's butterfly keyboard :)

smythp:
I'm the same way, there's something about programming that turns me into a bad speller. When I teach classes the students are always laughing at my spelling fails. Glad it was useful.

PR #498, opened 2022-04-29
📝 Add custom column type to doc

WintonLi:
Update doc - added "how to create custom column type" section to piccolo/schema/advanced

### DIFF
--- a/docs/src/piccolo/schema/advanced.rst
+++ b/docs/src/piccolo/schema/advanced.rst
@@ -242,7 +242,7 @@ will need to define your own column types.
 
 Generally there are two ways to define your own column types:
 
-* Create a subclass of an existing column type.
+* Create a subclass of an existing column type; or
 * Directly subclass the :ref:`Column <ColumnTypes>` class.
 
 Try to use the first method whenever possible because it is more straightforward and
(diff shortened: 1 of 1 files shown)


docs/src/piccolo/schema/advanced.rst
```diff
@@ -231,3 +231,58 @@ is not present, it will be reflected and returned.
 .. hint:: Reflection will automatically create ``Table`` classes for referenced
     tables too. For example, if ``Table1`` references ``Table2``, then
     ``Table2`` will automatically be added to ``TableStorage``.
+
+-------------------------------------------------------------------------------
+
+How to create custom column types
+------
```
dantownsend, line 238: Seems minor, but some Restructured Text linters complain if the underline isn't the same length as the title:

```
How to create custom column types
---------------------------------
```

dantownsend (commented):
Looks great! Thanks a lot of this. Just one minor suggestion.

docs/src/piccolo/schema/advanced.rst
WintonLi: Oh sorry didn't realize that. Will fix it now.

Issue #515, opened 2022-05-17
Be able to see the DDL generated by migrations


dantownsend:
### Discussed in https://github.com/piccolo-orm/piccolo/discussions/514

<div type='discussions-op-text'>

<sup>Originally posted by **theelderbeever** May 16, 2022</sup>
Throwing this out there as a possible suggestion. It would be potentially valuable to have an `--sql` option with the migration management where the output of `piccolo migrations new --sql` is a `.sql` file instead of using python code. This maybe useful in debugging and transparency into what will be run and how. I know of one other ORM that does this (Hasura) and it was nice to easily edit the migration.</div>

AliSayyah:
Hey @dantownsend.

dantownsend:
@AliSayyah Yes, that would be great.

We need to modify `MigrationManager` somehow so instead of running queries, it prints out the SQL.

https://github.com/piccolo-orm/piccolo/blob/master/piccolo/apps/migrations/auto/migration_manager.py

AliSayyah:
> @AliSayyah Yes, that would be great.
> 
> We need to modify `MigrationManager` somehow so instead of running queries, it prints out the SQL.
> 
> https://github.com/piccolo-orm/piccolo/blob/master/piccolo/apps/migrations/auto/migration_manager.py

Ok, thank you.

dantownsend:
Fixed by https://github.com/piccolo-orm/piccolo/pull/584

PR #516, opened 2022-05-17
Document the `migration` table

dantownsend:
Related to https://github.com/piccolo-orm/piccolo/issues/515


sinisaos:
@dantownsend I think we should close this PR because this is already done. ([here](https://github.com/piccolo-orm/piccolo/commit/66a53a3217c2a5f4b4772938d6ba534d5b94dfe7) and #1034)

dantownsend:
Yes, good point. I might just keep the bit I added to the docs about the migration table.

PR #527, opened 2022-05-24
add `tablespace` argument to `Table`

### LINKED ISSUE #526: Add support for tablespaces
### Discussed in https://github.com/piccolo-orm/piccolo/discussions/525

<div type='discussions-op-text'>

<sup>Originally posted by **theelderbeever** May 24, 2022</sup>
Background on the question... some of our tables are very large and our deployment infrastructure limits us to 16TB of disk space per provisioned drive. Having all of the tables on the same drive runs the risk of filling up our drive. We would like to set tables to tablespaces so we can provision drives on a per table/tablespace basis. 

Is there a way to `CREATE TABLE foo(i int) TABLESPACE space1;` when generating the ORM Models? 

Is the solution to generate empty tables using a raw SQL migration and then add the models in the next migration? Would Piccolo perform the correct `ALTER TABLE` commands if the tables weren't generated via the  ORM models?</div>

dantownsend:
Closes https://github.com/piccolo-orm/piccolo/issues/526

### DIFF
--- a/piccolo/apps/migrations/auto/diffable_table.py
+++ b/piccolo/apps/migrations/auto/diffable_table.py
@@ -92,6 +92,7 @@ class DiffableTable:
     tablename: str
     columns: t.List[Column] = field(default_factory=list)
     previous_class_name: t.Optional[str] = None
+    tablespace: t.Optional[str] = None
 
     def __post_init__(self):
         self.columns_map: t.Dict[str, Column] = {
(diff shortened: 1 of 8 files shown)


dantownsend:
This is pretty much done, but need to work out how to run the unit tests on a Postgres database which uses a custom tablespace.

lachiewalker:
Hi Dan,

> need to work out how to run the unit tests on a Postgres database which uses a custom tablespace

This is pretty easily done with testcontainers, which would also simplify test running in general. I set up a PR to do exactly that: #1410

PR #533, opened 2022-06-02
Docs for how to run async unit tests

dantownsend:
Discussion https://github.com/piccolo-orm/piccolo/discussions/530 made me realise that the docs for testing async Piccolo code were somewhat lacking.

This PR adds the following:

 * Describes the different ways of running async unit tests
 * Updates the docs for `ModelBuilder`

### DIFF
--- a/docs/src/piccolo/testing/index.rst
+++ b/docs/src/piccolo/testing/index.rst
@@ -177,7 +177,8 @@ like this:
         rows = await MyTable.select()
         assert len(rows) == 1
 
-If you prefer class based tests, then have a look at :class:`IsolatedAsyncioTestCase <unittest.IsolatedAsyncioTestCase>`
+If you prefer class based tests, and are using Python 3.8 or above, then have
+a look at :class:`IsolatedAsyncioTestCase <unittest.IsolatedAsyncioTestCase>`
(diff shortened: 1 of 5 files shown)


docs/src/piccolo/testing/index.rst
```diff
@@ -119,19 +127,64 @@ Alternatively, you can run the migrations to setup the schema if you prefer:
...
+from Python's standard library. You can then write tests like this:
+
+.. code-block:: python
+
+    from unittest import IsolatedAsyncioTestCase
+
+    class MyTest(IsolatedAsyncioTestCase):
+        async def test_select(self):
+            rows = await MyTable.select()
+            assert len(rows) == 1
```
sinisaos, line 190: @dantownsend You should add a disclaimer or warning that `unittest.IsolatedAsyncioTestCase` can only be used since Python version `3.8` and above ([docs](https://docs.python.org/3.8/library/unittest.html#unittest.IsolatedAsyncioTestCase)).
I already make same mistake in [Piccolo API PR](https://github.com/piccolo-orm/piccolo_api/pull/74)

dantownsend: Good point - I'll change it.

PR #540, opened 2022-06-11
add callbacks to queries

### LINKED ISSUE #344: Add callbacks to queries
It would be nice if there was a new clause which can be added to queries called `callback`.

For example:

```python
def callback_handler(success, value):
    if success:
        print(f'The query was successful and returned {value}')
    else:
        print("The query failed")

Band.select().callback(callback_handler).run_sync()
```

The callback could be a function or coroutine, which will be passed two values - a boolean for whether the query was successful, and the result of the query.

The reason this is useful is because it's quite common for a function to return a query, and not actually run it. But it would like to run something once the query is complete.

An example within Piccolo itself is the save method:

https://github.com/piccolo-orm/piccolo/blob/bffd3df448afd4bca123cab340f830984b42f0fb/piccolo/table.py#L316-L323

backwardspy:
resolves #344

here's my initial work on adding callbacks to queries as per #344

i would appreciate a code review before i take this any further! i also have a couple of questions:

1. at this stage you only get a callback when success = True, where & how would you suggest we detect a query failing and invoke the callback with success = False?
2. in `piccolo/query/base.py`, at the end of the `run` function, it branches based on whether there's more than one querystring or not. i was unable to cover the second branch with a test as i couldn't figure out how you'd end up with more than one querystring. would appreciate some pointers on how that bit works.

i look forward to hearing your thoughts.

### DIFF
--- a/piccolo/query/mixins.py
+++ b/piccolo/query/mixins.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import asyncio
 import typing as t
 from dataclasses import dataclass, field
 
@@ -235,6 +236,49 @@ class OutputDelegate:
     def copy(self) -> OutputDelegate:
         _output = self._output.copy() if self._output is not None else None
         return self.__class__(_output=_output)
+
+
+class SyncCallback(t.Protocol):
+    def __call__(self, success: bool, results: t.Any):
+        ...
+
+
+class AsyncCallback(t.Protocol):
+    async def __call__(self, success: bool, results: t.Any):
+        ...
+
+
+Callback = t.Union[SyncCallback, AsyncCallback]
+
+
+@dataclass
+class CallbackDelegate:
+    """
+    Example usage:
+
+    .callback(my_handler_function)
+    .callback(my_handler_coroutine)
+    .callback(lambda success, row: print(row))
+    """
+
+    _callback: t.Optional[Callback] = None
+
+    def callback(self, callback: Callback):
+        self._callback = callback
+
+    async def invoke(self, success: bool, results: t.Any):
+        """
+        Utility function that invokes the registered callback in the correct
+        way, handling both sync and async callbacks. If no callback is
+        registered, this function does nothing.
+        """
+        if self._callback is None:
+            return
+
+        if asyncio.iscoroutinefunction(self._callback):
+            await self._callback(success, results)
+        else:
+            self._callback(success, results)
 
 
 @dataclass
--- a/piccolo/query/base.py
+++ b/piccolo/query/base.py
@@ -12,7 +12,7 @@ from piccolo.utils.objects import make_nested_object
 from piccolo.utils.sync import run_sync
 
 if t.TYPE_CHECKING:  # pragma: no cover
-    from piccolo.query.mixins import OutputDelegate
+    from piccolo.query.mixins import CallbackDelegate, OutputDelegate
     from piccolo.table import Table  # noqa
 
 
@@ -198,18 +198,32 @@ class Query:
 
         querystrings = self.querystrings
 
+        callback: t.Optional[CallbackDelegate] = getattr(
+            self, "callback_delegate", None
+        )
+
         if len(querystrings) == 1:
             results = await engine.run_querystring(
                 querystrings[0], in_pool=in_pool
             )
-            return await self._process_results(results)
+            processed_results = await self._process_results(results)
+
+            if callback:
+                await callback.invoke(True, processed_results)
+
+            return processed_results
         else:
             responses = []
             for querystring in querystrings:
                 results = await engine.run_querystring(
                     querystring, in_pool=in_pool
                 )
-                responses.append(await self._process_results(results))
+                processed_results = await self._process_results(results)
+
+                if callback:
+                    await callback.invoke(True, processed_results)
+
+                responses.append(processed_results)
             return responses
 
     def run_sync(self, timed=False, in_pool=False, *args, **kwargs):
--- a/piccolo/query/methods/select.py
+++ b/piccolo/query/methods/select.py
@@ -12,6 +12,7 @@ from piccolo.columns.readable import Readable
 from piccolo.engine.base import Batch
 from piccolo.query.base import Query
 from piccolo.query.mixins import (
+    CallbackDelegate,
     ColumnsDelegate,
     DistinctDelegate,
     GroupByDelegate,
@@ -28,6 +29,7 @@ from piccolo.utils.warnings import colored_warning
 
 if t.TYPE_CHECKING:  # pragma: no cover
     from piccolo.custom_types import Combinable
+    from piccolo.query.mixins import Callback
     from piccolo.table import Table  # noqa
 
 
@@ -221,6 +223,7 @@ class Select(Query):
         "offset_delegate",
         "order_by_delegate",
         "output_delegate",
+        "callback_delegate",
         "where_delegate",
     )
 
@@ -243,6 +246,7 @@ class Select(Query):
         self.offset_delegate = OffsetDelegate()
         self.order_by_delegate = OrderByDelegate()
         self.output_delegate = OutputDelegate()
+        self.callback_delegate = CallbackDelegate()
         self.where_delegate = WhereDelegate()
 
         self.columns(*columns_list)
@@ -459,6 +463,10 @@ class Select(Query):
             load_json=load_json,
             nested=nested,
         )
+        return self
+
+    def callback(self, callback: Callback) -> Select:
+        self.callback_delegate.callback(callback)
         return self
 
     def where(self, *where: Combinable) -> Select:
--- a/tests/base.py
+++ b/tests/base.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import asyncio
 import sys
 import typing as t
 from unittest import TestCase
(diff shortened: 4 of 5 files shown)


backwardspy:
ah, the protocols i used for callback type hints don't seem to work in python 3.7. i'll fix those in the next revision.

dantownsend:
@backwardspy This looks great, thanks!

With the protocols, if you import `Protocol` from `typing_extensions` instead then it should work.

I'll respond to the rest of your questions shortly.

dantownsend:
These are good points - I might need to think it over a bit more, but these are my initial thoughts:

> at this stage you only get a callback when success = True, where & how would you suggest we detect a query failing and invoke the callback with success = False?

This is tricky. The query usually fails if an exception is raised by the database adapter (asyncpg). We could catch the exception, call the callback, and then re-raise the exception, but it could make the logic hard to follow.

Maybe we have different types of callback, and for now just add 'success' callbacks?

> in piccolo/query/base.py, at the end of the run function, it branches based on whether there's more than one querystring or not. i was unable to cover the second branch with a test as i couldn't figure out how you'd end up with more than one querystring. would appreciate some pointers on how that bit works.

I couldn't find any places we use multiple querystrings in the codebase any more. It's there in case you have an operation which requires multiple SQL queries. You're right - there's no code coverage for that logic. Don't worry about it for now I guess.

piccolo/query/mixins.py
```diff
@@ -237,6 +238,49 @@ def copy(self) -> OutputDelegate:
...
+
+    .callback(my_handler_function)
+    .callback(my_handler_coroutine)
+    .callback(lambda success, row: print(row))
+    """
+
+    _callback: t.Optional[Callback] = None
+
+    def callback(self, callback: Callback):
+        self._callback = callback
```
dantownsend, line 267: I think it would be cool if we could add multiple callbacks, So maybe it's `add_callback`, and `_callback` is a list.

backwardspy:
i like the idea of different callbacks.

what do you think of this as an API?

```python
def raise_if_none(result: Any):
    if result is None:
        raise HTTPException(
            status_code=HTTP_404_NOT_FOUND,
            detail="not found"
        )

def error_handler(ex: Exception):
    raise HTTPException(
        status_code=HTTP_500_INTERNAL_SERVER_ERROR,
        detail="an internal error occurred"
    ) from ex

await Band.select(Band.name).first().callback(raise_if_none).error_callback(error_handler)
```

we can keep this PR as just the success callback to start with.

on another note, i am interested in your thoughts on allowing callbacks to transform the query results, for example for unwrapping single-item results. i see some overlap with `.output(...)` in this idea, but i think there are probably cases where that capability could be useful.

backwardspy:
i have made some updates based on your feedback.

* callbacks now just take the results with no `success` boolean, with the idea of having a separate error callback later on
* you can now register multiple callbacks by passing a list, or by chaining `.callback(...)` calls. the original API still works too in that you can pass a single callback - i thought that looked cleaner than `.add_callback([handler])`.
* the protocol import is fixed and i moved the protocol definitions into the `TYPE_CHECKING` block at the top.

let me know what you think of these changes.

dantownsend:
> what do you think of this as an API?
> 
> def raise_if_none(result: Any):
>     if result is None:
>         raise HTTPException(
>             status_code=HTTP_404_NOT_FOUND,
>             detail="not found"
>         )
> 
> def error_handler(ex: Exception):
>     raise HTTPException(
>         status_code=HTTP_500_INTERNAL_SERVER_ERROR,
>         detail="an internal error occurred"
>     ) from ex
> 
> await Band.select(Band.name).first().callback(raise_if_none).error_callback(error_handler)

Yes, I think that works. Another options is to do something like:

```python
await Band.select(Band.name).callback(my_callback, on=Callback.success)
```

Where `Callback` is a enum - we could add things like `Callback.pre_run` in the future, which is triggered before the query runs.

> on another note, i am interested in your thoughts on allowing callbacks to transform the query results, for example for unwrapping single-item results. i see some overlap with .output(...) in this idea, but i think there are probably cases where that capability could be useful.

Yes, I was thinking about this too. I think it's definitely useful. It would mean each callback has to return the results, either modified or unmodified.

Another thing I was considering is whether it should be called 'callback'. In Piccolo API we use the term 'hook' a lot to refer to something similar ([for example](https://piccolo-api.readthedocs.io/en/latest/api_reference/index.html#loginhooks)). I had a look online and hook / callback / event are typically used interchangeably, so it's not really a big deal. I think callback is probably OK.

dantownsend:
> i have made some updates based on your feedback.
> 
> callbacks now just take the results with no success boolean, with the idea of having a separate error callback later on
> you can now register multiple callbacks by passing a list, or by chaining .callback(...) calls. the original API still works too in that you can pass a single callback - i thought that looked cleaner than .add_callback([handler]).
> the protocol import is fixed and i moved the protocol definitions into the TYPE_CHECKING block at the top.
> let me know what you think of these changes.

Yes, all of those changes sound perfect.

backwardspy:
i've pushed all the aforementioned.

i like your idea with the callback enum type, definitely cleaner than having a different function for each type. if you think that's the way to go, i'll implement that next. i think it's worth getting it in now so we don't have to change the API later (that said, i guess we could just default it to `success` to maintain backward compatibility.)

dantownsend:
> i like your idea with the callback enum type, definitely cleaner than having a different function for each type. if you think that's the way to go, i'll implement that next. i think it's worth getting it in now so we don't have to change the API later (that said, i guess we could just default it to on=Callback.success to maintain backward compatibility.)

I think it would be cool to have if you don't mind implementing it, but it can come later.

backwardspy:
> > i like your idea with the callback enum type, definitely cleaner than having a different function for each type. if you think that's the way to go, i'll implement that next. i think it's worth getting it in now so we don't have to change the API later (that said, i guess we could just default it to on=Callback.success to maintain backward compatibility.)
> 
> I think it would be cool to have if you don't mind implementing it, but it can come later.

more than happy to.

backwardspy:
i've just pushed two more commits. the first implements different kinds of callbacks via an enum like you suggested.

the second commit is an experiment to see what it would look like to allow the callbacks to transform data. i think it's quite elegant as it allows you to process query results in a functional manner. please let me know what you think of this, i can drop it from this PR if it's too much complexity for now.

dantownsend:
> the second commit is an experiment to see what it would look like to allow the callbacks to transform data. i think it's quite elegant as it allows you to process query results in a functional manner. please let me know what you think of this, i can drop it from this PR if it's too much complexity for now.

I like that - looks good 👍

piccolo/query/mixins.py
```diff
@@ -237,6 +249,56 @@ def copy(self) -> OutputDelegate:
...
+    """
+    Example usage:
+
+    .callback(my_handler_function)
+    .callback(print, on=Callback.success)
+    .callback(my_handler_coroutine)
+    .callback([handler1, handler2])
+    """
+
+    _callbacks: t.List[RegisteredCallback] = field(default_factory=list)
```
dantownsend, line 263: One suggestion here - I think it works well, but rather than having to loop through to find the right callbacks, it could be something like this instead, so it's easier to directly retrieve the callbacks we're interested in:

```python
@dataclass
class RegisteredCallbacks:
    success: t.List[Callback] = field(default_factory=list)

class CallbackDelegate:
    _callbacks: RegisteredCallbacks = field(default_factory=RegisteredCallbacks)

# When registering
self._callbacks.success.append(some_function)

# So then we can quickly get the callbacks we want
self._callbacks.success
```

Or maybe a dictionary is better:

```python
class CallbackDelegate:
    _callbacks: t.Dict[Callback, t.List[RegisteredCallback]] = field(default_factory={i: [] for i in Callback})

# When registering
self._callbacks[Callback.success].append(some_function)

# So then we can quickly get the callbacks we want
self._callbacks[Callback.success]
```

What do you think about renaming the `Callback` enum to something like `CallbackType`, or `InvokedOn`? Someone might assume `Callback` corresponds to a function.

backwardspy: i was thinking the loop & filter was a bit unwieldy. i think i like the dict storage best, otherwise we'll have to map the `kind` argument to the correct `RegisteredCallbacks` property ourselves.

RE changing the name, absolutely agree it's a bit confusing. changing that would also let me change `RegisteredCallback` back to just `Callback`, which read a bit nicer.

dantownsend: You're right - I think the dict is cleaner.

backwardspy:
i've renamed `Callback` to `CallbackType`, `RegisteredCallback` back to `Callback`, and stored them in a dict as discussed.

dantownsend:
It's looking good 👍 

At the moment we just have it for select queries - but it could probably work for every query type. This could come in the future though.

The only other thing is adding docs - just something simple like this:

https://piccolo-orm.readthedocs.io/en/latest/piccolo/query_types/select.html#columns

backwardspy:
i've added callbacks to objects queries because i figure those are similar enough to select queries that it's pretty low risk. let me know if you'd rather this happened in a separate PR.

i am writing up some docs now.

dantownsend:
> i've added callbacks to objects queries because i figure those are similar enough to select queries that it won't rock the boat too much. let me know if you'd rather this happened in a separate PR.

Good idea - I think adding it for objects makes sense, and leaving the others for now.

> i am writing up some docs now.

Amazing, thanks!

backwardspy:
i've just pushed my first draft of the docs. i've tried to match your style where i can, eager to get some feedback on it.

the build failure is an odd one, those tests pass locally and i'm not sure what the difference is. i'll play around with it and see if i can replicate what the CI is doing.

dantownsend:
> i've just pushed my first draft of the docs. i've tried to match your style where i can, eager to get some feedback on it.

The docs are perfect - thanks!

> the build failure is an odd one, those tests pass locally and i'm not sure what the difference is. i'll play around with it and see if i can replicate what the CI is doing.
> 
> edit: missed the fact it's on python 3.7. might be a difference in how mock call args work, looking into it now.

Yes, I've encountered similar issues before - I think you're right, and mocks are different in Python 3.7.

backwardspy:
`mock.call_args.args` was added in 3.8. `mock.call_args[0]` does the same thing. fixed!

dantownsend:
Cool - it all looks good 👍 

Thanks for all your work on this.

PR #557, opened 2022-07-03
fix _get_related_readable typo

sinisaos:
Fix for very subtle typo in `_get_related_readable` causing [Piccolo Admin issue](https://github.com/piccolo-orm/piccolo_admin/issues/170)


dantownsend:
Damn, good catch!

This is probably the cause of this issue: https://github.com/piccolo-orm/piccolo_admin/issues/170

sinisaos:
@dantownsend Yes, with this typo fix everything works great in Piccolo Admin.

PR #558, opened 2022-07-06
Return querystring string cast when casting Combination

destos:
Ran into this problem locally while trying to debug an And combination.


dantownsend:
@destos Good catch - thanks!

Issue #561, opened 2022-07-15
Feature Request: typehints (autocompletes) for query result


Bobronium:
> Tab completion support - works great with iPython and VSCode.

While autocompletion does work when building the query, unfortunately, it does not work for query results:

<img width="683" alt="image" src="https://user-images.githubusercontent.com/36469655/179208414-93fd2837-b848-42e7-aa1e-98f572d87725.png">

I can imagine that it can be nontrivial to implement, but SQLModel proved that it's possible:
Screenshot from https://github.com/tiangolo/sqlmodel/pull/58:
![image](https://user-images.githubusercontent.com/36469655/179210564-10bead41-483b-4c85-99ad-2cb52c39ccb5.png)

dantownsend:
@Bobronium  You're right, there's a few places where tab completion needs improving.

When I originally wrote the library, both iPython and VSCode used the [jedi](https://jedi.readthedocs.io/en/latest/) library under the hood for tab completion, and it gave pretty solid results. But now VSCode uses Pylance. And I think Pycharm uses something proprietary.

What I do as a work around is this:

```python
import typing as t

band: t.Optional[Band] = await Band.objects().first()
```

I have some ideas for how to fix it using Python's `TypeVar` - need to have a play around!

dantownsend:
I had a play around, and managed to get it working using a combination of `Generic` and `TypeVar`.

Still needs a bit of work, but at least we know it's possible.

PR #562, opened 2022-07-15
improve typing for `await Band.objects().first()`

dantownsend:
Resolves https://github.com/piccolo-orm/piccolo/issues/561

When doing this query:

```python
band = await Band.objects().first()
```

MyPy now knows the response is `t.Optional[Band]`. This was possible by using `Generic` and `TypeVar` from Python's `typing` module.


Bobronium:
I propose to use TypeVar("ClassT", bound="Class") instead of Self, since Self is relatively new and not supported by mypy/pycharm, afaik.

PR #569, opened 2022-07-20
Postgres schema support

dantownsend:
Resolves https://github.com/piccolo-orm/piccolo/issues/817

Needs the following doing:

- [x] Fix the schema generation tests
- [x] Add a bunch more tests
- [x] Allow the schema to be changed in `alter` queries
- [x] Update docs for `alter`
- [x] Be able to move tables between schemas in auto migrations
- [x] Write tests for migrations
- [x] Add a new section about schemas to the docs


dantownsend:
I made this prototype a while ago, but the approach seems solid - just need to update a bunch of schema generation tests.

Will finish it off - related to https://github.com/piccolo-orm/piccolo/issues/817

Issue #578, opened 2022-08-08
Can a db extra node be passed to batch query executions?


guruvignesh01:
While checking the batch documentation https://piccolo-orm.readthedocs.io/en/latest/piccolo/query_clauses/batch.html#batch, I couldn't find any configuration for passing read replica db connection as an extra node.

sinisaos:
@guruvignesh01 Not sure but I think this should work ([docs](https://piccolo-orm.readthedocs.io/en/latest/piccolo/engines/postgres_engine.html#source)) because you can pass `node` argument to [run() method](https://github.com/piccolo-orm/piccolo/blob/master/piccolo/query/base.py#L167).
```python
async with await YourTable.select().batch(batch_size=100).run(node="your_read_replica") as batch:
    async for _batch in batch:
        print(_batch)
```

dantownsend:
Yeah, it doesn't work at the moment unfortunately.

It's not a huge change to Piccolo to make it work, so hopefully we can add it soon.

dantownsend:
We just need to change this:

https://github.com/piccolo-orm/piccolo/blob/0c44bb22be451e27318ddc2594e7e5a775494f80/piccolo/engine/postgres.py#L393-L397

So instead it's something like:

```python
    async def batch(self, query: Query, batch_size: int = 100, node: t.Optional[str] = None) -> AsyncBatch:
        engine = self.extra_nodes.get(node) if node else self 
        connection = await engine.get_new_connection()
        return AsyncBatch(
            connection=connection, query=query, batch_size=batch_size
        )
```

sinisaos:
@guruvignesh01 @dantownsend Sorry for the wrong answer. I didn't know that.

dantownsend:
@sinisaos No worries - what you suggested made sense.

dantownsend:
@sinisaos Do you want to have a go at a PR for this? No pressure - I might be able to do it later today.

sinisaos:
@dantownsend No problem, I can try.

guruvignesh01:
Thanks @sinisaos @dantownsend

dantownsend:
Fixed by https://github.com/piccolo-orm/piccolo/pull/579

PR #579, opened 2022-08-08
add node argument to batch method

sinisaos:
Related to #578

### DIFF
--- a/docs/src/piccolo/query_clauses/batch.rst
+++ b/docs/src/piccolo/query_clauses/batch.rst
@@ -20,6 +20,19 @@ responses.
         async for _batch in batch:
             print(_batch)
 
+There is also an option to pass a additional database nodes (e.g. read replicas) as a 
+node argument to ``batch`` method.
+
+.. code-block:: python
+
+    # Returns 100 rows at a time from read_replica_db
+    async with await Manager.select().batch(
+        batch_size=100,
+        node="read_replica_db",
+    ) as batch:
+        async for _batch in batch:
+            print(_batch)
+
 There's currently no synchronous version. However, it's easy enough to achieve:
 
 .. code-block:: python
--- a/docs/src/piccolo/query_clauses/batch.rst
+++ b/docs/src/piccolo/query_clauses/batch.rst
@@ -8,6 +8,9 @@ You can use ``batch`` clauses with the following queries:
 * :ref:`Objects`
 * :ref:`Select`
 
+Example
+-------
+
 By default, a query will return as many rows as you ask it for. The problem is
 when you have a table containing millions of rows - you might not want to
 load them all into memory at once. To get around this, you can batch the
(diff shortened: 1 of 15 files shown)


dantownsend:
@sinisaos This is great - thanks!

We just need to add a test somehow. Maybe we can add an extra test to this file:

https://github.com/piccolo-orm/piccolo/blob/master/tests/engine/test_extra_nodes.py

It will be very similar, it just uses `batch`.

sinisaos:
@dantownsend Will this be enough? 
```python
from piccolo.utils.sync import run_sync

# Make sure the node is queried in batch
run_sync(Manager.select().batch(node="read_1"))
self.assertTrue(EXTRA_NODE.run_querystring.called)
```
Tests pass with it.

dantownsend:
@sinisaos Yes, I think that's OK.

docs/src/piccolo/query_clauses/batch.rst
```diff
@@ -20,6 +20,19 @@ responses.
         async for _batch in batch:
             print(_batch)
 
+There is also an option to pass a additional database nodes (e.g. read replicas) as a 
+node argument to ``batch`` method.
```
dantownsend, line 24: We should mention it's just for Postgres at the moment.

sinisaos: Great, thanks. I will change that.

I think the test example I wrote in the comments is not good because I don't think it does anything. There is currently no synchronous version of batch and I don't know how to test this properly. Can you check that?

dantownsend: Yes, I'll have a look.

dantownsend: Didn't have time to finish looking into this - will look in the morning.

dantownsend:
I just realised `Objects` and `Select` have a `batch` method, which needs to accept the `node` argument, and pass it through to the `Engine.batch` method.

sinisaos:
I'm pass a test database in `test_batch.py` as a `read_replica` to be able to test if batch method node argument is called. Locally all tests passed but failed on Sqlite in GH actions.

sinisaos:
@dantownsend I was finally able to pass all the tests in GH actions by importing the test database instance into the local namespace of the test method. Can you check if that is ok?

dantownsend:
@sinisaos Yeah, it's tricky to test. I'm having a look.

dantownsend:
@sinisaos I changed the testing approach slightly. What you did was good. The only problem is we were modifying the global `DB` variable - which would probably be OK, but I changed it so we use a local DB variable, so there's less chance of other tests being effected.

I see that you added some coverage config to `pyproject.toml` - it's a good idea. I just removed 'ValueError' because I'm worried there may be some legitimate tests which might be missing - all `ValueError` should be covered really.

sinisaos:
@dantownsend Your approach is much better, and everything looks good. Thanks.

PR #582, opened 2022-08-11
Multiple Unique Constraints with auto mirtaions

northpowered:
Full description is in pinned issue
ADDED:

1. UniqueConstraint Class with the same behavior as a Column
2. CREATE and DROP auto migrations
3. A bit of doc for a UniqueConstraint Class
4. Updated tests

IMPORTANT!
There`re no ALTER CONSTRAINT option, but will be added soon, I hope.
For changing constraint I recommend do it with 2 migrations: DROP->CREATE


dantownsend:
@northpowered Thanks for this, it looks promising!

I think the linters are currently failing because of isort.

northpowered:
> @northpowered Thanks for this, it looks promising!
> 
> I think the linters are currently failing because of isort.

Mb because we are using different linters
Due to testing, I failed only one test, but it`s the similar to the beginning of my work :D
Also #583 has some details

piccolo/columns/constraints.py
```diff
@@ -0,0 +1,42 @@
...
+        class FooTable(Table):
+            foo_field = Text()
+            bar_field = Text()
+            my_constraint_1 = UniqueConstraint(['foo_field','bar_field'])
+
+    SQL queries for creating and dropping constrains are similar to:
+        ALTER TABLE foo_table ADD CONSTRAINT my_constraint_1 UNIQUE (foo_field, bar_field);
+        ALTER TABLE foo_table DROP IF EXIST CONSTRAINT my_constraint_1;
+    """
+    def __init__(self, unique_columns: list[str]) -> None:
```
dantownsend, line 23: I think this is why the tests are failing.

We can't use this newer syntax yet, because we're keeping backwards compatibility with Python 3.7.

So it'll have to be `Typing[str]` instead.

piccolo/query/methods/alter.py
```diff
@@ -176,6 +176,17 @@ class SetLength(AlterColumnStatement):
     def ddl(self) -> str:
         return f'ALTER COLUMN "{self.column_name}" TYPE VARCHAR({self.length})'
 
+@dataclass
+class AddUniqueConstraint(AlterStatement):
+    __slots__ = ("constraint_name","columns")
+
+    constraint_name: str
+    columns: list[str]
```
dantownsend, line 184: Same as above -> `List[str]`

piccolo/query/methods/alter.py
```diff
@@ -433,6 +446,12 @@ def _get_constraint_name(self, column: t.Union[str, ForeignKey]) -> str:
         tablename = self.table._meta.tablename
         return f"{tablename}_{column_name}_fk"
 
+    def add_unique_constraint(self, constraint_name: str, columns: list[str]):
```
dantownsend, line 449: See above.

dantownsend:
Sorry it has taken me so long - let's try and get these tests passing.

piccolo/columns/constraints.py
northpowered: Sorry, I used to Python3.10, so I forgot about older versions

northpowered:
@dantownsend 
I changed list[...] to t.List[...] to keep this compatible with Python3.7
Also upgraded some files up to v.0.109.0

sinisaos:
Closed in favor of #1146.

PR #584, opened 2022-08-11
migration DDL preview functionality

AliSayyah:
Hey there!
Just some cleanups and initial base stuff.
Doesn't work right now.

### DIFF
--- a/piccolo/apps/migrations/auto/migration_manager.py
+++ b/piccolo/apps/migrations/auto/migration_manager.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import inspect
 import typing as t
 from dataclasses import dataclass, field
+from pprint import pprint
 
 from piccolo.apps.migrations.auto.diffable_table import DiffableTable
 from piccolo.apps.migrations.auto.operations import (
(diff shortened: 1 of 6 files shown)


dantownsend:
@AliSayyah This looks great, thanks!

It's exactly what I had in mind.

AliSayyah:
I'll work on more tests, but how do you think we should handle the command?
```

**Comment by dantownsend** [comment]
> I'll work on more tests, but how do you think we should handle the command?
> 
> piccolo migrations forwards my_app --preview
> piccolo migrations backwards 2018-09-04T19:44:09 --preview
> Would this work?

Yes, I like that 👍

**Comment by AliSayyah** [comment]
Does `targ` support command options like `--preview`?

**Comment by AliSayyah** [comment]
Hey, @dantownsend. This seems ok to me. But let me know if there are any problems, or the tests are insufficient.

File: piccolo/apps/migrations/auto/migration_manager.py [hunk 1]
```diff
@@ -739,16 +770,17 @@ async def run_backwards(self):
...
-                else:
-                    raw()
-
-            await self._run_add_columns(backwards=True)
-            await self._run_add_tables(backwards=True)
-            await self._run_drop_tables(backwards=True)
-            await self._run_rename_tables(backwards=True)
-            await self._run_drop_columns(backwards=True)
-            await self._run_rename_columns(backwards=True)
-            await self._run_alter_columns(backwards=True)
```
dantownsend, line 754: It looks good - my only concern was this bit.

You've merged `run` and `run_backwards` - I can understand why, as it reduces repetition. The order of the operations were slightly different before when running backwards. Having thought about it though, I think the new order should also be fine.

dantownsend:
@AliSayyah This looks good to me, thanks.

I think we should release it.

PR #592, opened 2022-08-18
Fixes Piccolo backwards raw migrations

translunar:
Previously, Piccolo would just run the forwards raw migrations again instead of running the `raw_backwards` migrations on the `MigrationManager`. This patch forces Piccolo to explicitly run forwards or backwards raw migrations. It includes two modifications to existing unit tests, which were also incorrect.


dantownsend:
@translunar Thanks a lot for this! 👍

Issue #593, opened 2022-08-19
Email column type


dantownsend:
We don't have an email column type at the moment. It's because Postgres doesn't have an email column type.

However, I think it would be useful to designate a column as containing an email.

## Option 1

We can define a new column type, which basically just inherits from `Varchar`:

```python
class Email(Varchar):
    pass
```

## Option 2

Or, let the user annotate a column as containing an email.

One option is using [Annotated](https://docs.python.org/3/library/typing.html#typing.Annotated), but the downside is it was only added in Python 3.9:

```python
from typing import Annotated

class MyTable(Table):
    email: Annotated[Varchar, 'email'] = Varchar()
```

Or something like this instead:

```python
class MyTable(Table):
    email = Varchar(validators=['email'])
```

## Benefits

The main reason it would be useful to have an email column is when using [create_pydantic_model](https://piccolo-orm.readthedocs.io/en/latest/piccolo/serialization/index.html#create-pydantic-model) to auto generate a Pydantic model, we can set the type to `EmailStr` (see [docs](https://pydantic-docs.helpmanual.io/usage/types/)).

## Input

Any input is welcome - once we decide on an approach, adding it will be pretty easy.

sinisaos:
@dantownsend That's a great idea. I think option 1 is valid and the easiest to implement. After that we just need to add 
```python
elif isinstance(column, Email):
    value_type = pydantic.EmailStr
elif isinstance(column, Varchar):
    value_type = pydantic.constr(max_length=column.length)
```
to `pydantic.py` and `pydantic[email]` to requirements. With that we could have email validation in Piccolo API and Piccolo Admin out of the box.

dantownsend:
@sinisaos I agree - option 1 is probably the best.

sinisaos:
@dantownsend If you want I can try to add this. This way later we can also add `ip`(`IPvAnyAddress`) or [url](https://pydantic-docs.helpmanual.io/usage/types/#urls) if some user need that pydantic validator in `create_pydantic_model`.

dantownsend:
@sinisaos That would be great, thanks

dantownsend:
Fixed by https://github.com/piccolo-orm/piccolo/pull/595

PR #595, opened 2022-08-20
add email column

sinisaos:
Related to #593


requirements/requirements.txt
```diff
@@ -5,3 +5,4 @@ targ>=0.3.7
 inflection>=0.5.1
 typing-extensions>=3.10.0.0
 pydantic>=1.6
+email-validator>=1.2.1
```
dantownsend, line 8: Can we do anything like `pydantic[email]` instead?

sinisaos: Yes. I will change that.

dantownsend: Cool, thanks

dantownsend:
@sinisaos I made some minor tweaks. When the tests pass I'll release it.

PR #599, opened 2022-08-25
Reverse lookup

### LINKED ISSUE #378: Reverse foreign key lookups
Now we have M2M, it makes sense to add something similar for reverse foreign key lookups.

This needs doing before v1.

Here's an example API design:

```python

class Manager(Table):
    name = Varchar()
    bands = ReverseLookup(LazyTableReference('Band', module_path=__module__))


class Band(Table):
    name = Varchar()
    manager = ForeignKey(Manager)


>>> await Manager.select(Manager.name, Manager.bands(Band.name, as_list=True))
[
    {
        "name": "Guido",
        "bands": ["Pythonistas"]
    }
]
```

sinisaos:
Related to #378. Prototype for reverse foreign key lookup. Any suggestions and help are welcome. If I missed the point or it's not a good api design feel free to close PR.
Usage of reverse FK lookup:
```python
from piccolo.columns import Varchar, ForeignKey
from piccolo.columns.reference import LazyTableReference
from piccolo.columns.reverse_lookup import ReverseLookup

class Manager(Table):
    name = Varchar()
    artists = ReverseLookup(
        LazyTableReference("Artist", module_path=__name__), reverse_fk="manager"
    )

class Artist(Table):
    name = Varchar()
    manager = ForeignKey(Manager)

rows = Manager.select(
        Manager.name, Manager.artists() # or as list Manager.artists(Artist.name, as_list=True)
    ).run_sync()
```


dantownsend:
Thanks for this! It looks very promising.

sinisaos:
Thanks. I used a lot of code (with some modifications) from M2M fields. I tried to use the best names of variables and methods. The main obstacle is that the user would have to enter `reverse_lookup_name`. I will try to make that optional (if table has only one `FK` default index value for `foreign_key_columns` will be `0`), so that the user has to specify `reverse_lookup_name` only if the table has multiple `FK` so `ReverseLookup` knows which column to use. If you have any ideas, any help is welcome.

powellnorma:
Just wanted to note that there is interest in this functionality. Is there anything specific holding this back (from being merged)?

Also I like how in peewee one can just set the `backref` parameter on the `ForeignKeyField`, see [here](https://docs.peewee-orm.com/en/latest/peewee/relationships.html?highlight=backref#model-definitions)
Could we have a similar API in piccolo?

piccolo/columns/reverse_lookup.py
```diff
@@ -0,0 +1,226 @@
...
+    @property
+    def foreign_key_columns_index(self) -> int:
+        reverse_lookup_table = (
+            self.reverse_lookup._meta.resolved_reverse_joining_table
+        )
+        fk_columns = [
+            i._meta.name
+            for i in reverse_lookup_table._meta.foreign_key_columns
+        ]
+        return fk_columns.index(self.table._meta.tablename)
```
powellnorma, line 69: It looks like we compare column names with table names here. They don't have to match. I think instead of
```
Manager.artists(table=Manager)
``` 
It would be better to be able to specify:
```
Manager.artists(column="manager")
``` 

Since the column that is used for the reverse lookup doesn't change(?), I think it would be best if we could also specify it in the `ReverseLookup` constructor.

dantownsend:
> Just wanted to note that there is interest in this functionality. Is there anything specific holding this back (from being merged)?

I haven't had chance to review it yet - but I agree that it's a feature we need.

piccolo/columns/reverse_lookup.py
```diff
@@ -0,0 +1,226 @@
...
+    def get_select_string(self, engine_type: str, with_alias=True) -> str:
+        reverse_lookup_table = (
+            self.reverse_lookup._meta.resolved_reverse_joining_table
+        )
+        reverse_lookup_table_name = reverse_lookup_table._meta.tablename
+        reverse_lookup_pk = reverse_lookup_table._meta.primary_key._meta.name
+        reverse_lookup_fk_table_name = (
+            reverse_lookup_table._meta.foreign_key_columns[
+                self.foreign_key_columns_index
+            ]._meta.name
```
powellnorma, line 80: I think instead of `_meta.name` it should be `_meta.table._meta.tablename`. Otherwise `reverse_lookup_fk_table_name` is a column name

piccolo/columns/reverse_lookup.py
```diff
@@ -0,0 +1,226 @@
...
+
+        if engine_type == "postgres":
+            if self.as_list:
+                column_name = self.columns[0]._meta.db_column_name
+                return f"""
+                    ARRAY(
+                        SELECT
+                            "{reverse_lookup_table_name}"."{column_name}"
+                        FROM {reverse_select}
+                    ) AS "{reverse_lookup_table_name}s"
```
powellnorma, line 103: I think we should use `AS "{original_column_name}s"`. The table must give us that name. See m2m.py: https://github.com/piccolo-orm/piccolo/blob/14012e59759c849ad732b4c4e4fb40636dc88aa6/piccolo/columns/m2m.py#L93

piccolo/columns/reverse_lookup.py
sinisaos: @powellnorma Thanks for your interest and review. It's been a while since I wrote this code, but if I remeber it has to be a table because we refer to the foreign key of parent table later in the code. We need this function to find the foreign key column index if there are multiple foreign keys per single table.

sinisaos:
@dantownsend It would be great if you could find the time to review this. @powellnorma This code works and passes all tests. If you think that things should be changed, it would be best after Daniel's review. Thank you both.

piccolo/columns/reverse_lookup.py
```diff
@@ -0,0 +1,226 @@
...
+        ]._meta.db_column_name
+
+        reverse_select = f"""
+            "{reverse_lookup_table_name}"
+            JOIN "{reverse_lookup_fk_table_name}" "inner_{reverse_lookup_fk_table_name}" ON (
+                "{reverse_lookup_table_name}"."{reverse_lookup_fk}"
+                = "inner_{reverse_lookup_fk_table_name}"."{reverse_lookup_pk}"
+            ) WHERE "{reverse_lookup_table_name}"."{reverse_lookup_fk}"
+                = "{reverse_lookup_fk_table_name}"."{reverse_lookup_pk}"
+        """  # noqa: E501
```
powellnorma, line 93: Why do we need this `JOIN` here? Isn't the following enough?
```python
        reverse_select = f"""
            "{reverse_lookup_table_name}"
            WHERE "{reverse_lookup_table_name}"."{reverse_lookup_fk}"
                = "{reverse_lookup_fk_table_name}"."{reverse_lookup_pk}"
        """
```

powellnorma:
> This code works and passes all tests

I haven't looked, but I guess the tests always use a ForeignKey named the same as referenced the Table. As e.g. here:
```python
class Manager(Table):
    name = Varchar()

class Artist(Table):
    name = Varchar()
    manager = ForeignKey(Manager)
```

But if this is not true (as in my case), as e.g. here:
```python
```python
class MainManager(Table):
    name = Varchar()

class Artist(Table):
    name = Varchar()
    manager = ForeignKey(Manager)
```
the sections I pointed lead to errors / wrong queries

powellnorma:
When I query an M2M field of the referenced table, I get:
```python
AttributeError: type object 'M2MSelect' has no attribute 'value_type'
```

powellnorma:
I have fixed the parts that I reviewed on my branch `reverse_lookup`, see: https://github.com/powellnorma/piccolo/commits/reverse_lookup

However querying an M2M field seems to be rather slow at the moment. When I set `self.serialisation_safe = False` in the case of the above mentioned `AttributeError`, it seems that all m2m joining keys are being used as arguments to a second query. This is overall (in my case) slower than using a `for` loop in python (which is located in a transaction) instead of the `ReverseLookup`.
@dantownsend what could we do to make this (querying m2m of reverse-referenced table) faster?

sinisaos:
> But if this is not true, as e.g. here:
> 
> ```python
> class MainManager(Table):
>     name = Varchar()
> 
> class Artist(Table):
>     name = Varchar()
>     manager = ForeignKey(Manager)
> ```
> 
> the sections I pointed lead to errors / wrong queries

I'm sorry, but I don't understand. How can we refer to a table that doesn't exist. Shouldn't it be like this
```python
class MainManager(Table):
    name = Varchar()

class Artist(Table):
    name = Varchar()
    manager = ForeignKey(MainManager)
```
I want to try your branch. In my reverse lookup attempt I use `LazyTableReference` like this
```python
class Manager(Table):
    name = Varchar()
    bands = ReverseLookup(
        LazyTableReference("Band", module_path=__name__),
    )


class Band(Table):
    name = Varchar()
    manager = ForeignKey(Manager)

# query for all managers bands
query = await Manager.select(
        Manager.name, Manager.bands(Band.name, as_list=True, table=Manager)
    )
```
Can you show me how to write a table schema and an example query because right now I get `TypeError: __init__() missing 1 required positional argument: 'reverse_fk'`. Thanks in advance

powellnorma:
> I'm sorry, but I don't understand. How can we refer to a table that doesn't exist.

The example I gave was wrong, I corrected it. What I meant was the situation in which column name (of the foreign key) has a different name than the table that it refers to

> Can you show me how to write a table schema and an example query because right now I get TypeError: __init__() missing 1 required positional argument: 'reverse_fk'.

Like this:

```python
class MainManager(Table):
    name = Varchar()
    artists = ReverseLookup(LazyTableReference("MainManager", module_path=__name__), "manager")

class Artist(Table):
    name = Varchar()
    manager = ForeignKey(MainManager)

# select:
await MainManager.select(MainManager.artists(Artist.name))
```

sinisaos:
> The example I gave was wrong, I corrected it. What I meant was a situation in which the column name (of the foreign key) has a different name than the table that it refers to
> 

You are right, it doesn't work if the column is different from the table name. I tried to use `foreign_key_columns_index` so that if the table has multiple fk columns we know the index of those columns, but it obviously doesn't work well.

> 
> ```python
> class MainManager(Table):
>     name = Varchar()
>     artists = ReverseLookup(LazyTableReference("Artist", module_path=__name__), "manager")
> 
> class Artist(Table):
>     name = Varchar()
>     manager = ForeignKey(MainManager)
> 
> # select:
> await MainManager.select(MainManager.artists(Artist.name))
> ```

This also doesn't work. Raise `ValueError: `_name` isn't defined - the Table Metaclass should set it.` on line 148 `name` property.

powellnorma:
> This also doesn't work. Raise ValueError: _name isn't defined - the Table Metaclass should set it. on line 148 name property.

Hm, for me that example works. I (already) made changes so that `_name` should (indeed) be set from the Table Metaclass [here](https://github.com/powellnorma/piccolo/blob/f98e0b360b4b8b54547bbe3d4e548320e31a1768/piccolo/table.py#L285-L287).
Have you made sure that you checked out my branch correctly? Maybe only parts of my changes got applied?

sinisaos:
Sorry, that was my mistake. I tried everything from scratch because I probably messed something up and you are right, the tests from my PR pass with your code. 
The only thing I've noticed is that if the column name is plural the tests fail, but singular name pass. I tried to strip column name if it ends with `s` and it works in `Sqlite` but not in `Postgres`.
I have to go now and will be gone for 2 days, but you should continue with this because your version is much better and easier to understand and it works. I'll be happy to delete my PR so you can make a new one. I'll try to help as much as I can.

powellnorma:
@sinisaos I [created](https://github.com/sinisaos/piccolo/pull/1) a PR on your fork/branch. If you accept it, my changes should be included in this PR.

> The only thing I've noticed is that if the column name is plural the tests fail

Do you have an example that demonstrates this problem? How can I reproduce this? Thank you

powellnorma:
> The only thing I've noticed is that if the column name is plural the tests fail

Does it still occur after this commit? https://github.com/sinisaos/piccolo/pull/1/commits/d6c64531246e20c00657e0dbd570268b548668ca

sinisaos:
@powellnorma I have added your changes. Thanks again for your help.

lqmanh:
Hi, just wonder what's the status quo of this PR, as it's been around for a long time. Are there any blockers left? I'd really love to see this feature getting merged.

Thanks you guys in advance.

dantownsend:
@lqmanh Will try and get this merged in soon. It just needs reviewing in detail to make sure we're happy with the API 👍

powellnorma:
One thing this is currently missing is the ability to use `.order_by` on the `ReverseLookupSelect`. Similarly with `.where` etc. @dantownsend Do you think this should be supported? And if so, do you see a way in which we could reuse the existing implementations for these methods?

erhuabushuo:
Wow, I'm waiting for this PR.

py-radicz:
it would be cool to have reverse lookup in the api, cant wait for it

dantownsend:
OK, seems like a lot of people want this. Getting Piccolo v1 out has taken most of my attention. Will take another look at this once I've released Python 3.12 support.

kayprogrammer:
You guys are doing well. 
Still waiting for this PR, including composite unique constraints and check constraints.

sam2x:
Hello, thank you for this lib, I plan to use it for an important sized project, and this feature is mandatory to me as it brings many benefits.
However, I am still confused, there is no update since about several years ? is it on the short time plan ? what is current issue/status ? I could not find any references in the master branch, if you could help updating us or provide an alternative ? Imho I can't imagine orm modern lib without a populated backref.

Best regards
> OK, seems like a lot of people want this. Getting Piccolo v1 out has taken most of my attention. Will take another look at this once I've released Python 3.12 support. Bear with me, I know this PR has been around for a while 🙏

sinisaos:
@sam2x I don't know if that makes sense to you, but you can try using the [unofficial branch](https://github.com/piccolo-orm/piccolo/pull/984#issuecomment-2564377055) (until everything is officially supported). The link contains an explanation and how to use it. Hope that helps.

sam2x:
@dantownsend Please it has been 3years now. Can you share your feedbacks on this or if uncomfortable the reasons this patch isn't yet applied in the master branch ? Thank you

sinisaos:
@snow-born You can already use `reverse lookup` in `M2M` (this PR is heavily inspired by M2M, but for O2M relationships) because Piccolo M2M is [basically wrapper around efficient SQL with joins](https://github.com/piccolo-orm/piccolo/blob/master/piccolo/columns/m2m.py#L60-L150) (I think that's great because we use databases and SQL which is always more efficient). You almost always have two tables and one joining table so you can easily go in both directions. 
[For tables as in the documentation](https://piccolo-orm.readthedocs.io/en/latest/piccolo/schema/m2m.html), you can do a `forward` query `await Band.select(Band.name, Band.genres())` to get all bands with genres or a `reverse` query `await Genre.select(Genre.name, Genre.bands())` to get all genres with their bands. 
The only problem might be with self-reference M2M (referencing the same table) which Piccolo doesn't currently support, but [here's a PR for that](https://github.com/piccolo-orm/piccolo/pull/1230) which allows forward and reverse queries with correct results. I hope that helps.

sinisaos:
I'm closing this PR just to clean up my open requests in the repo. Since it has been open for quite a while, I'll remove it from the list to keep things clean. Feel free to use it if needed. Thanks.

Issue #607, opened 2022-08-30
First class Cockroach DB support.


gnat:
### PR: https://github.com/piccolo-orm/piccolo/pull/608

We'd like to see Cockroach Labs promote Piccolo as one of, if not the recommended async ORM. `asyncpg` has [some support built-in](https://github.com/MagicStack/asyncpg/blob/bd192623d05668b5919d55d3b9a62629594e2fb7/asyncpg/connection.py#L2315) to help us get there.

In the mean time I will be adding notes on how to use CockroachDB with Piccolo, and general progress.

## Quick local setup for testing

* Download/extract (single binary): https://www.cockroachlabs.com/docs/stable/install-cockroachdb-linux.html
* Start Cockroach: `./cockroach start-single-node --insecure --store=node1 --listen-addr=localhost:26257 --http-addr=localhost:8080 --background`
* Enter SQL client: `./cockroach sql --insecure`
  * Can `create database ...;` `use ...;` from here.
* CTRL-C or `killall cockroach` to send graceful termination signal.

Default account is `root` without password. Fancy stats dashboard at `http://localhost:8080`

Both [Piccolo Examples](https://github.com/piccolo-orm/piccolo_examples) and [Piccolo Admin](https://github.com/piccolo-orm/piccolo_admin) are working.

![image](https://user-images.githubusercontent.com/24665/187473407-04064a60-3490-47cf-9b75-f8ab3bf056b5.png)

## What we still need

* [x] ~~Get build/test suite running on Github Actions for Piccolo QA tooling.~~
* [x] ~~Some way to identify we're running under Cockroach DB.~~
  * ~~Added CockroachDB Engine.~~
* [x] ~~Get test suite mostly / fully passing.~~
* [x] ~~Basic documentation~~ 
* [x] ~~Time Travel queries using AS OF SYSTEM TIME~~
  * ~~Some way to do Time Travel SELECT queries (`SELECT ... AS OF SYSTEM TIME`)~~
  * ~~https://www.cockroachlabs.com/docs/stable/as-of-system-time.html~~
  * :heavy_check_mark: See: https://github.com/piccolo-orm/piccolo/issues/607#issuecomment-1251330590
* [x] Sharding for Sequential Indexes
  * :warning: Done, but will go into its own PR after this one: https://github.com/piccolo-orm/piccolo/issues/622
  * Some way to do "Hash Sharded" Indexes (`INDEX ... USING HASH`)
[...]

PR #608, opened 2022-08-31
CockroachDB Support

gnat:
Work in progress for https://github.com/piccolo-orm/piccolo/issues/607


piccolo/apps/schema/commands/generate.py
```diff
@@ -307,6 +308,12 @@ def __add__(self, value: OutputSchema) -> OutputSchema:
     "uuid": UUID,
 }
 
+# Re-map for Cockroach compatibility.
+if engine.engine_type == 'cockroach':
+    COLUMN_TYPE_MAP["integer"] = BigInt
+    COLUMN_TYPE_MAP["json"] = JSONB
+    COLUMN_TYPE_MAP["smallint"] = BigInt
```
dantownsend, line 315: Should `smallint` be mapped to `BigInt`?

https://www.cockroachlabs.com/docs/stable/int.html#names-and-aliases

I'm not sure about modifying the dictionary - even though it works, maybe we can figure out a different approach.

Something like:

```python
COLUMN_TYPE_MAP = {
    'postgres': {...},
    'cockroach': {...}
}
```

Or:

```python
COCKROACH_COLUMN_TYPE_MAP = dict(
    COLUMN_TYPE_MAP,
    integer=BigInt,
    json=JSONB,
    smallint=BigInt
)
```

We would then have to pass around the engine_type to each function.

piccolo/columns/base.py
```diff
@@ -594,7 +594,7 @@ def ilike(self, value: str) -> Where:
         For SQLite, it's just proxied to a LIKE query instead.
 
         """
-        if self._meta.engine_type == "postgres":
+        if self._meta.engine_type == "postgres" or self._meta.engine_type == "cockroach":
```
dantownsend, line 597: We can do this instead - just slightly more compact:

```python
if self._meta.engine_type in ('postgres', 'cockroach'):
```

piccolo/columns/column_types.py
```diff
@@ -201,6 +207,22 @@ def get_postgres_interval_string(self, interval: timedelta) -> str:
...
+
+        """
+        output = []
+        for timedelta_key, cockroach_name in self.cockroach_attr_map.items():
+            timestamp_value = getattr(interval, timedelta_key)
+            if timestamp_value:
+                output.append(f"{timestamp_value} {cockroach_name}")
+        output_string = " ".join(output)
+        return f"'{output_string}'"
+
```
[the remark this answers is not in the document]
dantownsend: I think we can get rid of `cockroach_attr_map` too.

piccolo/columns/defaults/date.py
```diff
@@ -35,6 +35,11 @@ def postgres(self):
         interval_string = self.get_postgres_interval_string(["days"])
         return f"CURRENT_DATE + INTERVAL '{interval_string}'"
 
+    @property
+    def cockroach(self):
+        interval_string = self.get_postgres_interval_string(["days"])
+        return f"CURRENT_DATE + INTERVAL '{interval_string}'"
```
dantownsend, line 41: Might be better off just doing:

```python
return self.postgres
```

piccolo/columns/defaults/interval.py
```diff
@@ -53,6 +53,21 @@ def postgres(self):
...
+                "weeks",
+                "days",
+                "hours",
+                "minutes",
+                "seconds",
+                "milliseconds",
+                "microseconds",
+            ]
+        )
+        return f"'{value}'"
```
dantownsend, line 69: Maybe return `self.postgres` here too.

dantownsend (commented):
Will finish reviewing later - just some initial comments.

Thanks!

piccolo/columns/base.py
gnat: Figured the other would be faster but I like the readability of `if .. in ()` :+1:

piccolo/apps/schema/commands/generate.py
gnat: Great ideas!

gnat: Added. Gonna see if I can get SmallInt working.

gnat:
Thank you for the refinements and guidance so far @dantownsend !

piccolo/table.py
```diff
@@ -379,7 +379,10 @@ def __init__(
 
     @classmethod
     def _create_serial_primary_key(cls) -> Serial:
-        pk = Serial(index=False, primary_key=True, db_column_name="id")
+        if cls._meta.db.engine_type == "cockroach":
+            pk = Serial(index=False, primary_key=True, db_column_name="id", default="unique_rowid()") # Not sure where else to put "unique_rowid()". Has to be at table creation time.
```
dantownsend, line 383: We need to refactor this slightly.

It's problematic when installing piccolo in a fresh project - it tries to find `piccolo_conf` and fails.

Maybe we can change this instead:

https://github.com/piccolo-orm/piccolo/blob/7754498705540c81c40ec203a7bb429d59a07da5/piccolo/columns/column_types.py#L777-L783

If we add this, hopefully it still works:

```python
        elif engine_type == "cockroachdb":
            return 'unique_rowid()'
```

gnat: @dantownsend thank you for the invaluable feedback!

So while this works, I am concerned about the repercussions of using `unique_rowid()` instead of `DEFAULT` in piccolo operations after table creation time.

Cockroach should not care either way, though. I'm going to say while this is a low concern, should be fine.

gnat: I've gone ahead and made the changes, updated the cockroach tests. All green as expected.

dantownsend:
Thanks for adding the docs - looks great.

dantownsend:
@gnat  Sorry for the delay on getting this merged in. I haven't finished reviewing it yet.

@sinisaos If you get a chance, can you also take a look?

gnat:
@dantownsend Yeah, I've been dogfooding it locally in my own projects, so it has been getting some real usage. Thanks for keeping on it; it'll be great to wrap up a beta release.

sinisaos:
@dantownsend I tried this PR branch and everything looks and works great.
@gnat Thanks again for your great effort for managing this big PR.

I just have one question. Do we have to wait for the beta release of `V22.2.0-alpha-x` , because if it's lower than that the `format()` function is not supported and we get errors for `Readable` column tests, or do we have to put a warning to users that they have to use `V22.2.0-alpha-x`?

gnat:
@sinisaos Thanks for the review and recognition!

IMO, I don't think it's worth removing `format()` from Piccolo because Cockroach Labs specifically added `format()` for us in v22.2. Which was nice! :smiley: 

This is a non-issue in the near future- most devs using Cockroach on Piccolo will be on v22.2 when the stable version is out in the next month or two.

Normally, CockroachDB is designed to be a "rolling database" aka upgrading it is as simple as switching out to the next version of a binary (or changing a number in a `docker-compose.yml`). CockroachDB will auto-migrate to the new version. Migrations are live and do not have downtime.

This situation is special because you normally cannot upgrade an "alpha" release, so I recommend:

A. **If you're already shipping to production:** Use the mostly working v22.1 and plan to auto-migrate to v22.2 when the stable version is out. This will work fine!
B. **If you're in early development:** and are okay rebuilding your database once before v22.2 stable, you can use v22.2 alpha, including the `format()` features.

gnat:
Documented: https://github.com/piccolo-orm/piccolo/pull/608/commits/9f0c99bb753e0ba72d9711dfb2aae981215eb24f

sinisaos:
> IMO, I don't think it's worth removing `format()` from Piccolo (or skipping tests or writing new tests) because Cockroach Labs specifically added `format()` for us in v22.2. Which was nice!

I didn't think to remove `format()` because `format()` is important for `Readable()` column which is often used in Piccolo Admin for nicer display of foreign key. I thought we should put some warning to users, but that doesn't matter if you say that `v22.2` is coming out in a month or two. Thanks again for the great work.

docs/src/piccolo/getting_started/setup_cockroach.rst
```diff
@@ -0,0 +1,53 @@
+.. _setting_up_cockroach:
+
+##############
+Setup Cockroach
+##############
```
dantownsend, line 5: RST is incredibly fussy - the number of hashes needs to match the title length. It just needs one more top and bottom.

docs/src/piccolo/getting_started/setup_cockroach.rst
```diff
@@ -0,0 +1,53 @@
...
+
+Follow the  `instructions for your OS <https://www.cockroachlabs.com/docs/stable/install-cockroachdb.html>`_.
+
+
+Versions
+--------
+
+We support the latest stable version.
+
+Note: Features using ``format()`` will be available in v22.2 or higher, but we recommend using the stable version so you can upgrade automatically when it becomes generally available.
```
dantownsend, line 18: We could put this inside an RST note box.

```rst
.. note::

   Features using ``format()`` will be available in v22.2 or higher, but
   we recommend using the stable version so you can upgrade automatically when
   it becomes generally available.
```

piccolo/apps/migrations/auto/migration_manager.py
```diff
@@ -788,4 +788,11 @@ async def run(self, backwards=False):
             await self._run_drop_columns(backwards=backwards)
             await self._run_drop_tables(backwards=backwards)
             await self._run_rename_columns(backwards=backwards)
+            # We can remove this for cockroach when resolved.
+            # https://github.com/cockroachdb/cockroach/issues/49351
+            # "ALTER COLUMN TYPE is not supported inside a transaction"
+            if engine.engine_type != "cockroach":
+                await self._run_alter_columns(backwards=backwards)
```
dantownsend, line 795: Should we create a separate issue to track this, as a reminder to change it in the future?

tests/example_apps/mega/tables.py
```diff
@@ -22,37 +22,73 @@
...
+        timestamptz_col = Timestamptz()
+        uuid_col = UUID()
+        varchar_col = Varchar()
 
-    unique_col = Varchar(unique=True)
-    null_col = Varchar(null=True)
-    not_null_col = Varchar(null=False)
+        unique_col = Varchar(unique=True)
+        null_col = Varchar(null=True)
+        not_null_col = Varchar(null=False)
```
dantownsend, line 94: Are these the same, except for the `JSON` column? As `JSON.column_type` returns `JSONB` for Cockroach, should it still work without needing the modified version of `MegaTable`?

tests/example_apps/music/tables.py
```diff
@@ -24,14 +28,31 @@ def get_readable(cls) -> Readable:
...
+        Special version for Cockroach.
+        """
+
+        name = Varchar(length=50)
+        manager = ForeignKey(Manager, null=True)
+        popularity = BigInt(default=0)
+
+        @classmethod
+        def get_readable(cls) -> Readable:
+            return Readable(template="%s", columns=[cls.name])
```
dantownsend, line 55: What happens if someone uses `Integer` with Cockroach? Should we make `Integer` just maps to `BigInt` automatically, by adding `column_type` to `BigInt`?

.github/workflows/tests.yaml
```diff
@@ -132,6 +132,40 @@ jobs:
...
+              run: |
+                  python -m pip install --upgrade pip
+                  pip install -r requirements/requirements.txt
+                  pip install -r requirements/test-requirements.txt
+                  pip install -r requirements/extras/postgres.txt
+            - name: Setup CockroachDB
+              run: |
+                  wget -qO- https://binaries.cockroachdb.com/cockroach-${{ matrix.cockroachdb-version }}.linux-amd64.tgz | tar xz
+                  ./cockroach-${{ matrix.cockroachdb-version }}.linux-amd64/cockroach start-single-node --insecure --background
+                  ./cockroach-${{ matrix.cockroachdb-version }}.linux-amd64/cockroach sql --insecure -e 'create database piccolo;'
```
dantownsend, line 158: When running Cockroach locally with `cockroach start-single-node --insecure` I get quite slow performance. Should we pass in any extra flags here to give it more resources?

piccolo/apps/migrations/auto/migration_manager.py
gnat: Tracked: https://github.com/piccolo-orm/piccolo/issues/632

tests/example_apps/mega/tables.py
gnat: Yup, we need the special table here.

The test expects a `JSON` to be returned for `test_generate.py`. We can never know if what we sent to Cockroach was `JSON` or `JSONB`. Cockroach will always convert to `JSONB`.

The developer can use `JSON` or `JSONB`, it does not matter. This test is for integrity behind the scenes.

.github/workflows/tests.yaml
gnat: `./cockroach start-single-node --insecure --background` should give Cockroach uncapped resources.

The poor performance is likely just due to warmup / node generation. Although you could try adding `--store=node_1` to ensure it's writing to the local disk beside the binary. The tests on Github are already pretty fast though, I wouldn't worry about it (only a few minutes to run).

You can point Cockroach `--store=` to anything writable and it will use it as the database.

For best performance, I just keep a binary of Cockroach around that I can start at any time (it will re-use the on-disk state when available, in my case `./node_1`... which just looks like a folder beside the binary where Cockroach is ran).

Full reference here: https://www.cockroachlabs.com/docs/dev/cockroach-start-single-node.html

tests/example_apps/music/tables.py
gnat: Same story explained above for `JSON` vs `JSONB`

The developer can use `Integer` or `BigInt`, it does not matter, but Cockroach will always return BigInt, so the test here fails, and we need the special table.

tests/example_apps/mega/tables.py
gnat: Added docs: https://github.com/piccolo-orm/piccolo/pull/608/commits/ebe06e7e0286e4c2484bbd626740435eb50297f9

tests/example_apps/music/tables.py
gnat: Added docs: https://github.com/piccolo-orm/piccolo/pull/608/commits/ebe06e7e0286e4c2484bbd626740435eb50297f9

.github/workflows/tests.yaml
dantownsend: Cool, thanks for the explanation.

Issue #356, opened 2021-11-22
Pylance complains about import in auto generated migrations.


gmos:
![image](https://user-images.githubusercontent.com/1526544/142867514-9294c787-8fb7-4e23-a4cb-f0be7b52b0af.png)

Dan,

As the screenshot shows Pylance seems to ignore the relative import of the MigrationManager done in `piccolo/apps/migrations/auto/__init__.py` and wants to see:
`from piccolo.apps.migrations.auto.migration_manager import MigrationManager`
 
I'll also check Pylance's issue list.  Think it should use exactly the same import sematics as Python itself.
Nevertheless, this would be easy to fix by adjusting the generated import in Piccolo's auto migration code generator.

dantownsend:
@gmos That's an odd one. I'll change the import in the next release, thanks.

dantownsend:
@gmos Thanks for looking into it so thoroughly. I've done a PR for option 2 - using a more explicit import path.

Importing it as `__MigrationManager` is a clever solution, which would fix the issue for existing migration files. Hopefully there's an official solution in the future in Python or Pyright.

PR #629, opened 2022-10-03
Add named exports for column module

### LINKED ISSUE #628: Classes from the columns submodule are not named exports, meaning they aren't suggested by VS Code intellisense
Essentially, VS Code assumes that imported objects are private unless they are declared as a named export using the `__all__` variable.

Because there is no definition of an `__all__` variable in the `piccolo.columns` module, VS Code does not provide intellisense predictions for the imports, and Pylance gives the error `"Varchar" is not exported from module "piccolo.columns" Import from "piccolo.columns.column_types" instead`.

MiguelGuthridge:
Resolves #628


dantownsend:
@MiguelGuthridge Thanks for this!

It's a problem we've encountered before: https://github.com/piccolo-orm/piccolo/issues/356

Now we know that specifying `__all__` tells Pylance what we consider public vs private.

Issue #635, opened 2022-10-08
Add a `drop_table` method to the `Table` class


MiguelGuthridge:
Although (as per #283), it is possible to drop tables using the `drop_db_tables`, it can often be more clear to modify the table using a method attached directly to that table. This makes sense given that the `Table` class already has a `create_table` method.

dantownsend:
There's a `drop_table`, but it's not at the top level. For example:

```python
await Band.alter().drop_table()
```

The reason it's like this was to make it less likely to be accidentally triggered. But the downside is it's less obvious that it exists!

PR #636, opened 2022-10-08
WIP: Add drop table method to Table class - feedback wanted

MiguelGuthridge:
An attempt to address #635.

Since I don't know my way around the codebase very well currently, I may have done something horribly wrong. Would massively appreciate an explainer if I did anything wrong so that I can know how to avoid the issues in future PRs.

### DIFF
--- a/piccolo/table.py
+++ b/piccolo/table.py
@@ -982,6 +982,17 @@ class Table(metaclass=TableMetaclass):
         )
 
     @classmethod
+    def drop_table(cls):
+        """
+        Drop table, along with all columns.
+
+        .. code-block:: python
+
+            await Band.drop_table()
+        """
+        return drop_db_tables(cls)
+
+    @classmethod
     def alter(cls) -> Alter:
         """
         Used to modify existing tables and columns.


MiguelGuthridge:
Oop I guess that means I need to do add some tests for it

piccolo/table.py
```diff
@@ -981,6 +981,17 @@ def create_table(
...
+    @classmethod
+    def drop_table(cls):
+        """
+        Drop table, along with all columns.
+
+        .. code-block:: python
+
+            await Band.drop_table()
+        """
+        return drop_db_tables(cls)
```
dantownsend, line 993: Instead of this, it could be a proxy to `alter().drop_column()` - that way it can be used sync or async.

```python
    @classmethod
    def drop_table(cls):
        """
        Drop table, along with all columns.
        .. code-block:: python
            await Band.drop_table()
        """
        return cls.alter().drop_table()
```

So user's can then do either of these:

```python
# async
await MyTable.drop_table()

# sync
MyTable.drop_table().run_sync()
```

The main question though is whether to add this. It's a trade-off. The pro is it's more convenient. The con, is it's more likely someone will accidentally call it by accident. Hmm, tricky 🤔

MiguelGuthridge: Since the main issue was that I thought the feature was missing, I think things could be improved by making improvements to the documentation. Perhaps having a quick start page that has a list of common how tos. Something like this would work:

### I Want To:

* Define a new table: [code with explanation]
* Create a table: [code with explanation]
* Drop a table: [code with explanation]
* Perform a query: [code with explanation]
* Perform a join on a query: [code with explanation]

It's not that the documentation isn't there, it's just that there didn't seem to be a quick and easy cheat sheet which I think would really benefit new users (such as me).

dantownsend:
Cool, thanks.

PR #637, opened 2022-10-08
Style improvements for table module

MiguelGuthridge:
* Fixed a couple of typos
* Fixed a type conflict by adding a type cast (I'm assuming that the fact that these types conflicted is not an indication of a bug)
* Added instructions for which functions to use instead for deprecated `create_tables` and `drop_tables` functions

Pretty small improvements, but hopefully improvements nonetheless :)


dantownsend:
@MiguelGuthridge Cool, thanks for this! The small improvements are just as important.

PR #638, opened 2022-10-11
Documentation: New as_of() clause (Advanced, added with CockroachDB).

gnat:


dantownsend:
This looks good - maybe we can make it so `as_of` can accept a `timedelta` too:

```python
import datetime

delta = datetime.timedelta(minutes=5)

await Band.select().as_of(delta)
```

gnat:
I agree that would be nice! When I find some development time for more features/tests it would be something to do, but I am making sure the existing features are documented.

dantownsend:
@gnat OK, cool.

Issue #622, opened 2022-09-20
Cockroach DB: Sharding for Sequential Indexes


gnat:
Code and tests for this are ready to go, but waiting on https://github.com/piccolo-orm/piccolo/issues/607

https://www.cockroachlabs.com/docs/stable/hash-sharded-indexes.html

PR #639, opened 2022-10-11
CockroachDB: Hash Sharded Indexes. (#622)

gnat:
Issue: https://github.com/piccolo-orm/piccolo/issues/622

Implements "Sharding for Sequential Indexes" as noted in: https://github.com/piccolo-orm/piccolo/issues/607

Major performance feature for "sequential data" for clusters (timestamps, etc).
* Article: https://www.cockroachlabs.com/blog/hash-sharded-indexes-unlock-linear-scaling-for-sequential-workloads/
* Video: https://www.youtube.com/watch?v=JlT_Co_wYFo
* Reference manual: https://www.cockroachlabs.com/docs/stable/hash-sharded-indexes.html

### DIFF
--- a/piccolo/columns/base.py
+++ b/piccolo/columns/base.py
@@ -154,6 +154,7 @@ class ColumnMeta:
     help_text: t.Optional[str] = None
     choices: t.Optional[t.Type[Enum]] = None
     secret: bool = False
+    sharded: bool = False
 
     # Used for representing the table in migrations and the playground.
     params: t.Dict[str, t.Any] = field(default_factory=dict)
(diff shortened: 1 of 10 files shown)


gnat:
@dantownsend This is the last major addition to complete the Cockroach integration. I separated it out of the main PR to reduce complexity.

It's optional, but very highly recommended.

I wanted to get feedback on the approach here, if there was a more elegant way to do this.

I have to step away for today, but I will get all the linters and such passing shortly.

piccolo/columns/base.py
```diff
@@ -437,6 +438,12 @@ class Band(Table):
             >>> await Band.select(exclude_secrets=True)
             [{'name': 'Pythonistas'}]
 
+    :param sharded:
+        If ``True`` and primary_key or index is also set ``True``, this index
+        will automatically use sharding across a cluster. Highly recommended
+        for sequence columns, such as: Serial, Timestamp.
+        Also known as Hash Sharded Index.
```
dantownsend, line 445: I would add something here saying 'CockroachDB only'.

If we're likely to have several database specific options, maybe we should namespace them. For example, `cockroach_kwargs`, or something.

dantownsend:
> I have to step away for today, but I will get all the linters and such passing shortly.

@gnat Cool, no worries!

PR #649, opened 2022-10-21
Fix create_pydantic_model for nullable JSON and JSONB

eneacosta:
Even if the value of a `JSON` or `JSONB` column is Optional, the [pydantic.validator for json values will be added to the column](https://github.com/piccolo-orm/piccolo/blob/master/piccolo/utils/pydantic.py#L226), and this will raise an exception if a None value is retrieved, an exception like this: 

```
pydantic.error_wrappers.ValidationError: 1 validation error for Table
column
  the JSON object must be str, bytes or bytearray, not NoneType (type=type_error)
```

This brokes piccolo_admin panel when you add a table that has a nullable `JSON` or `JSONB` column and a null value is saved to it.

I prevented this bug by returning the None value if the column is not required.


piccolo/utils/pydantic.py
```diff
@@ -37,8 +37,10 @@ class Config(pydantic.BaseConfig):
     arbitrary_types_allowed = True
 
 
-def pydantic_json_validator(cls, value):
+def pydantic_json_validator(cls, value, field):
     try:
+        if not field.required and value is None:
+            return value
```
dantownsend, line 43: This can probably go outside of the `try` block.

dantownsend:
@eneacosta Thanks for this - good catch.

I just left one minor comment. We should probably add a simple test here too:

https://github.com/piccolo-orm/piccolo/blob/3fb22b01d492d9be2cd12d68a2d510d9964109cd/tests/utils/test_pydantic.py#L212

Just to make sure that `None` can be passed to the Pydantic model.

eneacosta:
@dantownsend I movel the validation outside of the `try` block and added tests for `None` values

dantownsend:
@eneacosta That's great - thanks again! 👍

PR #656, opened 2022-10-27
Playground: Option to use ipython profile

haffi96:
I started the piccolo playground locally and noticed that there was no syntax highlighting. Did I do something wrong?

Unless I misunderstood. It seems that when you start the playground, it creates an embedded ipython shell with the default config. I think this means IPython will use the user's own config. But if there is None, it will still use the default config because the playground embed instance doesn't specify a config. This mean no syntax highlighting

These changes will make it so that by default, running playground will start the embedded shell with syntax highlighting. This also overrides the user config (which is at ~/.ipython/profile_.....).

So to give user the option to use their own, i've added a param to allow that

### DIFF
--- a/piccolo/apps/playground/commands/run.py
+++ b/piccolo/apps/playground/commands/run.py
@@ -264,6 +264,6 @@ def run(
         print("\n")
         conf_args = {}
     else:
-        conf_args = {"colors": "Neutral", "mouse_support": True}
+        conf_args = {"colors": "neutral"}
 
     IPython.embed(using=_asyncio_runner, **conf_args)
--- a/piccolo/apps/playground/commands/run.py
+++ b/piccolo/apps/playground/commands/run.py
@@ -261,6 +261,9 @@ def run(
             "More info: https://ipython.org/ipython-doc/3/config/intro.html#"
         )
         print("\n")
+        # To try this out, set `c.TerminalInteractiveShell.colors = "Linux"`
+        # in `~/.ipython/profile_default/ipython_config.py` to set the terminal
+        # color.
         conf_args = {}
     else:
         conf_args = {"colors": "neutral"}
--- a/piccolo/apps/playground/commands/run.py
+++ b/piccolo/apps/playground/commands/run.py
@@ -24,6 +24,7 @@ from piccolo.columns.readable import Readable
 from piccolo.engine import PostgresEngine, SQLiteEngine
 from piccolo.engine.base import Engine
 from piccolo.table import Table
+from piccolo.utils.warnings import colored_string
 
 
 class Manager(Table):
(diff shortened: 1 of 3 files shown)


dantownsend:
@haffi96 Thanks for this.

You're right - we should be using syntax highlighting. Ironically, we added it to another part of the app, but not to the playground for some reason:

https://github.com/piccolo-orm/piccolo/blob/6c1571cfee945330135974eb1197892649f3c4d2/piccolo/apps/shell/commands/run.py#L29

On my machine, when I passed in ``--ipython_profile``, which sets `config=None`, then iPython wasn't picking up my config from `~/.ipython/profile_default/ipython_config.py`. Does this work for you?

piccolo/apps/playground/commands/run.py
```diff
@@ -252,5 +255,15 @@ def run(
...
 
-    IPython.embed(using=_asyncio_runner)
+    if not ipython_profile:
+        conf = get_config()
+        conf.InteractiveShellEmbed.colors = "Neutral"
+    else:
+        print("Using user ipython profile")
+        print("For details, see: https://ipython.org/ipython-doc/3/config/intro.html#ipythondir")
+        print("\n")
+        conf = None
```
dantownsend, line 267: I would generally flip these, so it's like:

```python
if ipython_profile:
    ...
else:
    ...
```

So the first `if` statement is truthy.

haffi96:
> then iPython wasn't picking up my config from ~/.ipython/profile_default/ipython_config.py. Does this work for you?

Yes it does, however my ipython_config.py is just the default with the `c.InteractiveShell.colors = "Neutral"` setting turned on 🤔 I'll look into it and double check

dantownsend:
> Yes it does, however my ipython_config.py is just the default with the c.InteractiveShell.colors = "Neutral" setting turned on 🤔 I'll look into it and double check

Cool, thanks!

haffi96:
The code snippet you linked from https://github.com/piccolo-orm/piccolo/blob/6c1571cfee945330135974eb1197892649f3c4d2/piccolo/apps/shell/commands/run.py#L29 showed me that there is a better way to do this.

We don't have to pass in a whole new Config class. We can just specify the config params we want to modify. This combines the default config but just overrides the params you specify. (in Most recent commit)

Also, i checked and not specifying any config args in `IPython.embed(using=_asyncio_runner)` call does use my local ipython config from `~/.ipython/profile_default/ipython_config.py`. But to override the default syntax highlight setting, I had to turn on this setting in ipython_config.py -> `c.TerminalInteractiveShell.colors = 'Neutral'`. I referred to the wrong one in an earlier comment

dantownsend:
You're right - I tried `c.TerminalInteractiveShell.colors = "Linux"` and it worked.

piccolo/apps/playground/commands/run.py
```diff
@@ -253,4 +256,14 @@ def run(
 
     from IPython.core.interactiveshell import _asyncio_runner  # type: ignore
 
-    IPython.embed(using=_asyncio_runner)
+    if ipython_profile:
+        print("Using user ipython profile")
+        print(
+            "More info: https://ipython.org/ipython-doc/3/config/intro.html#"
```
dantownsend, line 262: I'd change this link to this:

https://ipython.readthedocs.io/en/stable/config/intro.html

As it will point to the latest version of the docs.

piccolo/apps/playground/commands/run.py
```diff
@@ -253,4 +256,14 @@ def run(
...
+        )
+        print("\n")
+        conf_args = {}
+    else:
+        conf_args = {"colors": "Neutral", "mouse_support": True}
```
dantownsend, line 267: We might have to leave out `mouse_support`. When I try it on my Mac, it means I can no longer scroll.

dantownsend: If I press the the option key, then I can scroll when using a trackpad.

Hmm, I think we should probably leave this out, as it could catch people out.

I presume someone can enable `mouse_support` in their iPython config if they want it.

dantownsend:
@haffi96 I pushed a couple of tweaks. I took this opportunity to add a couple of small enhancements to the playground.

If you're happy with this, I'll merge it and release it.

piccolo/apps/playground/commands/run.py
haffi96: Yep, makes sense. I was sceptical of adding it myself

haffi96:
> @haffi96 I pushed a couple of tweaks. I took this opportunity to add a couple of other small enhancements to the playground.
> 
> If you're happy with this, I'll merge it and release it.

They changes look great. Thankyou!

dantownsend:
@haffi96 OK, cool - will merge it in. Thanks for this.

PR #699, opened 2022-12-01
Fix typo for `raw` queries in docs

StitiFatah:


dantownsend:
@StitiFatah Good catch - thanks!

PR #701, opened 2022-12-02
Add returning clause for delete operation

### LINKED ISSUE #698: Getting number of affected rows
With some operations, postgres returns the number of affected/processed rows. Other ORMs make that information available. My use case is for deletions, but perhaps it would be nice to have this in general.
I haven't looked at the codebase, but how hard would that be to implement? I might give that a try if it's not too complicated.

waldner:
Should fix #698 .

### DIFF
--- a/docs/src/piccolo/query_clauses/returning.rst
+++ b/docs/src/piccolo/query_clauses/returning.rst
@@ -7,6 +7,7 @@ You can use the ``returning`` clause with the following queries:
 
 * :ref:`Insert`
 * :ref:`Update`
+* :ref:`Delete`
 
 By default, an update query returns an empty list, but using the ``returning``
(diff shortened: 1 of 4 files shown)


dantownsend:
@waldner This looks great, thanks!

I just a minor tweak to skip the test for older versions of SQLite. Once the CI has finished will merge it in.

waldner:
> I just made a minor tweak to skip the test for older versions of SQLite.

Argh, I missed that bit! Thanks!

waldner:
Any chances this will make it to a release anytime soon?

dantownsend:
Yeah, will release it now.

dantownsend:
It's available in version `0.99.0`:

https://pypi.org/project/piccolo/0.99.0/

Issue #702, opened 2022-12-02
Improve support for array choices


dantownsend:
Related to https://github.com/piccolo-orm/piccolo_admin/issues/161

Piccolo Admin doesn't support choices for array fields currently. This is often requested, and would be very useful. For example:

```python
class MusicBlogPost(Table):
    class Tags(str, Enum):
        album = 'album'
        concert = 'concert'
        news = 'news'
        review = 'review'

    tags = Array(Varchar(), choices=Tags)
```

We need to double check that Piccolo itself handles choices correctly (like in the above example), and write a bunch of test cases.

PR #704, opened 2022-12-02
add support for choices with `Array` columns

dantownsend:
Fixes https://github.com/piccolo-orm/piccolo/issues/702

There were some issues before when specifying choices for `Array` columns, and it wasn't clear which of these to use:

```python
Array(Varchar(choices=SomeEnum))
Array(Varchar(), choices=SomeEnum))
```

This PR adds a bunch of tests, and updates the docs to make it clear that this is what you should do:

```python
Array(Varchar(), choices=SomeEnum))
```

### DIFF
--- a/docs/src/piccolo/schema/advanced.rst
+++ b/docs/src/piccolo/schema/advanced.rst
@@ -140,6 +140,42 @@ By using choices, you get the following benefits:
 * Improved storage efficiency (we can store ``'l'`` instead of ``'large'``).
 * Piccolo Admin support
 
+``Array`` columns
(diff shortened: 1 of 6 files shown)


sinisaos:
@dantownsend Just a reminder to check out Piccolo Admin as the new arrays with choices don't work in Admin. Other than that, great work.

dantownsend:
@sinisaos Thanks!

Here's the PR for Piccolo Admin if you want to give it a go:

https://github.com/piccolo-orm/piccolo_admin/pull/260

I think it's a useful feature - people have asked for it a few times, and I'll definitely use it.

PR #705, opened 2022-12-02
Expose pydantic config class to user

### LINKED ISSUE #467: Choices enum not reflected in Swagger
I'm using Piccolo with FastAPI, which generates documentation in Swagger. In my tables file, I've created a table (relationship) that has a field based on an enum

```python
    class RelationshipType(str, enum.Enum):
        colleague = "colleague"
        friend = "friend"
        ...

    type= Varchar(choices=RelationshipType)
```

When I use create_pydantic_model to generate a pydantic model based on the table, I had expected the "choices" field to get passed along to Swagger, but that doesn't seem to be the case. Swagger treats it as a vanilla string.

I dumped the schemas for both a Pydantic model created directly that uses an enum and a Pydantic model created using the create_pydantic_model method, and they're laid out pretty differently. I'll add them at the end.

Do you have any guidance on how to get the enum created in piccolo tables to get reflected in Swagger? Do I have to drop down and create Pydantic models directly if I'd like this functionality?

```
# Direct Pydantic model 
model.construct().schema()
{'title': 'RelationshipTest', 'type': 'object', 'properties': {'class_1': {'title': 'Class 1', 'type': 'integer'}, 'class2': {'title': 'Class2', 'type': 'integer'}, 'type': {'$ref': '#/definitions/RelationshipType'}}, 'required': ['class_1', 'class2', 'type'], 'definitions': {'RelationshipType': {'title': 'RelationshipType', 'description': 'An enumeration.', 'enum': ['colleague', 'friend', 'relative'], 'type': 'string'}}}
```

```
# Creating a model from Piccolo table
model.construct().schema()
{'title': 'RelationshipIn', 'type': 'object', 'properties': {'first_contact': {'title': 'First Contact', 'extra': {'foreign_key': True, 'to': 'contact', 'target_column': 'id', 'help_text': None, 'choices': None}, 'nullable': True, 'type': 'integer'}, 'second_contact': {'title': 'Second Contact', 'extra': {'foreign_key': True, 'to': 'contact', 'target_column': 'id', 'help_text': None, 'choices': None}, 'nullable': True, 'type': 'integer'}, 'type': {'title': 'Type', 'extra': {'help_text': None, 'choices': {'colleague': {'display_name': 'Colleague', 'value': 'colleague'}, 'friend': {'display_name': 'Friend', 'value': 'friend'}, 'relative': {'display_name': 'Relative', 'value': 'relative'}, 'partner': {'display_name': 'Partner', 'value': 'partner'}, 'ex': {'display_name': 'Ex', 'value': 'ex'}, 'child': {'display_name': 'Child', 'value': 'child'}, 'teacher': {'display_name': 'Teacher', 'value': 'teacher'}, 'student': {'display_name': 'Student', 'value': 'student'}}}, 'nullable': False, 'maxLength': 255, 'type': 'string'}}, 'help_text': None}
```

waldner:
`create_pydantic_model` accepts a new argument specifying the base class
to use for the generated model config.

Should partially fix #467.

### DIFF
--- a/piccolo/utils/pydantic.py
+++ b/piccolo/utils/pydantic.py
@@ -90,6 +90,7 @@ def create_pydantic_model(
     deserialize_json: bool = False,
     recursion_depth: int = 0,
     max_recursion_depth: int = 5,
+    pydantic_config_class: t.Type[pydantic.BaseConfig] = pydantic.BaseConfig,
     **schema_extra_kwargs,
 ) -> t.Type[pydantic.BaseModel]:
     """
@@ -132,6 +133,10 @@ def create_pydantic_model(
         Not to be set by the user - used internally to track recursion.
     :param max_recursion_depth:
         If using nested models, this specifies the max amount of recursion.
+    :param pydantic_config_class:
+        Config class to use as base for the generated pydantic model
+        (default: ``pydantic.BaseConfig``). You can create your own
+        subclass of ``pydantic.BaseConfig`` and pass it here.
     :param schema_extra_kwargs:
         This can be used to add additional fields to the schema. This is
         very useful when using Pydantic's JSON Schema features. For example:
@@ -301,7 +306,7 @@ def create_pydantic_model(
 
         columns[column_name] = (_type, field)
 
-    class CustomConfig(Config):
+    class CustomConfig(Config, pydantic_config_class):  # type: ignore
         schema_extra = {
             "help_text": table._meta.help_text,
             **schema_extra_kwargs,
--- a/docs/src/piccolo/serialization/index.rst
+++ b/docs/src/piccolo/serialization/index.rst
@@ -154,6 +154,29 @@ By default the primary key column isn't included - you can add it using:
 
     BandModel = create_pydantic_model(Band, include_default_columns=True)
 
+``pydantic_config_class``
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+You can specify a custom class to use as base for the pydantic model's config.
+This class should be a subclass of ``pydantic.BaseConfig``.
+For example, let's set the ``extra`` parameter to tell pydantic how to treat
+extra fields (that is, fields that would not otherwise be in the generated model).
+The allowed values are::
+
+* ``'ignore'`` (default): silently ignore extra fields
+* ``'allow'``: accept the extra fields and assigns them to the model
+* ``'forbid'``: fail validation if extra fields are present
+
+So if we want to disallow extra fields, we can do:
+
+.. code-block:: python
+
+    class MyPydanticConfig(pydantic.BaseConfig):
+        extra = 'forbid'
+
+    model = create_pydantic_model(table=MyTable, pydantic_config_class=MyPydanticConfig, ...)
+
+
 Required fields
 ~~~~~~~~~~~~~~~
 
--- a/tests/utils/test_pydantic.py
+++ b/tests/utils/test_pydantic.py
@@ -2,6 +2,7 @@ import decimal
 from unittest import TestCase
 
 import pydantic
+import pytest
 from pydantic import ValidationError
 
 from piccolo.columns import (
@@ -701,3 +702,37 @@ class TestSchemaExtraKwargs(TestCase):
 
         model = create_pydantic_model(Band, visible_columns=("name",))
         self.assertEqual(model.schema()["visible_columns"], ("name",))
+
+
+class TestPydanticExtraFields(TestCase):
+    def test_pydantic_extra_fields(self):
+        """
+        Make sure that the value of ``pydantic_extra_fields`` is correctly
+        propagated to the generated model.
+        """
+
+        class Band(Table):
+            name = Varchar()
+
+        for v in ["ignore", "allow", "forbid"]:
+
+            class MyConfig(pydantic.BaseConfig):
+                extra = v
+
+            model = create_pydantic_model(Band, pydantic_config_class=MyConfig)
+            self.assertEqual(model.Config.extra, v)
+
+    def test_pydantic_invalid_extra_fields(self):
+        """
+        Make sure that invalid values for ``pydantic_extra_fields``
+        are rejected.
+        """
+
+        class Band(Table):
+            name = Varchar()
+
+        class MyConfig(pydantic.BaseConfig):
+            extra = "foobar"
+
+        with pytest.raises(ValueError):
+            create_pydantic_model(Band, pydantic_config_class=MyConfig)


dantownsend:
@waldner Thanks a lot for this - it looks good.

I need to check if passing the choices to the Pydantic model has any effect on Piccolo Admin. If so, we might just have to add an option - like `include_choices`, so we can turn it off for Piccolo Admin.

@sinisaos I don't know if this is something we've tested before - do you remember?

sinisaos:
@dantownsend I only remember this [PR](https://github.com/piccolo-orm/piccolo/pull/469) and the integer validation didn't work. I hope this PR fixes that. We probably need to make changes to Piccolo Admin as well.

waldner:
My understanding is that piccolo admin was using the `choices` field of the `extra` attribute of the schema (as per https://github.com/piccolo-orm/piccolo/issues/467#issuecomment-1077419951), so I did not touch that in this patch. But I don't know the internals of piccolo admin.

dantownsend:
@waldner You're right - Piccolo Admin just uses the `extra` attribute to decide what to the render in the UI.

I'd better check quickly though, as having stronger validation on the Pydantic models might cause Piccolo Admin to fail in some way when posting back data.

@sinisaos Thanks for looking into it.

waldner:
So FWIW, it does indeed seem that piccolo admin has trouble with the new changes.

Using the following table schema:

```
class MyStrChoices(str, Enum):
    choice1 = 'choice1'
    choice2 = 'choice2'

class MyIntChoices(int, Enum):
    choice1 = 1
    choice2 = 6

class Task(Table):
    """
    An example table.
    """

    name = Varchar()
    stringcol = Varchar(choices=MyStrChoices)
    integercol = Integer(choices=MyIntChoices)
```

When I try to add a new row to the table using piccolo admin, no fields are shown and the following error is produced in the python interpreter (only the relevant part):

```
...
  File "pydantic/main.py", line 342, in pydantic.main.BaseModel.__init__
pydantic.error_wrappers.ValidationError: 2 validation errors for TaskOptional
stringcol
  value is not a valid enumeration member; permitted: 'choice1', 'choice2' (type=type_error.enum; enum_values=[<MyStrChoices.choice1: 'choice1'>, <MyStrChoices.choice2: 'choice2'>])
integercol
  value is not a valid enumeration member; permitted: 1, 6 (type=type_error.enum; enum_values=[<MyIntChoices.choice1: 1>, <MyIntChoices.choice2: 6>])
Server error
```

presumably because piccolo admin would set those fiels as empty initially, or with a placeholder value not among the allowed choices. What about setting it to the first value of the enumeration?

waldner:
If the propagation of `choices` to pydantic proves to be too complicated, for now I can rollback that part of the PR and only keep the `extra` policy part, which should not be problematic at all.

sinisaos:
@waldner As @dantownsend already said, Piccolo admin uses an `extra` attribute for rendering in UI and in it uses https://github.com/piccolo-orm/piccolo/blob/94c55f803f7757d266b9dd2d2498e44c52467224/piccolo/columns/base.py#L217 whose output is the `display_name` to nice display in the select dropdown and the` value` it uses as data. 
```
"choices":{
   "male":{
      "display_name":"Male", <-nice display in the select dropdown
      "value":"m" <- data
   },
   "female":{
      "display_name":"Female",
      "value":"f"
   },
   "non_binary":{
      "display_name":"Non Binary",
      "value":"n"
   }
},
```
I think with your changes we would have to use schema definitions for choices, but we need to see how or ignore it for Piccolo Admin.

dantownsend:
It would be great if Piccolo Admin worked properly with Pydantic choices, but I'm not sure how much work it will be currently.

So maybe lets move it out of this PR, and into another one, so we can get this merged in.

piccolo/utils/pydantic.py
```diff
@@ -90,6 +90,7 @@ def create_pydantic_model(
     deserialize_json: bool = False,
     recursion_depth: int = 0,
     max_recursion_depth: int = 5,
+    pydantic_extra_fields: str = "ignore",
```
dantownsend, line 93: Sorry - one minor thing. Can we make this:

```python
pydantic_extra_fields: t.Literal['ignore', 'allow', 'forbid'] = "ignore",
```

waldner: Originally I actually wanted to do this:

```
pydantic_extra_fields: pydantic.Extra = pydantic.Extra.ignore
```

Is there a stylistic or other reason for your suggested change? (I'm really curious, not trying to be confrontational)

dantownsend: Yeah, that works too:

```python
pydantic_extra_fields: pydantic.Extra = pydantic.Extra.ignore
```

I don't think there's clearly a best. The reason I suggested `Literal` is it means a linter can detect typos in the string.

I think the `Enum` is probably the best though. It's slightly more verbose to use, but means that if Pydantic adds another option in the future, we don't have to update `Literal` to include the new options.

Ideally it would be good to expose all of Pydantic's config options. We could take a different approach entirely, and do this:

```python
def create_pydantic_model(pydantic_config: t.Type[BaseConfig] = BaseConfig, ...):
    ...
    class CustomConfig(pydantic_config):
        schema_extra = {
            "help_text": table._meta.help_text,
            **schema_extra_kwargs,
        }
        extra = pydantic.Extra(pydantic_extra_fields)
```

Which means that every Pydantic option is exposed to the user. What do you think?

waldner: Looks good to me . I'll work on this.

waldner: I see there are quite a few places in piccolo-api that import `Config` from `utils/pydantic`, IIUC that class would now go away since we'd now be instantiating just `CustomConfig` based on `pydantic_config`, shall I (try to) adjust those other places too?

dantownsend: Ah, yeah - good point. I wonder if we keep `Config`, then this is possible?

```python
def create_pydantic_model(pydantic_config: t.Type[BaseConfig] = BaseConfig, ...):
    ...
    class CustomConfig(Config, pydantic_config):
        schema_extra = {
            "help_text": table._meta.help_text,
            **schema_extra_kwargs,
        }
        extra = pydantic.Extra(pydantic_extra_fields)
```

I think the `mro` still works, even though both classes inherit from `BaseConfig`.

waldner: Sorry for being a PITA, but now I wonder what we should do with the original `pydantic_extra_fields` parameter, as now the user could just do

```
class MyConfig(pydantic.BaseConfig):
    extra = 'forbid'

model = create_pydantic_model(pydantic_config=MyConfig, ...)
```

and obtain the same result as passing `pydantic_extra_fields='forbid'`. I don't have a strong opinion on this, although using `pydantic_extra_fields` seems slightly more convenient. Comments?

dantownsend: We probably don't need them both. I agree that having a `pydantic_extra_fields` argument is more convenient. But over time, as we support more Pydantic options, `create_pydantic_model` could end up with loads of arguments.

We could accept a dict, and anything passed into it automatically applied to the config (rather than having to pass in a config object):

```python
create_pydantic_model(pydantic_config={'extra': Extra.ignore})
```

It's more convenient, but there's no type checking. We could use `TypedDict`, but would constantly be updating it as Pydantic adds more arguments.

Sorry, I'm kind of thinking out loud here ... what do you think?

dantownsend:
@waldner This is great, thanks! Sorry about all of the back and forth. I think we have a nice design.

PR #719, opened 2022-12-14
an empty string is not allowed when required=True

sinisaos:
An empty string is not allowed when `Varchar` argument is `required=True`. Related to [this issue](https://github.com/piccolo-orm/piccolo_admin/issues/261)

### DIFF
--- a/piccolo/utils/pydantic.py
+++ b/piccolo/utils/pydantic.py
@@ -48,6 +48,12 @@ def pydantic_json_validator(cls, value, field):
         return value
 
 
+def pydantic_empty_string_validator(cls, value):
+    if value == "":
+        raise ValueError("Required field cannot be empty")
+    return value
+
+
 def is_table_column(column: Column, table: t.Type[Table]) -> bool:
     """
     Verify that the given ``Column`` belongs to the given ``Table``.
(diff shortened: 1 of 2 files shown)


piccolo/utils/pydantic.py
```diff
@@ -221,7 +227,15 @@ def create_pydantic_model(
         elif isinstance(column, Email):
             value_type = pydantic.EmailStr
         elif isinstance(column, Varchar):
-            value_type = pydantic.constr(max_length=column.length)
+            if not column._meta.required:
```
dantownsend, line 230: Having the truthy statement first makes the logic easier to follow. For example:

```python
if something is True:
    ...
else:
    ...
```

piccolo/utils/pydantic.py
```diff
@@ -221,7 +227,15 @@ def create_pydantic_model(
...
+                value_type = pydantic.constr(max_length=column.length)
+            else:
+                value_type = column.value_type
```
dantownsend, line 233: Shouldn't `value_type` still be `pydantic.constr(max_length=column.length)`?

piccolo/utils/pydantic.py
```diff
@@ -221,7 +227,15 @@ def create_pydantic_model(
         elif isinstance(column, Email):
             value_type = pydantic.EmailStr
         elif isinstance(column, Varchar):
```
dantownsend, line 229: This could also be true of `Text` columns.

Issue #749, opened 2023-01-07
auto-generated fastapi 0.89.0 error


mwmeyer:
Heads up: 

It looks like the [fastapi 0.89.0 release](https://github.com/tiangolo/fastapi/releases/tag/0.89.0) breaks the asgi code generated by piccolo ( `piccolo asgi new` )

```
(venv) $ python main.py
INFO:     Will watch for changes in these directories: ['/...']
INFO:     Uvicorn running on http://127.0.0.1:8000 (Press CTRL+C to quit)
INFO:     Started reloader process [47337] using StatReload
...
  File "/.../app.py", line 21, in <module>
    create_admin(
  File "/.../venv/lib/python3.9/site-packages/piccolo_admin/endpoints.py", line 1085, in create_admin
    return AdminRouter(
  File "/.../venv/lib/python3.9/site-packages/piccolo_admin/endpoints.py", line 523, in __init__
    private_app.add_api_route(
  File "/.../venv/lib/python3.9/site-packages/fastapi/applications.py", line 304, in add_api_route
    self.router.add_api_route(
  File "/.../venv/lib/python3.9/site-packages/fastapi/routing.py", line 572, in add_api_route
    route = route_class(
  File "/.../venv/lib/python3.9/site-packages/fastapi/routing.py", line 400, in __init__
    self.response_field = create_response_field(
  File "/.../venv/lib/python3.9/site-packages/fastapi/utils.py", line 90, in create_response_field
    raise fastapi.exceptions.FastAPIError(
fastapi.exceptions.FastAPIError: Invalid args for response field! Hint: check that <class 'starlette.responses.JSONResponse'> is a valid pydantic field type
```

Down-grading to `0.88.0` fixed the error.

dantownsend:
@mwmeyer Thanks for the heads up - we'll get a fix out for this. We could version pin to `0.88.0` for now.

dantownsend:
This should be fixed now. We release a new version of Piccolo Admin - `piccolo_admin==0.41.0` which works with the new FastAPI version.

PR #750, opened 2023-01-08
version pin fastapi

dantownsend:
Fixes https://github.com/piccolo-orm/piccolo/issues/749

FasAPI has some breaking changes in version `0.89.0`, so version pinning to `0.88.0` for now.

This is a temporary fix - would be better to fix the ASGI template so it works with the most recent FastAPI version.

### DIFF
--- a/piccolo/apps/asgi/commands/new.py
+++ b/piccolo/apps/asgi/commands/new.py
@@ -14,6 +14,7 @@ ROUTERS = ["starlette", "fastapi", "blacksheep", "xpresso", "starlite"]
 ROUTER_DEPENDENCIES = {
     "starlite": ["starlite>=1.46.0"],
     "xpresso": ["xpresso==0.43.0", "di==0.72.1"],
+    "fastapi": ["fastapi==0.88.0"],
 }
 
 


sinisaos:
@dantownsend Fix is in Piccolo Admin. If we replace https://github.com/piccolo-orm/piccolo_admin/blob/cba193071639f94b3240e9f515853e9a4885e13b/piccolo_admin/endpoints.py#L820 with `t.Any` everything is fine. I can do a PR.

dantownsend:
@sinisaos It seems like the latest FastAPI version lets you determine the `response_model` using type annotations. But the implementation seems flawed if it tries treating `JSONResponse` as a Pydantic model, and then fails.

We could try replacing this:

```
from starlette.responses import HTMLResponse, JSONResponse
```

With:

```
from fatapi.responses import HTMLResponse, JSONResponse
```

Just in case FastAPI handles it's own `JSONResponse` correctly.

Alternatively, we can just remove the response type:

```python
async def post_single_form(self, request: Request, form_slug: str):
    ...
```

sinisaos:
> With:
> 
> ```
> from fatapi.responses import HTMLResponse, JSONResponse
> ```
> 
> Just in case FastAPI handles it's own `JSONResponse` correctly.
>

I already tried that, but it doesn't work as it should either.
 
> Alternatively, we can just remove the response type:
> 
> ```python
> async def post_single_form(self, request: Request, form_slug: str):
>     ...
> ```
Or add `t.Any` to response_model. I was just creating a PR for that fix when I received a notification about your PR in Piccolo ORM. If you want, I can do it with `t.Any` or remove the response type, as you want, and then we don't need version pin PR in Piccolo ORM.

dantownsend:
@sinisaos Yeah, I'll close this PR. It doesn't seem to fix the problem anyway. As you say, the problem seems to be with Piccolo Admin instead.

Having `-> t.Any` should be the same as removing it, so both should work.

The only other thing we could do is return a Pydantic model instead. But if it's a 422 error, we return a different data structure, so maybe we leave that for now.

sinisaos:
> The only other thing we could do is return a Pydantic model instead. But if it's a 422 error, we return a different data structure, so maybe we leave that for now.

We could return `t.Dict[str, t.Any]`, but as you say we need a `status_code=422` response if there is an error. I will make a PR and remove the response model and you close this PR.

dantownsend:
@sinisaos OK, lets just return the response type for now.

dantownsend:
@sinisaos That seemed to solve the problem - I ran the integration tests, and now they pass 👍

dantownsend:
@sinisaos Do you think we should version pin the routers anyway in `piccolo asgi new`?

sinisaos:
@dantownsend I don't think this is necessary because custom forms response `t.Any` also works with previous versions of FastAPI.

dantownsend:
@sinisaos You're right - it's backwards compatible. It's just that sometimes changes to the router libraries break our CI. I guess it's good in a way though, as it alerts us to make changes. OK, we won't version pin the routers for now.

PR #755, opened 2023-01-31
fix output of alias usage (only band names)

pdpotter:


dantownsend:
@pdpotter Good catch - thanks!

PR #772, opened 2023-02-24
update Starlite asgi template

sinisaos:
Updated Starlite asgi app template-


dantownsend:
@sinisaos Looks good - thanks!

Issue #785, opened 2023-03-06
Sub queries for `is_in` and `not_in` clause


dantownsend:
It will allow us to do this (using this [example schema](https://piccolo-orm.readthedocs.io/en/latest/piccolo/getting_started/example_schema.html)):

```python
bands = await Band.objects().where(
    Band.id.is_in(
        Concert.select(Concert.band_1).where(
            Concert.starts_on > some_datetime
        )
    )
)
```

Currently it would be two separate queries, which isn't a huge hassle, but less efficient:

```python
band_ids = await Concert.select(Concert.band_1).where(
    Concert.starts_on > some_datetime
).output(as_list=True)

bands = await Band.objects().where(
    Band.id.is_in(band_ids)
)
```

Adding this should be pretty straight forward. Here are the [Postgres docs](https://www.postgresql.org/docs/current/functions-subquery.html).

It's a very common use case which I come across all of the time.

PR #786, opened 2023-03-06
Sub queries for `is_in` and `not_in`

dantownsend:
Resolves https://github.com/piccolo-orm/piccolo/issues/785


sinisaos:
@dantownsend Sorry to bother, but I wonder why you didn't merge this (I know the tests are missing) because I think it's really useful and convenient, and the performance is approx. 50% better on a GET request because we don't have to make two separate queries but just one?

dantownsend:
@sinisaos Yeah, it's a good point. Like you say, it just needs test and docs, and they shouldn't take too long to add. I'll try and finish it up and get it merged.

sinisaos:
@dantownsend Great. Thanks.

powellnorma:
Can we perhaps make it so that piccolo handles the case where we put in an empty list? It can be useful when the list is dynamically generated, and there are different conditions in the where clause. Maybe [this](https://stackoverflow.com/a/13210306) would work?

dantownsend:
@powellnorma I usually have to add a check before any query with an `is_in` to make sure the list isn't empty. It has tripped me up a few times.

PR #788, opened 2023-03-12
Typo: "import OnDelete" -> "import OnUpdate"

BurkhalterY:
Import was also wrong.
Sorry for the double PR.


dantownsend:
@BurkhalterY Ah, totally missed that - good catch!

Issue #707, opened 2022-12-05
Piccolo doesn't handle duplicate entries automatically for in .add_m2m 


Drapersniper:
### Discussed in https://github.com/piccolo-orm/piccolo/discussions/683

<div type='discussions-op-text'>

<sup>Originally posted by **Drapersniper** November 21, 2022</sup>
```py

class Track(Table):
    track_id = BigInt(primary_key=True)

class Playlist(Table):
    id = BigInt(primary_key=True)
    tracks = M2M(LazyTableReference("TrackToPlaylist", module_path=__module__))

class Album(Table):
    id = BigInt(primary_key=True)
    tracks = M2M(LazyTableReference("TrackToAlbum", module_path=__module__))

class TrackToPlaylist(Table):
    tracks = ForeignKey(Track)
    playlists= ForeignKey(Playlist)    

class TrackToAlbum(Table):
    tracks = ForeignKey(Track)
    albums= ForeignKey(Album)

# Step 1 -  I expect this to work fine if this is the first time the tracks with id 0-9 are being added to the db
playlist = await Playlist.objects().get_or_create(Playlist.id == 1)
tracks = [Track(track_id=i) for i in range(10)]
await playlist.add_m2m(*tracks, m2m=Playlist.tracks)

# Step 2 - After step 1 do the following
playlist2 = await Playlist.objects().get_or_create(Playlist.id == 2)
tracks2 = [Track(track_id=i) for i in range(5)]
await playlist2 .add_m2m(*tracks2, m2m=Playlist.tracks)

# Step 3
playlist3 = await Playlist.objects().get_or_create(Playlist.id == 3)
tracks3 = [Track(track_id=i) for i in range(5)]
tracks3 +=  [Track(track_id=i) for i in range(5)]
await playlist3 .add_m2m(*tracks3, m2m=Playlist.tracks)

# Step 4 - Run after step 1
album = await Album.objects().get_or_create(Album.id == 1)
tracks4 = [Track(track_id=i) for i in range(5)]
await album .add_m2m(*tracks4, m2m=Album.tracks)

```

What is expected to happen after steps 2, 3, and 4?

</div>

Currently it raises an exception

sinisaos:
[...]

PR #798, opened 2023-03-31
Api for upserting

sinisaos:
Related to #252

### DIFF
--- a/piccolo/query/methods/insert.py
+++ b/piccolo/query/methods/insert.py
@@ -6,7 +6,7 @@ from piccolo.custom_types import TableInstance
 from piccolo.query.base import Query
 from piccolo.query.mixins import (
     AddDelegate,
-    Conflict,
+    OnConflict,
     OnConflictDelegate,
     ReturningDelegate,
 )
@@ -24,13 +24,12 @@ class Insert(
         "add_delegate",
         "returning_delegate",
         "on_conflict_delegate",
-        "values_delegate",
     )
 
     def __init__(
         self,
         table: t.Type[TableInstance],
-        on_conflict: t.Optional[Conflict] = None,
+        on_conflict: t.Optional[OnConflict] = None,
         *instances: TableInstance,
         **kwargs,
     ):
@@ -52,7 +51,7 @@ class Insert(
         self.returning_delegate.returning(columns)
         return self
 
-    def on_conflict(self: Self, conflict: Conflict) -> Self:
+    def on_conflict(self: Self, conflict: OnConflict) -> Self:
         self.on_conflict_delegate.on_conflict(conflict)
         return self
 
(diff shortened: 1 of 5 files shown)


dantownsend:
Thanks for this, and the docs.

sinisaos:
No problem. It would be good if you have time to try and review this because if you merge this, it would also solve the problem with [duplicate entries in the M2M](https://github.com/piccolo-orm/piccolo/issues/707), because then we will be able to add the `on_conflict` parameter to the `unused` rows insert query like this
```python
if unsaved:
  await rows[0].__class__.insert(
      *unsaved, on_conflict=OnConflict.do_nothing
  ).run()
```

piccolo/query/methods/insert.py
```diff
@@ -55,21 +74,47 @@ def _raw_response_callback(self, results):
...
     def default_querystrings(self) -> t.Sequence[QueryString]:
-        base = f'INSERT INTO "{self.table._meta.tablename}"'
+        engine_type = self.engine_type
+        if (
+            engine_type == "sqlite"
+            and self.table._meta.db.get_version_sync() < 3.24
+        ):  # pragma: no cover
+            if self.on_conflict_delegate._on_conflict == OnConflict.do_nothing:
+                base = f'INSERT OR IGNORE INTO "{self.table._meta.tablename}"'
+            else:
```
dantownsend, line 84: I would explicitly check this against `OnConflict.do_update`, and have the else clause raise a `ValueError`.

sinisaos: I need clarification. Which `ValueError` should I raise because the user only has two valid options (nothing or update) and if the user enters a wrong value for `on_conflict` the `AttributeError` is already raised? If no `on_conflict` clause is specified, the standard insert query is executed.

dantownsend: The value is meant to be an enum but because it's Python could be anything if the user set it to say a string by accident.

```python
if self.on_conflict_delegate._on_conflict == OnConflict.do_nothing:
    base = f'INSERT OR IGNORE INTO "{self.table._meta.tablename}"'
elif self.on_conflict_delegate._on_conflict == OnConflict.do_update:
    ...
else:
    raise ValueError('Invalid on conflict value')

Comment (sinisaos) [reply piccolo/query/methods/insert.py:84]: OK. I will change that.

File: piccolo/query/methods/insert.py [hunk 2]
```diff
@@ -36,6 +51,10 @@ def returning(self: Self, *columns: Column) -> Self:
         self.returning_delegate.returning(columns)
         return self
 
+    def on_conflict(self: Self, conflict: OnConflict) -> Self:
```
dantownsend, line 54: Do you think we should allow the user to pass in specific columns?

```python
await Band.insert(Band(name="Pythonistas")).on_conflict(Band.name, do_nothing=True)
```

piccolo/query/methods/insert.py
```diff
@@ -15,14 +20,24 @@
...
+        "add_delegate",
+        "returning_delegate",
+        "on_conflict_delegate",
+    )
 
     def __init__(
-        self, table: t.Type[TableInstance], *instances: TableInstance, **kwargs
+        self,
+        table: t.Type[TableInstance],
+        on_conflict: t.Optional[OnConflict] = None,
```
dantownsend, line 32: I would drop consider dropping this, and just using the `on_conflict` method instead.

The reason being, we can add extra arguments to the `on_conflict` method in the future, but we don't want to add too many to `__init__`.

piccolo/query/methods/insert.py
sinisaos: If you look at the tests, even now the user can update one or more columns on `do_update` conflict. For `do_nothing` conflict I don't see point of allowing specifying columns because on `do_nothing` conflict columns don't change anyway.

piccolo/query/methods/insert.py
sinisaos: I don't think we should change it, but feel free to do as you think is best. With this method, we can also easily solve the [problem with duplicate entries in M2M](https://github.com/piccolo-orm/piccolo/issues/707). I'm sorry if I didn't understand well what you wanted to say and feel free to change this if you think `on_conflict` as method would be better way.

sinisaos:
Implemented in #816

Issue #812, opened 2023-04-24
allow .distinct() to specify a column for DISTINCT ON


dantownsend:
### Discussed in https://github.com/piccolo-orm/piccolo/discussions/808

<div type='discussions-op-text'>

<sup>Originally posted by **williamflaherty** April 18, 2023</sup>
As the title says, would be great if you could specify a column in TableName.select().distinct() to handle distinct on or something similar? 

I started poking around to add it myself but got stopped up by order_by needing to be in the same order issues from asyncpg. (I'm far from a PostgreSQL expert). </div>

PR #813, opened 2023-04-24
initial implementation for `DISTINCT ON`

dantownsend:
Resolves https://github.com/piccolo-orm/piccolo/issues/812

Inspired by https://github.com/piccolo-orm/piccolo/compare/master...sinisaos:piccolo:distinct_on_clause by @sinisaos 

Remaining tasks:

- [x] Make sure a `ValueError` is raised if being used by SQLite
- [x] ~~Make sure joins work (should be OK, as we need to specify the same column for order_by for it to work anyway)~~
- [x] Use a different example table, to make it consistent with other tests
- [x] Adapt docs from @sinisaos PR
- [x] Add test to make sure passing in something other than a sequence raises an error.

### DIFF
--- a/tests/table/test_select.py
+++ b/tests/table/test_select.py
@@ -3,10 +3,11 @@ from unittest import TestCase
 import pytest
 
 from piccolo.apps.user.tables import BaseUser
+from piccolo.columns import Varchar
 from piccolo.columns.combination import WhereRaw
 from piccolo.query import OrderByRaw
 from piccolo.query.methods.select import Avg, Count, Max, Min, SelectRaw, Sum
-from piccolo.table import create_db_tables_sync, drop_db_tables_sync
+from piccolo.table import Table, create_db_tables_sync, drop_db_tables_sync
 from tests.base import (
     DBTestCase,
     engine_is,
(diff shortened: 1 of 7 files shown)


tests/table/test_select.py
```diff
@@ -1237,3 +1259,53 @@ def test_order_by_raw(self):
...
+
+
+class TestDistinctOn(TestCase):
+    def setUp(self):
+        Color.create_table().run_sync()
+
+    def tearDown(self):
+        Color.alter().drop_table().run_sync()
+
+    @engines_only("postgres", "cockroach")
```
sinisaos, line 1276: Be careful with this. `DISTINCT ON` works in CockroachDB, but I didn't know how to write tests for Cockroach because there is no sequential primary key and the order of result is different than in Postgres.

williamflaherty: This looks good to me. Much cleaner than my roughshod take when I first realized it wasn't supported (and done so quickly too!) I will say that I'd prefer if the library did the order_by validation, makes things really clear rather than letting asyncpg or the database tell me something. Sorry I wasn't more help thank you both for your efforts!

dantownsend: @williamflaherty Thanks - will see if I we can add some validation.

dantownsend: I've added validation now.

dantownsend:
@sinisaos @williamflaherty I think this is good to go now. If you both agree will merge and release it.

sinisaos (approved):
It works and looks great.

PR #816, opened 2023-04-28
On conflict clause / upserts

dantownsend:
Resolves https://github.com/piccolo-orm/piccolo/issues/252

Based on https://github.com/piccolo-orm/piccolo/pull/798 by @sinisaos with the following changes:

* We're not supporting older SQLite versions, to keep things simpler.
* Gives a bit more granular control over which constraints are used, and which values to update.

Remaining tasks:

- [x] Update docs
- [x] Add more tests
- [x] Raise an exception for older SQLite versions


tests/table/test_insert.py
```diff
@@ -76,3 +81,23 @@ def test_insert_returning_alias(self):
         )
 
         self.assertListEqual(response, [{"manager_name": "Maz"}])
+
+
+class TestOnConflict(DBTestCase):
```
sinisaos, line 86: You can use all the tests from my PR (with modification to the new` on_conflict` method) as they cover all possibilities (single column, multiple columns, etc.) for all supported databases.

piccolo/query/mixins.py
```diff
@@ -613,12 +616,132 @@ def __str__(self):
...
+        return self.querystring.__str__()
+
+
+@dataclass
+class OnConflictDelegate:
+    """
+    Used with insert queries to specify what to do when a query fails due to
+    a constraint::
+
+        .on_conflict(action='DO NOTHING')
```
sinisaos, line 717: I can't try PR right now but it looks great to me. This should also work for duplicate M2M entries like this
```python
if unsaved:
  await rows[0].__class__.insert(*unsaved).on_conflict(action='DO NOTHING').run()
```

tests/table/test_insert.py
dantownsend: Yes, I'll reuse most of those tests - thanks.

piccolo/query/mixins.py
dantownsend: Yes, hopefully 🤞

sinisaos: I just checked this and it works.

sinisaos:
@dantownsend What is the syntax for updating multiple `on_conflict` columns? I tried a list of tuples like this 
```python
values=[(Band.name, "Javas"),(Band.popularity, 1000)] 
```
but it doesn't work. How do we write test like this https://github.com/piccolo-orm/piccolo/pull/798/files#diff-fac5af1d6e85e27669c58e982989dfcf8327f312f8b29a272a1188f86ecb7f39R126-R127 or that's not possible at this time?

dantownsend:
> @dantownsend What is the syntax for updating multiple on_conflict columns? I tried a list of tuples like this

@sinisaos My thinking was that the `values` list can either take a column, or a tuple. If it's a column, it uses the new value. If it's a tuple, it uses the value specified in the tuple.

```python
# If there's a conflict, it sets the popularity to 2000:
await Band.insert(
    Band(name='Pythonistas', popularity=2000)
).on_conflict(
    action='DO UPDATE',
    target=[Band.name],
    values=[Band.popularity]
)

# If there's a conflict, it sets the popularity to something custom:
await Band.insert(
    Band(name='Pythonistas', popularity=2000)
).on_conflict(
    action='DO UPDATE',
    target=[Band.name],
    values=[(Band.popularity, 3000)]
)
```

I haven't had time to fully test this though - so there may be bugs.

piccolo/query/mixins.py
dantownsend: Thanks 👍

sinisaos:
@dantownsend I understand it for one column but how to do it for multiple columns. With the old API we didn't have to set any columns or values ​​because **Piccolo did everything** for us. For example if we have rows like this
```python
[
    {'id': 1, 'name': 'Band 1', 'popularity': 1}, 
    {'id': 2, 'name': 'Band 2', 'popularity': 2}, 
    {'id': 3, 'name': 'Band 3', 'popularity': 3}
]
```
we can set one or more columns to update `on_conflict` like this for one column (e.g `Band.popularity` change and `Band.name` stay the same)
```python
await Band.insert(
    Band(id=1, name="Band 1", popularity=111),
    Band(id=2, name="Band 2", popularity=222),
    Band(id=3, name="Band 3", popularity=333),
    on_conflict=OnConflict.do_update,
).run()
# result
[
    {'id': 1, 'name': 'Band 1', 'popularity': 111}, 
    {'id': 2, 'name': 'Band 2', 'popularity': 222}, 
    {'id': 3, 'name': 'Band 3', 'popularity': 333}
]
```
or for multiple column (`Band.popularity` and `Band.name` both changes)
```python
await Band.insert(
    Band(id=1, name="Band 1111", popularity=1111),
    Band(id=2, name="Band 2222", popularity=2222),
    Band(id=3, name="Band 3333", popularity=3333),
    on_conflict=OnConflict.do_update,
).run()
# result
[
    {'id': 1, 'name': 'Band 1111', 'popularity': 1111}, 
    {'id': 2, 'name': 'Band 2222', 'popularity': 2222}, 
    {'id': 3, 'name': 'Band 3333', 'popularity': 3333}
]
```
How can we do this with the new API, and if not implemented, please do. Sorry for long comment.

sinisaos:
I'm sorry, but that doesn't work. Both `Sqlite` and `Postgres` throw a syntax error (`sqlite3.OperationalError: near ""popularity"": syntax error` and `asyncpg.exceptions.PostgresSyntaxError: syntax error at or near ""popularity""`). I hope this helps you find what the problem is.

sinisaos:
@dantownsend Yes. Missing comma between EXCLUDED columns after UPDATE SET in querystring. This is log
`INSERT INTO "band" ("id","name","popularity") VALUES (1,'Band 1',1),(2,'Band 2',2),(3,'Band 3',3) ON CONFLICT ("id") DO UPDATE SET "name"=EXCLUDED."name" "popularity"=EXCLUDED."popularity"`

piccolo/query/mixins.py
```diff
@@ -613,12 +616,132 @@ def __str__(self):
...
+                values = []
+                query = f"{OnConflictAction.do_update.value} SET"
+
+                if not self.values:
+                    raise ValueError("No values specified for `on conflict`")
+
+                for value in self.values:
+                    if isinstance(value, Column):
+                        column_name = value._meta.db_column_name
+                        query += f' "{column_name}"=EXCLUDED."{column_name}"'
```
sinisaos, line 675:
```suggestion
                        query += f' "{column_name}"=EXCLUDED."{column_name}",'
```
for multiple columns

piccolo/query/mixins.py
```diff
@@ -613,12 +616,132 @@ def __str__(self):
...
+                            column_name = column._meta.db_column_name
+                        elif isinstance(column, str):
+                            column_name = column
+                        else:
+                            raise ValueError("Unsupported column type")
+
+                        query += f' "{column_name}"={{}}'
+                        values.append(value_)
+
+                return QueryString(query, *values)
```
sinisaos, line 689:
```suggestion
                return QueryString(query.rstrip(","), *values)
```
for multiple columns

sinisaos:
@dantownsend I'm sorry if I caused problems because I never used Github suggested changes, but it seemed to me that it was easiest to show what changes I think should be made.

dantownsend:
>  @dantownsend I'm sorry if I caused problems because I never used Github suggested changes, but it seemed to me that it was easiest to show what changes I think should be made.

It's good - works well. Thanks for suggesting those changes 👍

dantownsend:
I've added some docs. Just need to decide on the names of the arguments.

It's currently:

```python
>>> await Band.insert(
...     Band(name="Pythonistas", popularity=1200)
... ).on_conflict(
...     action="DO UPDATE",
...     target=[Band.name],
...     values=[Band.popularity],
... )
```

The names `action`, `target` and `values` were inspired by the Postgres docs, but I'm wondering if this is better?

```python
>>> await Band.insert(
...     Band(name="Pythonistas", popularity=1200)
... ).on_conflict(
...     action="DO UPDATE",
...     constraints=[Band.name],
...     update=[Band.popularity],
... )
```

sinisaos:
> The names `action`, `target` and `values` were inspired by the Postgres docs, but I'm wondering if this is better?

Yeah, naming is always difficult. I think we should stick with names inspired by Postgres docs because people are more familiar with Postgres naming conventions.

sinisaos:
@dantownsend Sorry for the long comment, but these are modified tests for Postgres and Cockroach (without assertion error). You can just copy and paste it (hope it saves your time).
```python
@postgres_only
def test_insert_on_conflict_do_nothing(self):
  """
  Check that the record has not changed because of the
  `on_conflict` clause.
  """
  self.insert_rows()

  Band.insert(
      Band(id=1, name="Javas", popularity=100),
  ).on_conflict(action="DO NOTHING").run_sync()

  response = (
      Band.select(Band.name).where(Band.id == 1).first().run_sync()
  )
  self.assertEqual(response["name"], "Pythonistas")

@postgres_only
def test_insert_on_conflict_do_update_single_column(self):
  """
  Check that the record has changed because of the
  `on_update` clause.
  """
  self.insert_rows()

  Band.insert(
      Band(id=1, name="Pythonstas-updated", manager=1, popularity=1000),
      Band(id=2, name="Rustaceans-updated", manager=2, popularity=2000),
      Band(id=3, name="CSharps-updated", manager=3, popularity=10),
  ).on_conflict(
      action="DO UPDATE", target=[Band.id], values=[Band.name]
  ).run_sync()

  response = Band.select().run_sync()
  self.assertEqual(
      response,
      [
          {
              "id": 1,
              "name": "Pythonstas-updated",
              "manager": 1,
              "popularity": 1000,
          },
          {
              "id": 2,
              "name": "Rustaceans-updated",
              "manager": 2,
              "popularity": 2000,
          },
          {
              "id": 3,
              "name": "CSharps-updated",
              "manager": 3,
              "popularity": 10,
          },
      ],
  )

@postgres_only
def test_insert_on_conflict_do_update_single_column_tuple_values(self):
  """
  Check that the record has changed because of the
  `on_update` clause.
  """
  self.insert_rows()

  Band.insert(
      Band(id=1, name="Pythonstas-updated", manager=1, popularity=1000),
      Band(id=2, name="Rustaceans-updated", manager=2, popularity=2000),
      Band(id=3, name="CSharps-updated", manager=3, popularity=10),
  ).on_conflict(
      action="DO UPDATE",
      target=[Band.id],
      values=[(Band.name, "Rustaceans-updated")],
  ).run_sync()

  response = Band.select().run_sync()
  self.assertEqual(
      response,
      [
          {
              "id": 1,
              "name": "Rustaceans-updated",
              "manager": 1,
              "popularity": 1000,
          },
          {
              "id": 2,
              "name": "Rustaceans-updated",
              "manager": 2,
              "popularity": 2000,
          },
          {
              "id": 3,
              "name": "Rustaceans-updated",
              "manager": 3,
              "popularity": 10,
          },
      ],
  )

@postgres_only
def test_insert_on_conflict_do_update_multiple_columns(self):
  """
  Check that the record has changed because of the
  `on_update` clause.
  """
  self.insert_rows()

  Band.insert(
      Band(id=1, name="Pythonstas-updated", manager=3, popularity=200),
      Band(id=2, name="Rustaceans-updated", manager=2, popularity=1000),
      Band(id=3, name="CSharps-updated", manager=1, popularity=20),
  ).on_conflict(
      action="DO UPDATE",
      target=[Band.id],
      values=[Band.name, Band.popularity, Band.manager],
  ).run_sync()

  response = Band.select().run_sync()
  self.assertEqual(
      response,
      [
          {
              "id": 1,
              "name": "Pythonstas-updated",
              "manager": 3,
              "popularity": 200,
          },
          {
              "id": 2,
              "name": "Rustaceans-updated",
              "manager": 2,
              "popularity": 1000,
          },
          {
              "id": 3,
              "name": "CSharps-updated",
              "manager": 1,
              "popularity": 20,
          },
      ],
  )

@cockroach_only
def test_insert_on_conflict_do_nothing_cockroach(self):
  """
  Check that the record has not changed because of the
  `on_conflict` clause.
  """
  self.insert_rows()

  results = Band.select().run_sync()

  Band.insert(
      Band(
          id=results[0]["id"],
          name="Javas",
          manager=results[0]["manager"],
          popularity=100,
      )
  ).on_conflict(action="DO NOTHING").run_sync()

  response = (
      Band.select(Band.name)
      .where(Band.id == results[0]["id"])
      .first()
      .run_sync()
  )
  self.assertEqual(response["name"], "Pythonistas")

@cockroach_only
def test_insert_on_conflict_do_update_single_column_cockroach(self):
  """
  Check that the record has changed because of the
  `on_update` clause.
  """
  self.insert_rows()

  results = Band.select().run_sync()

  Band.insert(
      Band(
          id=results[0]["id"],
          name="Pythonstas-updated",
          manager=results[0]["manager"],
          popularity=1000,
      ),
      Band(
          id=results[1]["id"],
          name="Rustaceans-updated",
          manager=results[1]["manager"],
          popularity=2000,
      ),
      Band(
          id=results[2]["id"],
          name="CSharps-updated",
          manager=results[2]["manager"],
          popularity=10,
      ),
  ).on_conflict(
      action="DO UPDATE", target=[Band.id], values=[Band.name]
  ).run_sync()

  response = Band.select().run_sync()

  self.assertEqual(
      response,
      [
          {
              "id": results[0]["id"],
              "name": "Pythonstas-updated",
              "manager": results[0]["manager"],
              "popularity": 1000,
          },
          {
              "id": results[1]["id"],
              "name": "Rustaceans-updated",
              "manager": results[1]["manager"],
              "popularity": 2000,
          },
          {
              "id": results[2]["id"],
              "name": "CSharps-updated",
              "manager": results[2]["manager"],
              "popularity": 10,
          },
      ],
  )

@cockroach_only
def test_insert_on_conflict_do_update_single_column_tuple_values_cockroach(
  self,
):
  """
  Check that the record has changed because of the
  `on_update` clause.
  """
  self.insert_rows()

  results = Band.select().run_sync()

  Band.insert(
      Band(
          id=results[0]["id"],
          name="Pythonstas-updated",
          manager=results[0]["manager"],
          popularity=1000,
      ),
      Band(
          id=results[1]["id"],
          name="Rustaceans-updated",
          manager=results[1]["manager"],
          popularity=2000,
      ),
      Band(
          id=results[2]["id"],
          name="CSharps-updated",
          manager=results[2]["manager"],
          popularity=10,
      ),
  ).on_conflict(
      action="DO UPDATE",
      target=[Band.id],
      values=[(Band.name, "Rustaceans-updated")],
  ).run_sync()

  response = Band.select().run_sync()

  self.assertEqual(
      response,
      [
          {
              "id": results[0]["id"],
              "name": "Rustaceans-updated",
              "manager": results[0]["manager"],
              "popularity": 1000,
          },
          {
              "id": results[1]["id"],
              "name": "Rustaceans-updated",
              "manager": results[1]["manager"],
              "popularity": 2000,
          },
          {
              "id": results[2]["id"],
              "name": "Rustaceans-updated",
              "manager": results[2]["manager"],
              "popularity": 10,
          },
      ],
  )

@cockroach_only
def test_insert_on_conflict_do_update_multiple_columns_cockroach(self):
  """
  Check that the record has changed because of the
  `on_update` clause.
  """
  self.insert_rows()

  results = Band.select().run_sync()

  Band.insert(
      Band(
          id=results[0]["id"],
          name="Pythonstas-updated",
          manager=results[2]["manager"],
          popularity=200,
      ),
      Band(
          id=results[1]["id"],
          name="Rustaceans-updated",
          manager=results[1]["manager"],
          popularity=1000,
      ),
      Band(
          id=results[2]["id"],
          name="CSharps-updated",
          manager=results[0]["manager"],
          popularity=20,
      ),
  ).on_conflict(
      action="DO UPDATE",
      target=[Band.id],
      values=[Band.name, Band.popularity, Band.manager],
  ).run_sync()

  response = Band.select().run_sync()
  self.assertEqual(
      response,
      [
          {
              "id": results[0]["id"],
              "name": "Pythonstas-updated",
              "manager": results[2]["manager"],
              "popularity": 200,
          },
          {
              "id": results[1]["id"],
              "name": "Rustaceans-updated",
              "manager": results[1]["manager"],
              "popularity": 1000,
          },
          {
              "id": results[2]["id"],
              "name": "CSharps-updated",
              "manager": results[0]["manager"],
              "popularity": 20,
          },
      ],
  )
```

piccolo/query/methods/insert.py
```diff
@@ -70,16 +93,28 @@ def default_querystrings(self) -> t.Sequence[QueryString]:
...
+        # TODO - check SQLite version
+        on_conflict = self.on_conflict_delegate._on_conflict
+        if on_conflict:
+            querystring = QueryString(
+                "{}{}",
+                querystring,
+                on_conflict.querystring,
+                query_type="insert",
+                table=self.table,
+            )
```
sinisaos, line 105:
```suggestion
        if engine_type in ("postgres", "cockroach") or (
            engine_type == "sqlite"
            and self.table._meta.db.get_version_sync() >= 3.24
        ):
            on_conflict = self.on_conflict_delegate._on_conflict
            if on_conflict:
                querystring = QueryString(
                    "{}{}",
                    querystring,
                    on_conflict.querystring,
                    query_type="insert",
                    table=self.table,
                )
        else:
            raise ValueError(
                "Sqlite versions lower than 3.24 are not supported"
            )
```

dantownsend: Thanks. Having thought about it, I think we should move this check to the `Insert.on_conflict` method.

Also `NotImplementedError` is probably most appropriate as an exception.

```python
def on_conflict(...):
    if self.engine_type == 'sqlite' and self.table._meta.db.get_version_sync() < 3.34:
        raise NotImplementedError("SQLite versions lower than 3.24 are not supported") 
    ...
```

dantownsend: I pushed this change, and checked for other similar methods which are only available in certain databases - they should all return `NotImplementedError`.

sinisaos: Yes, that be much better.

piccolo/query/methods/insert.py
```diff
@@ -54,6 +54,14 @@ def on_conflict(
             t.List[t.Union[Column, t.Tuple[t.Union[str, Column], t.Any]]]
         ] = None,
     ) -> Self:
+        if (
+            self.engine_type == "sqlite"
+            and self.table._meta.db.get_version_sync() < 3.34
```
sinisaos, line 59: Should be version less than 3.24 `and self.table._meta.db.get_version_sync() < 3.24`

dantownsend: Oops - good catch!

dantownsend:
@sinisaos I think it's pretty much done now 🤞 

The only thing I know of missing is we're letting the user specify the target using a string. Postgres lets you specify the target using the name of a column, or an `index expression`:

https://www.postgresql.org/docs/current/sql-insert.html#index_expression

I need to read a bit more about how the `index expression` is meant to work.

tests/table/test_select.py
```diff
@@ -1364,12 +1364,12 @@ def test_distinct_on_sqlite(self):
         SQLite doesn't support ``DISTINCT ON``, so a ``ValueError`` should be
```
sinisaos, line 1364: I think it should be `NotImplementedError` instead `ValueError`.

sinisaos:
This works and looks great. Can I suggest two more tests that will cover tuple value as columns and tuple value as columns name in `values` ​​argument.
```python
def test_do_update_tuple_values(self):
    """
    We can use tuple values in ``values``.
    """
    Band = self.Band

    new_popularity = self.band.popularity + 1000
    new_name = "Rustaceans"

    Band.insert(
        Band(
            id=self.band.id,
            name=new_name,
            popularity=new_popularity,
        )
    ).on_conflict(
        action="DO UPDATE",
        targets=[Band.id],
        values=[
            (Band.name, new_name),
            (Band.popularity, new_popularity + 2000),
        ],
    ).run_sync()

    self.assertListEqual(
        Band.select().run_sync(),
        [
            {
                "id": self.band.id,
                "name": new_name,
                "popularity": new_popularity + 2000,
            }
        ],
    )

def test_do_update_tuple_values_string_column_name(self):
    """
    We can use string column name in tuple values.
    """
    Band = self.Band

    new_popularity = self.band.popularity + 1000
    new_name = "Rustaceans"

    Band.insert(
        Band(
            id=self.band.id,
            name=new_name,
            popularity=new_popularity,
        )
    ).on_conflict(
        action="DO UPDATE",
        targets=[Band.id],
        values=[
            ("name", new_name),
            ("popularity", new_popularity + 2000),
        ],
    ).run_sync()

    self.assertListEqual(
        Band.select().run_sync(),
        [
            {
                "id": self.band.id,
                "name": new_name,
                "popularity": new_popularity + 2000,
            }
        ],
    )
```

dantownsend:
> Can I suggest two more tests that will cover tuple value as columns and tuple value as columns name in values ​​argument.

Good point - I'll add those.

I realised I made a small mistake. When specifying the `target`, you can specify multiple columns. If you specify `column_a` and `column_b`, it means it's targeting a composite unique constraint for `column_a` and `column_b`. I realised when I was writing a test - need to update the docs to reflect this.

dantownsend:
I added a where clause, but it's failing on cockroach.

I also need to move it. The first one is just useful for conditional constraints, but the second one is more likely to be used:

Screenshot 2023-05-02 at 23 17 41

sinisaos:
> I added a where clause, but it's failing on cockroach.
> 
> I also need to move it. The first one is just useful for conditional constraints, but the second one is more likely to be used:

I agree with you and think you should move `self.where` below `self.actions` to get correct querystring like this 
```python
if self.action:
    query += " {}"
    values.append(self.action_string)

if self.where:
    query += " WHERE {}"
    values.append(self.where.querystring)
```
because Cockroach `ON CONFLICT` with `where` clause syntax is like this (from [docs](https://www.cockroachlabs.com/docs/v2.0/insert.html))
```
You can also use a WHERE clause to apply the DO UPDATE SET expression conditionally:

INSERT INTO accounts (id, balance)
    VALUES (8, 700)
    ON CONFLICT (id)
    DO UPDATE SET balance = excluded.balance
    WHERE excluded.balance > accounts.balance;
```
After that you should also update the test for `where`

```python
query = Band.insert(
    Band(name=self.band.name, popularity=new_popularity)
).on_conflict(
    target=Band.name,
    where=Band.popularity < new_popularity,
    action="DO UPDATE",
    values=[Band.popularity],
)

self.assertIn(
    f'WHERE "band"."popularity" < {new_popularity}',
    query.__str__(),
)
```
I hope that makes sense.

dantownsend:
@sinisaos Yes, you're right - thanks. I've made those changes.

Once the tests pass, I'll release this. There are some minor things missing, but I think this is fine for 99% of use cases.

dantownsend:
Follow up actions in the future:

- Maybe expose some of this via `save`.
- Allow conditional constraints (using `where`)
- Use `on_conflict='DO NOTHING'` with `Table.add_m2m`

sinisaos:
> * Use `on_conflict='DO NOTHING'` with `Table.add_m2m`

When you release the new Piccolo version, if you want, I can open a PR to do it as we discussed.

dantownsend:
@sinisaos Composite unique constraints is definitely something we need to add soon. I just haven't figured out a great API for it. Because most of Piccolo works with column references instead of strings.

Could be something like this (using strings):

```python
class Band(Table, constraints=[UniqueTogether('name', 'manager')]):
    name = Varchar()
    manager = ForeignKey()
```

Or this:

```python
class Band(Table):
    name = Varchar()
    manager = ForeignKey()
    constraints = [UniqueTogether('name', 'manager')])
```

This might work ... I'm not sure though:

```python
class Band(Table):
    name = Varchar()
    manager = ForeignKey()
    constraints = [UniqueTogether(name, manager)])
```

I considered something like this, but it's hard to track for migrations:

```python
class Band(Table):
    name = Varchar()
    manager = ForeignKey()

    @classmethod
    def constraints(cls):
        return [UniqueTogether(cls.name, cls.manager)]
```

If you have any ideas / preference let me know.

sinisaos:
@dantownsend If I may ask, why not use this PR #582?

dantownsend:
@sinisaos I think that PR is good. Just considering all the options, as once it's in, the API is very hard to change.

sinisaos:
@dantownsend These are all valid choices and I don't know how difficult it is to implement them, but I like this example the most.

> ```python
> class Band(Table):
>     name = Varchar()
>     manager = ForeignKey()
>     constraints = [UniqueTogether('name', 'manager')])
> ```

dantownsend:
@sinisaos Yeah, something like that probably makes more sense than passing the args into like `class Band(Table, constraints=...)`. 

Constraints and schema support are the main things to add next. Will try and go through that PR soon.

dantownsend:
@sinisaos Thanks for checking that - we know it's the right approach then.

Do you want to create a PR, adding `on_conflict` to that `insert` query? It might be worth merging in straight away, as at least the user can manually add the composite unique constraint.

PR #821, opened 2023-05-05
on_conflict clause to insert into m2m

sinisaos:
The first part of fixing duplicates in `m2m`. For now, we need to add unique constraints (using raw sql) to the m2m `joining_table` via migration or script, until `UniqueTogether` is implemented.


dantownsend:
@sinisaos Thanks for this 👍

PR #825, opened 2023-05-16
Support `DISTINCT` with `COUNT`

lqmanh:
Resolves #826.

PR #836, opened 2023-05-30
`count` query improvements - support for `distinct` and better docs

dantownsend:
Related to https://github.com/piccolo-orm/piccolo/pull/825

Remaining tasks:

- [x] Fix tests
- [x] Update docs
- [x] Write a test to make sure that passing in `distinct` and `column` raises a `ValueError`
- [x] Update docstring for `piccolo.query.methods.select.Count`


tests/table/test_count.py
```diff
@@ -1,11 +1,19 @@
...
     def test_exists(self):
         self.insert_rows()
 
-        response = Band.count().where(Band.name == "Pythonistas").run_sync()
+        response = Band.count().where(Band.popularity == 10).run_sync()
 
-        self.assertEqual(response, 1)
+        self.assertEqual(response, 2)
+
+    @engines_only("postgres", "cockroach")
```
sinisaos, line 13: You should remove this line because with your new updates `count distinct `also works in `Sqlite` :)

sinisaos (approved):
@dantownsend I think your tweaks are great because `count distinct` now also works in `Sqlite`.

dantownsend:
> @dantownsend I think your tweaks are great because count distinct now also works in Sqlite.

@sinisaos That's a good point, I didn't consider that. We can't use `DISTINCT ON` in the subquery with SQLite.

The problem with how `count` is currently implemented in Piccolo is it does a query like this:

```sql
-- Note the subquery:
SELECT count(*) FROM (SELECT * FROM some_table WHERE ...);
```

I did it this way, because it seemed like an elegant way to do it at the time with the Python code. But I think I need to refactor it, as using a subquery is less efficient. Just doing this instead:

```sql
SELECT count(*) FROM some_table WHERE ...;
```

Ideally I'd like `await MyTable.count()` to just be a `await MyTable.select(Count())`.

I need to play around with it a bit more.

piccolo/query/methods/select.py
```diff
@@ -123,19 +123,35 @@ class Count(Selectable):
...
         self._alias = alias
 
     def get_select_string(
         self, engine_type: str, with_alias: bool = True
     ) -> str:
-        if self.column is None:
-            column_name = "*"
+        expression: str
+
+        if self.distinct:
```
sinisaos, line 143: Just a question. If user pass multiple columns to `distinct` in Sqlite like this
```python
Band.count(distinct=[Band.popularity, Band.name]).run_sync()
```
Sqlite will raise `sqlite3.OperationalError: row value misused`
If we make change like this (using concatenation operator)
```python
if self.distinct:
    if engine_type == "sqlite":
        column_names = " || ".join(
            i._meta.get_full_name(with_alias=False)
            for i in self.distinct
        )
    else:
        column_names = ", ".join(
            i._meta.get_full_name(with_alias=False) for i in self.distinct
        )
    expression = f"DISTINCT ({column_names})"
```
error will be prevented and everything is fine. Can you look at that. Thanks

sinisaos:
@dantownsend This is even better, I left a comment (question) so if you can take a look at it. Thanks.

tests/table/test_count.py
dantownsend: Good point - changed.

piccolo/query/methods/select.py
dantownsend: Interesting - good idea. 👍

Issue #842, opened 2023-06-18
M2M and Joining tables that are in schemas


dantownsend:
### Discussed in https://github.com/piccolo-orm/piccolo/discussions/841

<div type='discussions-op-text'>

<sup>Originally posted by **quinnalfaro** June 17, 2023</sup>
Hey there!
Is there support for joining of two tables in the same schema with a m2m connection?
ie)

```
class Band(Table, schema="music"):
    name = Varchar()
    genres = M2M(LazyTableReference("GenreToBand", module_path=__name__))


class Genre(Table, schema="music"):
    name = Varchar()
    bands = M2M(LazyTableReference("GenreToBand", module_path=__name__))


# This is our joining table:
class GenreToBand(Table, schema="music"):
    band = ForeignKey(Band)
    genre = ForeignKey(Genre)
```</div>

PR #843, opened 2023-06-19
Fix M2M queries when in a schema

dantownsend:
Resolves https://github.com/piccolo-orm/piccolo/issues/842

Remaining tasks:

- [x] Write test

### DIFF
--- a/tests/columns/m2m/test_m2m_schema.py
+++ b/tests/columns/m2m/test_m2m_schema.py
@@ -9,6 +9,7 @@ from piccolo.columns.column_types import (
 from piccolo.columns.m2m import M2M
 from piccolo.schema import SchemaManager
 from piccolo.table import Table
+from tests.base import engines_skip
 
 from .base import M2MBase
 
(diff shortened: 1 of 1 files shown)


sinisaos (commented):
@dantownsend This is great and after the second commit everything works on both `Postgres` and `SQLite`. I find one problem when I change the schema in migrations. If we specify `schema="music"` and do the migrations, everything works, but if we remove the schema argument from the table definition and try to do the migrations, the migrations will fail with the error `schema "None" does not exist`. If I change https://github.com/piccolo-orm/piccolo/blob/cd5595c08d94dcf048f30ae68812379e9bb0d992/piccolo/apps/migrations/auto/migration_manager.py#L187 like this 
```python
def change_table_schema(
        self,
        class_name: str,
        tablename: str,
        new_schema: t.Optional[str] = None,
        old_schema: t.Optional[str] = None,
    ):
        self.change_table_schemas.append(
            ChangeTableSchema(
                class_name=class_name,
                tablename=tablename,
                new_schema=new_schema if new_schema is not None else "public",
                old_schema=old_schema if old_schema is not None else "public",
            )
        )
```

everything works.

dantownsend:
@sinisaos Good catch - thanks!

Issue #854, opened 2023-07-02
Drop Python 3.7 support


dantownsend:
Python 3.7 is now end of life - https://devguide.python.org/versions/#unsupported-versions

We should do the following:

- [ ] Update release.yaml to use a more modern Python version https://github.com/piccolo-orm/piccolo/blob/36e80f58b87822244e405736be141e68b5ae7fa4/.github/workflows/release.yaml#L17
- [ ] Update our [CI scripts](https://github.com/piccolo-orm/piccolo/blob/master/.github/workflows/tests.yaml)
  - [ ] Remove 3.7 from `python-version: ["3.7", "3.8", "3.9", "3.10", "3.11"]`
  - [ ] Update `if: matrix.python-version == '3.7'` to `if: matrix.python-version == '3.11'`.
- [ ] Update `setup.py`
  - [ ] `python_requires=">=3.7.0"` -> `python_requires=">=3.8.0"` https://github.com/piccolo-orm/piccolo/blob/36e80f58b87822244e405736be141e68b5ae7fa4/setup.py#L61
  - [ ] And drop this line: https://github.com/piccolo-orm/piccolo/blob/36e80f58b87822244e405736be141e68b5ae7fa4/setup.py#L87
- [ ] Wherever we use `from typing_extensions import Literal` or `from typing_extensions import Protocol` we can now import them directly from the `typing` module instead. `typing_extensions` can then be moved from `requirements.txt` to `test-requirements.txt`.

The reason for dropping older versions is it makes our CI run faster, and we can adopt newer Python features without having to worry about backwards compatibility.

KekmaTime:
@dantownsend hey i would like to work on this issue
can you assign me thiss

KekmaTime:
@dantownsend i have done all the tasks mentioned in the tasklist. Let me know if there are any fixes i need to make. 
Here is the link to the PR- https://github.com/piccolo-orm/piccolo/pull/855

dantownsend:
Resolved by https://github.com/piccolo-orm/piccolo/pull/855/

PR #855, opened 2023-07-02
Drop Python 3.7 support

KekmaTime:
Done all the tasks mentioned in the tasklist - https://github.com/piccolo-orm/piccolo/issues/854


KekmaTime:
@dantownsend

dantownsend:
@KekmaTime Thanks for this! Will go out in then next release.

Issue #858, opened 2023-07-06
Update Litestar asgi template


sinisaos:
`PiccoloORMPlugin` is no longer used in Litestar 2. A new feature is `PiccoloDTO` (not released yet) which I tried and works well but I found some limitations. For example `Interval` column does not work because `timedelta` is not currently supported in `msgspec`. I would simply update the Litestar asgi template with `create_pydantic_model` (which we also use in other templates) because that way everything will work well and we will be much less dependent on changes in Litestar (we could unpin dependencies from `Litestar==2.0.0a3` to the latest version). With `create_pydantic_model` we can easily solve the problem with `Interval` column because we can add custom type encoder and pass it to app constructor like this
```python
TYPE_ENCODERS = {datetime.timedelta: str}

app = Litestar(
   route_handlers=[],
   type_encoders=TYPE_ENCODERS,
)
```
That change does not work in `PiccoloDTO` due to a serialization error. @dantownsend  What do you think about that, would that be a smart thing to do? If yes, I can make those changes.

dantownsend:
@sinisaos Your suggested changes make sense. We can get the litestar template working again without having to version pin it, and then if `msgspec` supports `timedelta` in the future, and `PiccoloDTO` is released, we can switch over to using `PiccoloDTO`.

PR #859, opened 2023-07-07
update Litestar template to use create_pydantic_model

sinisaos:
Related to #858. I also pinned Pydantic to version `1.10.11` which is the last version before V2 because we are not ready for `Pydantic V2` yet.


dantownsend:
Damn, the integration tests are failing now ...

I think I need to release the latest Piccolo Admin and Piccolo API first.

Issue #862, opened 2023-07-12
Something like «makemigrations»


hoosnick:
### Discussed in https://github.com/piccolo-orm/piccolo/discussions/861

<div type='discussions-op-text'>

<sup>Originally posted by **hoosnick** July 13, 2023</sup>
How can we implement the `makemigrations` command like in [Django](https://docs.djangoproject.com/en/4.2/ref/django-admin/#django-admin-makemigrations)? 
Why is it not implemented? It is not convenient to manually create migrations for each application and then migrate.</div>

dantownsend:
This is the Piccolo equivalent:

```
piccolo migrations new my_app --auto
```

As mentioned in the discussion, we should reorder the [docs](https://piccolo-orm.readthedocs.io/en/latest/piccolo/migrations/create.html#auto-migrations), so this appears nearer the top of the page, as mention manual migrations first, which is a less common use case.

hoosnick:
@dantownsend Yes I know this command, if I have a project with 7-8 apps to create automatic migration files for all apps at one? We need such a command, which itself creates migration files in the directories related to the changes in the tables in each app.

Here's how I see it:
`piccolo migrations new all`
```
Creating migrations for 'myapp-1':
    Created tables                        2
    Dropped tables                        0
    Renamed tables                        0
    Created table columns                 5
    Dropped columns                       0
    Columns added to existing tables      0
    Renamed columns                       0
    Altered columns                       0
Creating migrations for 'myapp-2':
    Created tables                        2
    Dropped tables                        0
    Renamed tables                        0
    Created table columns                 5
    Dropped columns                       0
    Columns added to existing tables      0
    Renamed columns                       0
    Altered columns                       0
Creating migrations for 'myapp-3':
    Created tables                        2
    Dropped tables                        0
    Renamed tables                        0
    Created table columns                 5
    Dropped columns                       0
    Columns added to existing tables      0
    Renamed columns                       0
    Altered columns                       0
```
`piccolo migrations forwards all`
```
                          SESSION_AUTH
----------------------------------------------------------------
👍 1 migration already complete
🏁 No migrations need to be run

                              USER
----------------------------------------------------------------
👍 3 migrations already complete
🏁 No migrations need to be run

                         PICCOLO_ADMIN
----------------------------------------------------------------
👍 0 migrations already complete
🏁 No migrations need to be run

                             myapp-1
----------------------------------------------------------------
👍 8 migrations already complete
⏩ 1 migration not yet run
[...]

PR #864, opened 2023-07-13
862: allow `piccolo migrations new all --auto`

dantownsend:
Resolves https://github.com/piccolo-orm/piccolo/issues/862

Remaining tasks:

- [x] Add some more tests (e.g. make sure that passing in 'all' without `--auto` raises an exception)

### DIFF
--- a/piccolo/apps/migrations/commands/new.py
+++ b/piccolo/apps/migrations/commands/new.py
@@ -240,7 +240,11 @@ async def new(
         )
 
     app_names = (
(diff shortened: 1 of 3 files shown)


dantownsend:
Here's what it looks like now if we use `piccolo migrations new all --auto`:

Screenshot 2023-07-14 at 12 11 28

hoosnick (approved):
Great, better than expected!

dantownsend:
@hoosnick Cool, thanks - will merge it in and will release later today.

PR #866, opened 2023-07-16
pydantic v2 support

dantownsend:
based off https://github.com/piccolo-orm/piccolo/compare/master...sinisaos:piccolo:pydanticV2_test by @sinisaos

Related to https://github.com/piccolo-orm/piccolo_admin/issues/313


piccolo/utils/pydantic.py
```diff
@@ -284,40 +285,94 @@ def create_pydantic_model(
...
         elif isinstance(column, Text):
-            field = pydantic.Field(format="text-area", extra=extra, **params)
+            field = pydantic.Field(
+                json_schema_extra={"extra": extra, "format": "text-area"},
+                **params,
+            )
         elif isinstance(column, (JSON, JSONB)):
-            field = pydantic.Field(format="json", extra=extra, **params)
+            field = pydantic.Field(
+                json_schema_extra={"extra": extra, "format": "json"}, **params
```
sinisaos, line 307: @dantownsend Thanks for the review. Just one thing. I think you need to put `**params` kwarg inside `json_schema_extra` dict in **all fields** like im my PR because [extra kwarg is deprecated ](https://docs.pydantic.dev/latest/api/fields/#pydantic.fields.Field)and you get warnings when tests are executed. Everything else looks good to me.

piccolo/utils/pydantic.py
```diff
@@ -284,40 +285,94 @@ def create_pydantic_model(
...
+            )
         elif isinstance(column, Secret):
-            field = pydantic.Field(extra={"secret": True, **extra}, **params)
+            field = pydantic.Field(
+                json_schema_extra={"extra": {"secret": True, **extra}},
+                **params,
```
sinisaos, line 312: Same here.

piccolo/utils/pydantic.py
```diff
@@ -284,40 +285,94 @@ def create_pydantic_model(
...
+            )
+        elif isinstance(column, Interval):
+            field = pydantic.Field(
+                json_schema_extra={
+                    "extra": extra,
+                    "format": "duration",
+                },
+                **params,
```
sinisaos, line 320: And here.

piccolo/utils/pydantic.py
```diff
@@ -284,40 +285,94 @@ def create_pydantic_model(
...
+            )
+        elif isinstance(column, Time):
+            field = pydantic.Field(
+                json_schema_extra={
+                    "extra": extra,
+                    "format": "time",
+                },
+                **params,
```
sinisaos, line 336: And here.

piccolo/utils/pydantic.py
```diff
@@ -284,40 +285,94 @@ def create_pydantic_model(
...
+            )
+        elif isinstance(column, (Timestamp, Timestamptz)):
+            field = pydantic.Field(
+                json_schema_extra={
+                    "extra": extra,
+                    "format": "date-time",
+                },
+                **params,
```
sinisaos, line 344: And here.

piccolo/utils/pydantic.py
```diff
@@ -284,40 +285,94 @@ def create_pydantic_model(
...
+            )
+        elif isinstance(column, Array):
+            field = pydantic.Field(
+                json_schema_extra={
+                    "extra": extra,
+                    "type": "array",
+                },
+                **params,
```
sinisaos, line 352: And here.

piccolo/utils/pydantic.py
```diff
@@ -284,40 +285,94 @@ def create_pydantic_model(
...
+            )
+
         else:
-            field = pydantic.Field(extra=extra, **params)
+            field = pydantic.Field(
+                json_schema_extra={"extra": extra}, **params
```
sinisaos, line 357: And here.

piccolo/utils/pydantic.py
```diff
@@ -284,40 +285,94 @@ def create_pydantic_model(
...
-                    **extra,
+                json_schema_extra={
+                    "extra": {
+                        "foreign_key": True,
+                        "to": tablename,
+                        "target_column": column._foreign_key_meta.resolved_target_column._meta.name,  # noqa: E501
+                        **extra,
+                    }
                 },
                 **params,
```
sinisaos, line 296: And here

piccolo/utils/pydantic.py
```diff
@@ -284,40 +285,94 @@ def create_pydantic_model(
...
             )
             if include_readable:
                 columns[f"{column_name}_readable"] = (str, None)
         elif isinstance(column, Text):
-            field = pydantic.Field(format="text-area", extra=extra, **params)
+            field = pydantic.Field(
+                json_schema_extra={"extra": extra, "format": "text-area"},
+                **params,
```
sinisaos, line 303: And here.

sinisaos:
@dantownsend I left a few comments about `**params` kwarg. Everything else looks good to me. I'm writing from mobile phone and sorry if there are any mistakes.

docs/src/piccolo/v1/index.rst
```diff
@@ -0,0 +1,42 @@
...
+However, quite a bit of internal code in Piccolo and its sister libraries
+`Piccolo API <https://piccolo-api.readthedocs.io>`_ and
+`Piccolo Admin <https://piccolo-admin.readthedocs.io>`_ had to be changed to
+support Pydantic v2. Supporting both Pydantic v1 and Pydantic v2 would be quite
+burdensome.
+
+So Piccolo v1 will just use Pydantic v2 and above.
+
+If you can't upgrade to Pydantic v2, then pin your Piccolo version to ``0.108.0``.
+You can find the `docs here for 0.108.0 <https://piccolo-orm.readthedocs.io/en/v0.108.0/>`_.
```
sinisaos, line 36: I have one more question. Why isn't it pinned to `v0.118.0` as the last version with Pydantic v1 because there are some useful features between `v0.108.0` and `v0.118.0` that should work?

dantownsend: Good catch - that's a typo. It should be `v0.118.0`!

dantownsend: I've fixed this.

sinisaos: Great. What about `**params` outside of `json_schema_extra`? When I did that, I got a lot of deprecated warnings when running the tests until I put `**params` inside `json_schema_extra`. Can you check that if it's not a problem?

dantownsend: I'll have a look. It think what happened with Pydantic v1 is if you had this:

```python
Field(foo='bar')
```

It just added `foo` to the JSON schema. That's not allowed any more, which triggers the warnings.

But if the the argument is valid - for example `default`, then it should be OK.

https://github.com/pydantic/pydantic/blob/d249af9c7895020fee8e34f59ffb5c9220453719/pydantic/fields.py#L639C1-L640


For example:

```python
params = {'default': None'}
# This is OK
Field(**params)
# As it's equivalent to:
Field(default=None)
# And default is a permitted argument
```

I'll do some testing though to make sure.

sinisaos: Thanks. I had that problem with unpacked params outside `json_schema_extra`. I'd check it out, but I don't have a laptop.

sinisaos: You're right. I had issues with deprecated warnings because I left a `nullable` field in params kwarg, which is not allowed in Pydantic V2. You moved the `nullable` field to the `extra` kwarg located inside the `json_schema_extra` and everything is fine. I'm sorry I didn't notice it before, but check it out anyway.

dantownsend: Ah, cool - thanks for checking. Will run the tests once more to check everything is ok, then wil merge in.

sinisaos:
@dantownsend I found one more line to change and then there are no more deprecated warnings.
https://github.com/piccolo-orm/piccolo/blob/b13c383247532352752828bb41aa8ec830a0dd67/piccolo/apps/fixtures/commands/load.py#L47

should be:
```python
fixture_pydantic_model = pydantic_model_class.model_validate_json(json_string)
```

sinisaos:
> @dantownsend I found one more line to change and then there are no more deprecated warnings.
> 
> https://github.com/piccolo-orm/piccolo/blob/b13c383247532352752828bb41aa8ec830a0dd67/piccolo/apps/fixtures/commands/load.py#L47
> 
> should be:
> 
> ```python
> fixture_pydantic_model = pydantic_model_class.model_validate_json(json_string)
> ```

@dantownsend Could you please, when you have time, change the above line to fix all the deprecated warnings, then merge this PR and release a new version. After that we can also add Pydantic v2 to Piccolo API and Piccolo Admin (PRs already exists). Thanks.

dantownsend:
@sinisaos Thanks - I've changed that now.

PR #881, opened 2023-08-23
feat: bring password hashing inline with industry best practices

Skelmis:
This pull request accomplishes two things:
- Brings Piccolo's internal password hashing inline with industry best practices. See [here](https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html#pbkdf2) for more information
- Implements the functionality required to automatically migrate hashes to the newest iteration count upon login

This pull request does not propose a solution to the storage of legacy hashes

### DIFF
--- a/piccolo/apps/user/tables.py
+++ b/piccolo/apps/user/tables.py
@@ -49,6 +49,7 @@ class BaseUser(Table, tablename="piccolo_user"):
 
     _min_password_length = 6
     _max_password_length = 128
+    _pbkdf2_iteration_count = 600_000
 
     def __init__(self, **kwargs):
         # Generating passwords upfront is expensive, so might need reworking.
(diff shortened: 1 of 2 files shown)


dantownsend:
@Skelmis Thanks a lot for this - you're right, we should increase the iterations. We just need to be careful to make it backwards compatible.

Skelmis:
Heya @dantownsend, backwards compatibility is key I do agree. This pull request handles that under the assumption that piccolo only supports `pbkdf2_sha256` as the password hashing mechanism, an assumption that is true at the time of this pr.

Essentially when a user attempts to log in, the users current hash is introspected to get the required iteration count (`10_000` in the current case). 
The two hashes are then compared and if a user is authenticated, the users current iteration count is compared to `BaseUser._pbkdf2_iteration_count` which denotes the current iteration count. If the two values differ, Piccolo knows the password hash requires upgrading and performs this operation using the now authenticated password provided by the end user.

I have performed manual testing locally and received the expected hash upgrade when logging in as a user with a hash set with `10_000` iterations

dantownsend:
@Skelmis Thanks again for this. I've gone over it a few times, and done some testing, and it all seems solid to me 👍 

I wrote a unit test to make sure nothing breaks in the future. I'm happy to merge this in, and put it in the next release.

PR #886, opened 2023-09-15
Support Postgres 16

dantownsend:
Running tests for Postgres 16 in CI.


dantownsend:
It looks like the Postgres 16 Docker container isn't available yet, so the tests are failing - need to retry later.

PR #888, opened 2023-10-04
add Python 3.12 support

dantownsend:
Making sure the tests run for Python 3.12, and it's shown in PyPI.


dantownsend:
The CI is failing because of slotscheck - Python 3.12 has changed generics.

dantownsend:
Asked about slotscheck with Python 3.12 generics here: https://github.com/ariebovenberg/slotscheck/issues/201

For now, will skip it for Python 3.12 until I can figure out what to do.

PR #906, opened 2023-11-30
Pylance fixes

dantownsend:
Builds on the changes in PR https://github.com/piccolo-orm/piccolo/pull/905

Similarly to Mypy, Pylance has become much stricter / more accurate over time. There are now quite a few minor type errors being reported by Pylance.

As VSCode is such a popular code editor it would be good to clean up most (if not all) of these minor errors.

I'm not sure how easy it is to add Pyright (which powers Pylance) as a linter to our CI, but will investigate.


dantownsend:
This PR is insanely big - I'm going to break it down into smaller chunks, and merge those in (starting with any changes to test files).

PR #907, opened 2023-12-01
Add support for Esmerald

tarsil:
- Adding support for [Esmerald](https://esmerald.dev) with Piccolo.

### DIFF
--- a/piccolo/apps/asgi/commands/templates/app/_esmerald_app.py.jinja
+++ b/piccolo/apps/asgi/commands/templates/app/_esmerald_app.py.jinja
@@ -1,0 +1,111 @@
+import typing as t
+
+from pathlib import Path
+
+from piccolo_admin.endpoints import create_admin
+from piccolo_api.crud.serializers import create_pydantic_model
+from piccolo.engine import engine_finder
+
+from esmerald import (
+    Esmerald,
+    Include,
+    Gateway,
+    JSONResponse,
+    APIView,
+    get,
+    post,
+    put,
+    delete
+)
+from esmerald.config import StaticFilesConfig
+
+from home.endpoints import home
+from home.piccolo_app import APP_CONFIG
+from home.tables import Task
+
+
+async def open_database_connection_pool():
+    try:
+        engine = engine_finder()
+        await engine.start_connection_pool()
+    except Exception:
+        print("Unable to connect to the database")
+
+
+async def close_database_connection_pool():
+    try:
+        engine = engine_finder()
+        await engine.close_connection_pool()
+    except Exception:
+        print("Unable to connect to the database")
+
+
+TaskModelIn: t.Any = create_pydantic_model(
+    table=Task,
+    model_name='TaskModelIn'
+)
+TaskModelOut: t.Any = create_pydantic_model(
+    table=Task,
+    include_default_columns=True,
+    model_name='TaskModelOut'
+)
+
+
+class TaskAPIView(APIView):
+    path: str = "/"
+    tags: str = ["Task"]
+
+    @get("/")
+    async def tasks() -> t.List[TaskModelOut]:
+        return await Task.select().order_by(Task.id)
+
+
+    @post('/tasks')
+    async def create_task(payload: TaskModelIn) -> TaskModelOut:
+        task = Task(**payload.dict())
+        await task.save()
+        return task.to_dict()
+
+
+    @put('/tasks/{task_id}')
+    async def update_task(payload: TaskModelIn, task_id: int) -> TaskModelOut:
+        task = await Task.objects().get(Task.id == task_id)
+        if not task:
+            return JSONResponse({}, status_code=404)
+
+        for key, value in payload.dict().items():
+            setattr(task, key, value)
+
+        await task.save()
+
+        return task.to_dict()
+
+
+    @delete('/tasks/{task_id}')
+    async def delete_task(task_id: int) -> JSONResponse:
+        task = await Task.objects().get(Task.id == task_id)
+        if not task:
+            return JSONResponse({}, status_code=404)
+
+        await task.remove()
+
+        return JSONResponse({})
+
+
+app = Esmerald(
+    routes=[
+        Gateway("/", handler=home),
+        Gateway("/tasks", handler=TaskAPIView),
+        Include(
+            "/admin/",
+            create_admin(
+                tables=APP_CONFIG.table_classes,
+                # Required when running under HTTPS:
+                # allowed_hosts=['my_site.com']
+            ),
+        ),
+    ],
+    static_files_config=StaticFilesConfig(path="/", directory=Path("static")),
+    on_startup=[open_database_connection_pool],
+    on_shutdown=[close_database_connection_pool],
+)
--- a/piccolo/apps/asgi/commands/templates/app/home/_esmerald_endpoints.py.jinja
+++ b/piccolo/apps/asgi/commands/templates/app/home/_esmerald_endpoints.py.jinja
@@ -1,0 +1,21 @@
+import os
+
+import jinja2
+from esmerald import MediaType, Request, Response, get
+from esmerald.responses import HTMLResponse
+
+ENVIRONMENT = jinja2.Environment(
+    loader=jinja2.FileSystemLoader(
+        searchpath=os.path.join(os.path.dirname(__file__), "templates")
+    )
+)
+
+
+@get(path="/", include_in_schema=False)
+def home(request: Request) -> HTMLResponse:
+    template = ENVIRONMENT.get_template("home.html.jinja")
+
+    content = template.render(title="Piccolo + ASGI",)
+
+    return HTMLResponse(content)
+
--- a/docs/src/index.rst
+++ b/docs/src/index.rst
@@ -59,8 +59,8 @@ Give me an ASGI web app!
 
    piccolo asgi new
 
-FastAPI, Starlette, BlackSheep, and Litestar are currently supported, with more
-coming soon.
+FastAPI, Starlette, BlackSheep, Litestar and Esmerald are currently supported,
+with more coming soon.
 
 -------------------------------------------------------------------------------
 
--- a/docs/src/piccolo/asgi/index.rst
+++ b/docs/src/piccolo/asgi/index.rst
@@ -21,8 +21,8 @@ Routing frameworks
 ******************
 
 Currently, `Starlette <https://www.starlette.io/>`_, `FastAPI <https://fastapi.tiangolo.com/>`_,
-`BlackSheep <https://www.neoteroi.dev/blacksheep/>`_, and
-`Litestar <https://litestar.dev/>`_ are supported.
+`BlackSheep <https://www.neoteroi.dev/blacksheep/>`_,
+`Litestar <https://litestar.dev/>`_ and `Esmerald <https://esmerald.dev/>`_ are supported.
 
 Other great ASGI routing frameworks exist, and may be supported in the future
 (`Quart <https://pgjones.gitlab.io/quart/>`_ ,
@@ -32,8 +32,8 @@ Other great ASGI routing frameworks exist, and may be supported in the future
 Which to use?
 =============
 
-All are great choices. FastAPI is built on top of Starlette, so they're
-very similar. FastAPI and BlackSheep are great if you want to document a REST
+All are great choices. FastAPI and Esmerald are built on top of Starlette, so they're
+very similar. FastAPI, BlackSheep and Esmerald are great if you want to document a REST
 API, as they have built-in OpenAPI support.
 
 -------------------------------------------------------------------------------
--- a/piccolo/apps/asgi/commands/templates/app/home/endpoints.py.jinja
+++ b/piccolo/apps/asgi/commands/templates/app/home/endpoints.py.jinja
@@ -4,4 +4,6 @@
     {% include '_blacksheep_endpoints.py.jinja' %}
 {% elif router == 'litestar' %}
     {% include '_litestar_endpoints.py.jinja' %}
+{% elif router == 'esmerald' %}
+    {% include '_esmerald_endpoints.py.jinja' %}
 {% endif %}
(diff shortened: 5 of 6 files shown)


dantownsend:
@tarsil Thanks for this.

If you add `esmerald` to this list:

https://github.com/piccolo-orm/piccolo/blob/b8026da68112686952a10ff2db3f37f4d0fdf1bd/piccolo/apps/asgi/commands/new.py#L13

Then when someone uses `piccolo asgi new` it will show `esmerald` as an option.

If you need to lock `esmerald` to a certain version, you can do so using this:

https://github.com/piccolo-orm/piccolo/blob/b8026da68112686952a10ff2db3f37f4d0fdf1bd/piccolo/apps/asgi/commands/new.py#L14-L16

We run integration tests against all of the ASGI frameworks that Piccolo supports:

https://github.com/piccolo-orm/piccolo/blob/master/tests/apps/asgi/commands/test_new.py

We basically test that the generated Python code is valid, and we run a really simplistic ASGI server to make sure the app runs.

tarsil:
Hi @dantownsend thank you for your message. I will have a look at all of this in the next days (not in my machine at the moment) but I was also wondering where I could do that test you mentioned. Great to know. Great ORM btw 🙂

dantownsend:
@tarsil It looks like there's a dependency issue. Piccolo Admin has FastAPI as a dependency, which itself depends on Starlette. FastAPI has very narrowly defined dependency versions which means it sometimes doesn't play well with other libraries which depend on Starlette etc. Which version of Starlette does `esmerald` need?

tarsil:
@dantownsend  I see. I know that limitation from Fast. 

Esmerald uses the version 0.32.0.post1. Besides the lastest released today basically

tarsil:
@dantownsend let me know your thoughts how to go from here then if you don't mind

dantownsend:
@tarsil I think the problem is that  `esmerald` needs `starlette==0.32.0`:

https://github.com/dymmond/esmerald/blob/34cb7cc77f42b40c7a9ef8566dc659439be750d6/pyproject.toml#L58C10-L58C10

But the latest FastAPI only supports `starlette==0.28.0`:

https://github.com/tiangolo/fastapi/blob/8cf2fa0fe4f8701411db779176cc4cfac68d4216/pyproject.toml#L43

So when trying to `pip install` all the dependencies, it's using really old versions of Piccolo and FastAPI (the older versions specify their dependencies looser, so are considered OK by `pip`, but actually don't run correctly together).

I think the only solutions are:

* `esmerald` supporting a broader range of versions for `starlette`
* `fastapi` releasing a new version which supports a recent version of `starlette` (it will happen eventually, but we're not in control of when)
* Remove `fastapi` from Piccolo Admin, but that would be quite a big refactor.
* Don't use Piccolo Admin in the `esmerald` template - which removes `fastapi` as a dependency

It's a shame that FastAPI is like this, because it makes it makes it harder for libraries based on `starlette` to interoperate.

tarsil:
@dantownsend thank you for this detailed explanation. I will be releasing a version soon and I can lose some tight versions and put it in a range. This way it will allow the same version of Starlette to be installed 🙂

tarsil:
@dantownsend I did a new release today and made the requirements more broad, from >=0.28.0,<=0.32.0.post1. I think this should now do the trick?

dantownsend:
@tarsil Thanks, I'll re-run the CI now.

tarsil:
@dantownsend is still the same issue happening?

dantownsend:
@tarsil Yeah, it's another dependency clash.

If you make a virtualenv, and make this `requirements.txt` file:

```
esmerald==2.4.3
piccolo[postgres]>1.0.0
piccolo_admin>1.0.0
```

`pip install -r requirements.txt` can't resolve it - I think the problem now is Pydantic being too tightly defined.

Screenshot 2023-12-08 at 20 26 54

piccolo/apps/asgi/commands/templates/app/_esmerald_app.py.jinja
```diff
@@ -0,0 +1,111 @@
...
+
+    @delete('/tasks/{task_id}')
+    async def delete_task(self, task_id: int) -> JSONResponse:
+        task = await Task.objects().get(Task.id == task_id)
+        if not task:
+            return JSONResponse({}, status_code=404)
+
+        await task.remove()
+
+        return JSONResponse({})
```
sinisaos, line 92:
```suggestion
    async def delete_task(self, task_id: int) -> None:
        task = await Task.objects().get(Task.id == task_id)
        if not task:
            return JSONResponse({}, status_code=404)

        await task.remove()
```
Fix `esmerald.exceptions.ImproperlyConfigured: 500: A status code 204, 304 or in the range below 200 does not support a response body`

piccolo/apps/asgi/commands/templates/app/home/_esmerald_endpoints.py.jinja
```diff
@@ -0,0 +1,21 @@
+import os
+
+import jinja2
+from esmerald import MediaType, Request, Response, get
```
sinisaos, line 4:
```suggestion
from esmerald import Request, Response, get
```
Fix `ImportError: cannot import name 'MediaType' from 'esmerald'`

tarsil:
@sinisaos thank you very much for the suggestions above. Odd as my local version has something like you suggested. Maybe I actually didn't push what I needed before but either way, this is great. Thank you so much

tarsil:
@sinisaos i applied your suggestions. I confirm that in my local I actually had similar but not pushed since the MediaType is actually inside the enums but we actually don't use it here. The last suggestion, if I'm not mistaken it was already in the PR but I appreciate that. @dantownsend i think with these changes suggested by @sinisaos , we can try again 🙂

sinisaos:
@tarsil No problem. I'm always glad if I can help somehow. As for Piccolo Admin, unfortunately, I think we have to wait for the FastAPI upgrade, which is not always very fast.

> The last suggestion, if I'm not mistaken it was already in the PR but I appreciate that. 

You're right. I wrote it wrong (but I thought), you have to add Esmerald [here as well](https://github.com/piccolo-orm/piccolo/blob/master/piccolo/apps/asgi/commands/templates/app/app.py.jinja)

tarsil:
> @tarsil No problem. I'm always glad if I can help somehow. As for Piccolo Admin, unfortunately, I think we have to wait for the FastAPI upgrade, which is not always very fast.
> 
> > The last suggestion, if I'm not mistaken it was already in the PR but I appreciate that.
> 
> You're right. I wrote it wrong (but I thought), you have to add Esmerald [here as well](https://github.com/piccolo-orm/piccolo/blob/master/piccolo/apps/asgi/commands/templates/app/app.py.jinja)
> 
> Also remove the Piccolo Admin [import](https://github.com/piccolo-orm/piccolo/pull/907/files#diff-116f4ee0a2adeefb5c3c31cfb5d4e6833d32816e8af3e2271f61d180909902a1R5) and `Include` the Piccolo Admin in the routes in [here](https://github.com/piccolo-orm/piccolo/pull/907/files#diff-116f4ee0a2adeefb5c3c31cfb5d4e6833d32816e8af3e2271f61d180909902a1R97-R104), otherwise it won't work and the integration test will fail.

Will definitely do @sinisaos thank you once again

tarsil:
@sinisaos I'm actually still having some issues locally. How it does look like in your final `esmerald` template? Maybe I missed something, if you don't mind sharing.

sinisaos:
@tarsil You also need to add Emserald [here](https://github.com/piccolo-orm/piccolo/blob/master/piccolo/apps/asgi/commands/templates/app/app.py.jinja) like this 
```python
{% if router == 'fastapi' %}
    {% include '_fastapi_app.py.jinja' %}
{% elif router == 'starlette' %}
    {% include '_starlette_app.py.jinja' %}
{% elif router == 'blacksheep' %}
    {% include '_blacksheep_app.py.jinja' %}
{% elif router == 'litestar' %}
    {% include '_litestar_app.py.jinja' %}
{% elif router == 'esmerald' %}
    {% include '_esmerald_app.py.jinja' %}
{% endif %}
```
to add `app.py` for Esmerald template. After that run `./scripts/test-integration.sh`

tarsil:
> @tarsil You also need to add Emserald [here](https://github.com/piccolo-orm/piccolo/blob/master/piccolo/apps/asgi/commands/templates/app/app.py.jinja) like this
> 
> ```python
> {% if router == 'fastapi' %}
>     {% include '_fastapi_app.py.jinja' %}
> {% elif router == 'starlette' %}
>     {% include '_starlette_app.py.jinja' %}
> {% elif router == 'blacksheep' %}
>     {% include '_blacksheep_app.py.jinja' %}
> {% elif router == 'litestar' %}
>     {% include '_litestar_app.py.jinja' %}
> {% elif router == 'esmerald' %}
>     {% include '_esmerald_app.py.jinja' %}
> {% endif %}
> ```
> 
> to add `app.py` for Esmerald template. After that run `./scripts/test-integration.sh`

Ahhhh, yes, totally missed that one and yet you mentioned but I was mistaken with another file. Done! If we could run the CI here to see, that would be great @sinisaos @dantownsend .

Hopefully now we can merge this and once FastAPI moves some of the versions up, we can add the admin as well. Thank you very much for your help @sinisaos , you are a ⭐ . I used Piccolo with Esmerald a few versions ago (of Esmerald) and I simply loved it.

sinisaos:
@tarsil I'm glad you made it :smiley: . For re-run in CI you must wait for @dantownsend. Just two more small thing. 
First, you must replace 
```python
static_files_config=StaticFilesConfig(path="/", directory=Path("static")),
```
to 

```python
static_files_config=StaticFilesConfig(path="/static", directory=Path("static")),
```
in Esmerald application class, to make static files work.
The other thing is that you should shorten the path in the `TaskAPIView` methods to something like this `@put("/{task_id}")`,
because `/tasks/tasks/1` looks a little weird. 
Thanks.

tarsil:
> @tarsil I'm glad you made it 😃 . For re-run in CI you must wait for @dantownsend. Just two more small thing. First, you must replace
> 
> ```python
> static_files_config=StaticFilesConfig(path="/", directory=Path("static")),
> ```
> 
> to
> 
> ```python
> static_files_config=StaticFilesConfig(path="/static", directory=Path("static")),
> ```
> 
> in Esmerald application class, to make static files work. The other thing is that you should shorten the path in the `TaskAPIView` methods to something like this `@put("/{task_id}")`, because `/tasks/tasks/1` looks a little weird. Thanks.

Oh, true, I was doing an experiment in the static files config and forgot to add back the /static, nice catch (been doing so many things that I miss a lot sometimes). 

The tasks path should now be correct. Nice suggestions @sinisaos .

I'm actually happily impressed how clear you are with Esmerald 😃

sinisaos:
@tarsil Great. Thanks for the kind words.

tarsil:
> @tarsil Great. Thanks for the kind words.

I'm the one who needs to thank you. Without your help I would take longer to get this done 🙂. 

I hope you like Esmerald to be honest 🙂

dantownsend:
It looks like the tests are passing. @sinisaos thanks for your help with this.

So it looks like Piccolo Admin was omitted to get around the dependency issues, which I think makes sense.

But as Piccolo Admin is in `requirements.txt.jinja`:

https://github.com/tarsil/piccolo/blob/master/piccolo/apps/asgi/commands/templates/app/requirements.txt.jinja

It still gets installed. I checked the logs for the integration tests, and it installed a really old version of Piccolo Admin. As `esmerald` isn't using Piccolo Admin, it no longer throws an error when running the `esmerald` app. However, I think it's just better to remove `piccolo_admin` from `requirements.txt.jinja`.

We could do something simple like:

```python
{% if 'esmerald' not in router_dependencies %}
piccolo_admin
{% endif }
```

There might be a cleaner way of doing this. In here instead:

https://github.com/tarsil/piccolo/blob/b8026da68112686952a10ff2db3f37f4d0fdf1bd/piccolo/apps/asgi/commands/new.py#L13-L16

It could be:

```python
ROUTERS = ["starlette", "fastapi", "blacksheep", "litestar"]
ROUTER_DEPENDENCIES = {
    "starlette": ["starlette", "piccolo_admin"],
    "fastapi": ["fastapi>=0.100.0", "piccolo_admin"],
    "blacksheep": ["blacksheep", "piccolo_admin"],
    "litestar": ["litestar", "piccolo_admin"],
    "esmerald": ["esmerald"],
}
```

I noticed that `requirements.txt.jinja` has `jinja2` and `requests` listed.

https://github.com/tarsil/piccolo/blob/b8026da68112686952a10ff2db3f37f4d0fdf1bd/piccolo/apps/asgi/commands/templates/app/requirements.txt.jinja#L5-L8

We might be able to remove both of these. `jinja2` is already a requirement of `piccolo`, and I think `requests` was in there for an older version of `starlette`.

tarsil:
@dantownsend based what you have said, which makes sense, what do you say you create a code change suggestion and we can simply merge it? I'm sure you are more familiar with this procedure than I am and this way we can move forward?

sinisaos:
@dantownsend Unfortunately this does not run the tests for `Esmerald`
> It could be:
> 
> ```python
> ROUTERS = ["starlette", "fastapi", "blacksheep", "litestar"]
> ROUTER_DEPENDENCIES = {
>     "starlette": ["starlette", "piccolo_admin"],
>     "fastapi": ["fastapi>=0.100.0", "piccolo_admin"],
>     "blacksheep": ["blacksheep", "piccolo_admin"],
>     "litestar": ["litestar", "piccolo_admin"],
>     "esmerald": ["esmerald"],
> }
> ```

I think this is much better:
> ```python
> {% if 'esmerald' not in router_dependencies %}
> piccolo_admin
> {% endif %}
> ```

so `requirements.txt.jinja` should look something like this

```python
{%- for router_dependency in router_dependencies -%}
{{ router_dependency }}
{% endfor -%}
{{ server }}
piccolo[postgres]
{% if not 'esmerald' in router_dependencies %}
piccolo_admin
{% endif %}
```
This runs the tests but does not install Piccolo Admin for Esmerald.

tarsil:
If @dantownsend is happy with that, @sinisaos you can add as suggestion for change as well

sinisaos:
@tarsil You'll have to wait for @dantownsend . If that's good enough, you or @dantownsend need to add it because `requirements.txt.jinja` is not in your PR and I can't add suggestions.

tarsil:
@sinisaos of course! Thank you, if @dantownsend agrees, I can add it to the PR.

dantownsend:
@tarsil @sinisaos I'll merge this in now - will make the changes to `requirements.txt` in another PR. Thanks both 👍

PR #911, opened 2023-12-12
Update ASGI requirements

dantownsend:
Related to https://github.com/piccolo-orm/piccolo/pull/907

* Removing `jinja2` as it's already a direct dependency of Piccolo
* Removing `requests` as I don't think it's needed any more (I think this was for Starlette at some point)
* Not installing `piccolo_admin` for `esmerald`

### DIFF
--- a/piccolo/apps/asgi/commands/new.py
+++ b/piccolo/apps/asgi/commands/new.py
@@ -10,15 +10,11 @@ from jinja2 import Environment, FileSystemLoader
 
 TEMPLATE_DIR = os.path.join(os.path.dirname(__file__), "templates/app/")
 SERVERS = ["uvicorn", "Hypercorn"]
-ADMIN_DEPENDENCIES = ["piccolo_admin>=1.0.0"]
 ROUTER_DEPENDENCIES = {
-    "starlette": ["starlette", *ADMIN_DEPENDENCIES],
-    "fastapi": ["fastapi>=0.100.0", *ADMIN_DEPENDENCIES],
-    "blacksheep": ["blacksheep", *ADMIN_DEPENDENCIES],
-    "litestar": ["litestar", *ADMIN_DEPENDENCIES],
-    # We don't currently use Piccolo Admin with esmerald, due to dependency
-    # conflicts with FastAPI. We'll revisit this at some point in the future
-    # when newer versions of FastAPI are released (DT - 12th December 2023).
+    "starlette": ["starlette"],
+    "fastapi": ["fastapi>=0.100.0"],
+    "blacksheep": ["blacksheep"],
+    "litestar": ["litestar"],
     "esmerald": ["esmerald"],
 }
 ROUTERS = list(ROUTER_DEPENDENCIES.keys())
--- a/piccolo/apps/asgi/commands/new.py
+++ b/piccolo/apps/asgi/commands/new.py
@@ -10,10 +10,18 @@ from jinja2 import Environment, FileSystemLoader
 
 TEMPLATE_DIR = os.path.join(os.path.dirname(__file__), "templates/app/")
 SERVERS = ["uvicorn", "Hypercorn"]
-ROUTERS = ["starlette", "fastapi", "blacksheep", "litestar", "esmerald"]
+ADMIN_DEPENDENCIES = ["piccolo_admin>=1.0.0"]
 ROUTER_DEPENDENCIES = {
-    "fastapi": ["fastapi>=0.100.0"],
+    "starlette": ["starlette", *ADMIN_DEPENDENCIES],
+    "fastapi": ["fastapi>=0.100.0", *ADMIN_DEPENDENCIES],
+    "blacksheep": ["blacksheep", *ADMIN_DEPENDENCIES],
+    "litestar": ["litestar", *ADMIN_DEPENDENCIES],
+    # We don't currently use Piccolo Admin with esmerald, due to dependency
+    # conflicts with FastAPI. We'll revisit this at some point in the future
+    # when newer versions of FastAPI are released (DT - 12th December 2023).
+    "esmerald": ["esmerald"],
 }
+ROUTERS = list(ROUTER_DEPENDENCIES.keys())
 
 
 def print_instruction(message: str):
--- a/piccolo/apps/asgi/commands/templates/app/requirements.txt.jinja
+++ b/piccolo/apps/asgi/commands/templates/app/requirements.txt.jinja
@@ -2,5 +2,4 @@
 {{ router_dependency }}
 {% endfor -%}
 {{ server }}
-piccolo[postgres]
-piccolo_admin
+piccolo[postgres]>=1.0.0
--- a/piccolo/apps/asgi/commands/templates/app/requirements.txt.jinja
+++ b/piccolo/apps/asgi/commands/templates/app/requirements.txt.jinja
@@ -3,3 +3,4 @@
 {% endfor -%}
 {{ server }}
 piccolo[postgres]>=1.0.0
+piccolo_admin>=1.0.0
--- a/piccolo/apps/asgi/commands/templates/app/requirements.txt.jinja
+++ b/piccolo/apps/asgi/commands/templates/app/requirements.txt.jinja
@@ -2,7 +2,6 @@
 {{ router_dependency }}
 {% endfor -%}
 {{ server }}
-jinja2
 piccolo[postgres]
 piccolo_admin
 requests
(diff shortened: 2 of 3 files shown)


piccolo/apps/asgi/commands/new.py
```diff
@@ -10,10 +10,18 @@
...
-    "fastapi": ["fastapi>=0.100.0"],
+    "starlette": ["starlette", *ADMIN_DEPENDENCIES],
+    "fastapi": ["fastapi>=0.100.0", *ADMIN_DEPENDENCIES],
+    "blacksheep": ["blacksheep", *ADMIN_DEPENDENCIES],
+    "litestar": ["litestar", *ADMIN_DEPENDENCIES],
+    # We don't currently use Piccolo Admin with esmerald, due to dependency
+    # conflicts with FastAPI. We'll revisit this at some point in the future
+    # when newer versions of FastAPI are released (DT - 12th December 2023).
+    "esmerald": ["esmerald"],
 }
```
dantownsend, line 23: I decided to omit `piccolo_admin` from `esmerald` this way instead of having this in `requirements.txt`:

```jinja
{% if "esmerald" not in router_dependencies %}
piccolo_admin
{% endif %}
```

Because at some point I'd like to version pin the routers a bit more. So it might be that we have "esmerald>X.X.X" in which case the `if` statement would fail.

piccolo/apps/asgi/commands/templates/app/requirements.txt.jinja
```diff
@@ -2,7 +2,4 @@
 {{ router_dependency }}
 {% endfor -%}
 {{ server }}
-jinja2
-piccolo[postgres]
-piccolo_admin
-requests
+piccolo[postgres]>=1.0.0
```
dantownsend, line 5: Not sure if this will cause any problems ... i.e. it means that all of routers must support Pydantic v2. Will see when the CI runs.

sinisaos:
@dantownsend This should work because all router frameworks, used by Piccolo, support Pydantic v2, but I want to try each router template to be sure. I'll do it tonight because I can't right now, if you're okay with that?

dantownsend:
@sinisaos No worries - thanks!

sinisaos (commented):
The changes from this PR are good and don't affect how the Piccolo asgi template works, but I found two things that don't work. First is that Litestar Swagger docs don't work, but that's not on our side. Fix for that is made in Litestar main branch but not released yet (when they release new version everything will be fine - [see issue](https://github.com/litestar-org/litestar/issues/2883)). Second thing is for Esmerald. We have to remove `"piccolo_admin.piccolo_app"` only for Esmerald template from here https://github.com/piccolo-orm/piccolo/blob/4fb169d81551c122f353f1e9018974f5f077b3bd/piccolo/apps/asgi/commands/templates/app/piccolo_conf.py.jinja#L17 because we don't use Piccolo Admin in Esmerald template at the moment. If we don't remove Piccolo Admin from `APP_REGISTRY`  `Unable to connect to database` exception is raised and we can't run migrations for `home` app. We can use something like this.
```python
{% if router == 'esmerald' %}
APP_REGISTRY = AppRegistry(
    apps=["home.piccolo_app"]
)
{% else %}
APP_REGISTRY = AppRegistry(
    apps=["home.piccolo_app", "piccolo_admin.piccolo_app"]
)
{% endif %}
```
It is also necessary to somehow notify the user that they cannot currently run `session_auth` and `user` migrations for Esmerald (dependency conflicts with FastAPI)

dantownsend:
@sinisaos Good points. Let's not worry about `litestar`, as you say - it'll be fixed when they release the next version.

With `esmerald`, omitting Piccolo Admin is definitely trickier than I imagined. I can add the `if` block to `piccolo_conf.py.jinja` like you suggest. And maybe just put something in the docs about support being beta at the moment.

sinisaos:
> With `esmerald`, omitting Piccolo Admin is definitely trickier than I imagined. I can add the `if` block to `piccolo_conf.py.jinja` like you suggest. And maybe just put something in the docs about support being beta at the moment.

That would be great.

I found another small issue with Blacksheep and Pydantic v2. If we create a Pydantic model with `BaseModel` everything works, but when we create a Piccolo model with `create_pydantic_model`, the request body (and `example value`) is empty like this

empty_request_body

I think it's not a Piccolo problem, but a problem with the old version of SwaggerUI in [BlackSheep](https://github.com/Neoteroi/BlackSheep/blob/5f9bc45f199bc10f2274fda91414fb5168eb14e1/blacksheep/server/res/swagger-ui.html#L6-L10). If I change `swagger-ui-dist@3.30.0` for CSS and JS to `swagger-ui-dist@5` everything works fine like this

not_empty_request_body

I know you have been in contact with the author of BlackSheep for mounting the ASGI apps to BlackSheep (for Piccolo Admin). If you could raise issue in their repo, referring to this issue, that would be great.

sinisaos:
Litestar problem is solved with newest version (**v2.4.4**)

sinisaos:
BlackSheep issue will also be resolved with the next BlackSheep release (SwaggerUI has been upgraded in the BlackSheep `main` branch)

dantownsend:
@sinisaos Great, thanks a lot for checking that 👍

sinisaos:
@dantownsend FastAPI has been upgraded ([Starlette >=0.29.0,<0.33.0](https://github.com/tiangolo/fastapi/releases/tag/0.108.0)) and we can now add Piccolo Admin to the Esmerald template. The integration test passed, but when I try to make a new Esmerald project in a fresh virtualenv, I had problems pip installing [awesome-slugify](https://github.com/dymmond/esmerald/blob/eabf3d158c6319c92e5b5f9dbad4b7063c2ebf32/pyproject.toml#L45) (an Esmerald dependency) @tarsil, can you check it?

tarsil:
> @dantownsend FastAPI has been upgraded ([Starlette >=0.29.0,<0.33.0](https://github.com/tiangolo/fastapi/releases/tag/0.108.0)) and we can now add Piccolo Admin to the Esmerald template. The integration test passed, but when I try to make a new Esmerald project in a fresh virtualenv, I had problems pip installing [awesome-slugify](https://github.com/dymmond/esmerald/blob/eabf3d158c6319c92e5b5f9dbad4b7063c2ebf32/pyproject.toml#L45) (an Esmerald dependency) @tarsil, can you check it?

Thank you @sinisaos I created actually some venvs in the past with all python versions and no issues were found. What error are you receiving?

sinisaos:
@tarsil I'm not at my machine right now, but when I can, I'll post the error log here. I tried Py3.9, Py3.10, Py3.11 and always got the same error (can't install `awesome-slugify`) when I try to `pip install esmerald`  in fresh virualenv.

tarsil:
> @tarsil I'm not at my machine right now, but when I can, I'll post the error log here. I tried Py3.9, Py3.10, Py3.11 and always got the same error (can't install `awesome-slugify`) when I try to `pip install esmerald` in fresh virualenv.

 Also away for a couple of days and then I will be back at my machine but it is a first like this

sinisaos:
@tarsil I'm very sorry, but the problem was my machine (very strange, but somehow setuptools was not installed when I created a new virtualenv and that caused problems with installing Esmerald). After restarting the laptop, everything is fine. 
@dantownsend I can confirm that you can add Piccolo Admin support to Esmerald template, everything works fine.

tarsil:
> @tarsil I'm very sorry, but the problem was my machine (very strange, but somehow setuptools was not installed when I created a new virtualenv and that caused problems with installing Esmerald). After restarting the laptop, everything is fine. @dantownsend I can confirm that you can add Piccolo Admin support to Esmerald template, everything works fine.

Thank you @sinisaos for confirming. This was indeed odd but it could have happened

dantownsend:
@sinisaos Cool, thanks for checking. It'll be much easier to not exclude Piccolo Admin.

sinisaos:
@dantownsend Now that FastAPI is upgraded there is no need to exclude Piccolo Admin. Just need to add ADMIN_DEPENDENCIES to `esmerald` and add `create_admin` to Esmerald application class like this 
```python
# in _esmerald_app.py.jinja
from piccolo_admin.endpoints import create_admin
...
app = Esmerald(
    routes=[
        Gateway("/", handler=home),
        Gateway("/tasks", handler=TaskAPIView),
        Include(
            "/admin/",
            create_admin(
                tables=APP_CONFIG.table_classes,
                # Required when running under HTTPS:
                # allowed_hosts=['my_site.com']
            ),
        ),
    ],
    static_files_config=StaticFilesConfig(path="/static", directory=Path("static")),
    on_startup=[open_database_connection_pool],
    on_shutdown=[close_database_connection_pool],
)
```

and we can use Piccolo Admin in Esmerald template.

PR #921, opened 2024-01-22
add "piccolo user list" command

waldner:


piccolo/apps/user/commands/list.py
```diff
@@ -0,0 +1,26 @@
+from piccolo.apps.user.tables import BaseUser
+
+
+def list():
+    """
+    List existing users.
+    """
+    users = BaseUser.select(
+        BaseUser.all_columns(exclude=[BaseUser.password])
+    ).run_sync()
```
dantownsend, line 10: I'd probably make this `async`.

piccolo/apps/user/commands/list.py
```diff
@@ -0,0 +1,26 @@
...
+
+    if len(users) < 1:
+        return
```
dantownsend, line 13: Is it worth printing out something like 'No users'?

piccolo/apps/user/commands/list.py
dantownsend, line 10: We can sort the results - I guess `username` is most useful?

piccolo/apps/user/commands/list.py
```diff
@@ -0,0 +1,26 @@
...
+
+    for user in users:
+        for column in ref_order:
+            if len(str(user[column])) > width[column]:
+                width[column] = len(str(user[column]))
+
+    format_string = "|".join([f"{{:<{width[w]}}}" for w in ref_order])
+    print(format_string.format(*[w for w in ref_order]))
+    for user in users:
+        print(format_string.format(*[str(user[w]) for w in ref_order]))
```
dantownsend, line 26: I think what you've done looks cool.

Screenshot 2024-01-22 at 15 34 00

What do you think about adding a space either side of the `|` symbols? Only if it doesn't add loads of complexity.

dantownsend: There's a file called `piccolo/utils/printing.py` which has utilities for this kind of thing. We don't have anything which does exactly what you're doing, but maybe it's worth add it there so it's reusable?

dantownsend (commented):
This looks great - thanks!

I left a few comments.

piccolo/apps/user/commands/list.py
waldner: I've checked other `user` subcommands and they are synchronous.

piccolo/apps/user/commands/list.py
waldner: Depends on how much of a purist you are. Personally, I'd just leave it with no output at all, but your call.

piccolo/apps/user/commands/list.py
waldner: Done.

piccolo/apps/user/commands/list.py
waldner: Done both things.

piccolo/apps/user/commands/list.py
dantownsend: You're right - not a big problem, lets leave it.

piccolo/apps/user/commands/list.py
dantownsend: My only concern with not printing something out is the user might think it's broken.

waldner: Alright, I added a message.

dantownsend:
OK, this looks good - thanks! I'll merge it in.

I'll add a really basic test for code coverage.

waldner:
Thanks!

PR #928, opened 2024-01-26
Allow `Migration` table's name and schema to be overridden

dantownsend:
Related discussion https://github.com/piccolo-orm/piccolo/discussions/895

Piccolo has some important tables which are provided out of the box - most importantly `Migration` for tracking database migrations, and `BaseUser` for user management.

The name and Postgres schema of these tables is currently fixed. The problem is:

* If there's already a table in the database with a matching name
* If you want these tables to be in a different Postgres schema

The solution up until now has required hacks, or copying the Piccolo apps into your own codebase and modifying them.

I've been through a few potential solutions. The most obvious is adding configuration options in `piccolo_conf.py`. The problem though is when `piccolo_conf.py` is loaded, it runs some validation on your project to make sure all of the registered apps and tables are available. So if our tables rely on configuration in `piccolo_conf.py` to be created, it would create a circular dependency.

This PR lets the user create a `piccolo_overrides.toml` file at the root of the their project (alongside the `piccolo_conf.py` file, where the Python interpreter will be launched). For example:

```toml
PICCOLO_MIGRATION_SCHEMA="piccolo_migrations"
```

Because it's a static file, it has some benefits:

* It won't change at run time, which we really don't want.
* It can safely be imported very early in the app life-cycle, without causing any circular imports.
* No Python code can be run in this file, causing potential side-effects.

Users can use this file to provide configuration for their own custom apps too, if they want.

```toml
TICKETS_PER_USER=10
```

What do people think?


dantownsend:
@sinisaos What do you think of this idea?

sinisaos (approved):
This is great and works as usual :). Another benefit (besides migrations etc.) is that, as you wrote, the user can use this file to provide custom configuration that can be used anywhere in the app as needed, which I think is great.

aabmets:
@dantownsend 

TL;DR - no

Longer explanation:

I'm going to be straight-up honest here. The improvements which you have proposed are essential and very welcome, however I must disagree with the entire current concept of how Piccolo globals and config files must be used. 

Here's what an ideal solution would look like:
Instead of demanding users to employ `piccolo_conf.py`, `piccolo_app.py` and `piccolo_overrides.toml` files so Piccolo would be able to understand how it needs to work, Piccolo should not require **any** specific files for its functionality, because it's a bad practice to drown the end-user (developer) into endless configuration files, which is exacerbated when multiple libraries think (incorrectly) that its a fantastic approach to solve library configuration.

What should happen instead is that the global state of the library should be managed inside the library itself and the user should just import and call functions from within the library which alter the global database engine configuration, migrations schema name, etc. The less config files a library needs, the more elegant you are able to define its interface and improve Developer Experience. For instance, if it would be implemented correctly, there would be no need for some kind of `engine_finder` or some convoluted logic how engines are defined and used. It could be simplified so much while retaining the power of configurability.

It must be also kept in mind, that the users project code most likely might want to build a Docker image from their app, which means that their app is executed from some kind of `main.py` file. The user must be responsible for calling the necessary functions inside their code flow to ensure the proper configuration of the Piccolo ORM.

dantownsend:
@aabmets The reason we have `piccolo_conf.py` is as a convenience to the user. Instead they can do this:

```python
DB = PostgresEngine(...)

class MyTable(Table, db=DB):
    ...
```

Which works fine, but the downsides are:

1. You're passing DB around everywhere.
2. If you're not the author of a table class (i.e. you're importing it from another module) it won't work.

Alternatively, the user could does this:

```python
# main.py
for table in (MyTable, SomeOtherTable, ...):
    table._meta.db = DB
```

This works fine, as long as the user doesn't miss out any tables. It gets around point 2 above, because you can configure tables which you didn't write yourself.

The problem though, is Piccolo has two potential entry points. The first is something like `main.py`, which runs a web app etc. And the second is the Piccolo CLI.

With the Piccolo CLI, it doesn't give the user a chance to configure all of their tables, so the `for loop` option above doesn't work well in that situation.

I'm not saying things at the moment are perfect. I can see the value in a very explicit style, and Piccolo will evolve with time. But I hope that gives some context for the existence of the whole `piccolo_conf.py` and `engine_finder` thing.

I'm not 100% sure the `piccolo_overrides.toml` is the right approach yet. I've prototyped a few different ideas.

aabmets:
@dantownsend 
Well, lets start with small steps. Would it be possible to allow `piccolo_app.py` content into `piccolo_conf.py` for single-app projects? What conditions or steps should be fulfilled to realize this feature?

dantownsend:
> Well, lets start with small steps. Would it be possible to allow piccolo_app.py content into piccolo_conf.py for single-app projects? What conditions or steps should be fulfilled to realize this feature?

@aabmets  That's a good point - at the moment `AppRegistry` accepts a list of strings to import, but I see no reason it can't also accept `AppConfig` as well.

```python
# piccolo_conf.py
DB = PostgresEngine(...)

app = AppConfig(app_name="main", ...)

APP_REGISTRY = AppRegistry(apps=[app])
```

It just requires a small change here:

https://github.com/piccolo-orm/piccolo/blob/46e72f439ba5e9e441c57e3942622d0d6d1024bf/piccolo/conf/apps.py#L231-L254

So if it's an `AppConfig` instance, add it straight to `self.app_configs`.

sinisaos:
> Here's what an ideal solution would look like:
> Instead of demanding users to employ `piccolo_conf.py`, `piccolo_app.py` and `piccolo_overrides.toml` files so Piccolo would be able to understand how it needs to work, Piccolo should not require **any** specific files for its functionality, because it's a bad practice to drown the end-user (developer) into endless configuration files, which is exacerbated when multiple libraries think (incorrectly) that its a fantastic approach to solve library configuration.

@aabmets Sorry to jump into this discussion, but I think it would be best if you make a PR (I hope @dantownsend agrees) on how Piccolo should look and work without these config files and remain suitable for the average user like me. Thanks in advance.

aabmets:
@sinisaos 
I'll just probably create my own ORM when I get around to it, I'm not into fixing other peoples code, as I'm somewhat of a stickler on how libraries should be structured and interfaced with and it would be improper of me to force my solutions onto other peoples projects (so far, I've only offered arugmented suggestions).

sinisaos:
@aabmets Thank you for your reply and clarification.

tarsil:
Hey @dantownsend i love your idea but I also understand what @aabmets was mentioning above. A good way to do this is to so what Esmerald for instance does with some lazy global settings where you could override the defaults if needed to?

Dymmond made it available that well known lazy settings approach that was inspired by Django.

[Link here](https://settings.dymmond.com/) maybe I'm missing something but I do believe this could meet what you are looking for without incurring the developer with extra configuration files?

sinisaos:
@tarsil Thanks for your suggestion. I use [pydantic-settings](https://docs.pydantic.dev/latest/concepts/pydantic_settings/) (I didn't go into details, but it seems to me that `dymmond-settings` is similar) for configuration files, but here I was thinking more about how Piccolo would work without `piccolo_conf.py` and `piccolo_app.py` (as @aabmets mention) which are important configuration files for Piccolo and what it would mean for the library and usage of library (which is now very easy).

aabmets:
@sinisaos
If I have understood @dantownsend correctly (feel free to correct me), Piccolo needs specific configuration files in the users project due to Piccolos tight integration with its CLI. This indicates that the library was built around the CLI, not the other way around, but it would be incorrect to assume that the CLI is a necessary component of working with the library. Yes, the CLI can be optionally used as a convenience by the dev during the dev process, but if we are talking about using Piccolo for serious enterprise grade apps (my scenario), then there is a guaranteed need to precisely control Piccolos behavior through a Python script, even for migrations. Because Piccolo is CLI-first, I need to perform some hacks and patches to get Piccolo work the way I need it to work in my companys project. As a side note, SQLAlchemy, Tortoise-ORM, Pony ORM, etc. are even worse to work with than Piccolo, so kudos on that. From a professional standpoint, a library should be designed as a library-first, and then offer an optional CLI component if the user wants to clutter their project with bootstrap files.

dantownsend:
Thanks everyone for their input.

One thing we could do is let the user pass in the `Migration` table to the relevant migration commands.

We do something similar in other places - like [Piccolo Admin](https://github.com/piccolo-orm/piccolo_admin/blob/7f3383bfd403f07e2af4c3c2fd2437380c41360c/piccolo_admin/endpoints.py#L1063-L1064
).

The Piccolo CLI commands are nothing special - they're just standard functions.

https://github.com/piccolo-orm/piccolo/blob/46e72f439ba5e9e441c57e3942622d0d6d1024bf/piccolo/apps/migrations/commands/check.py#L116-L124

We could accept the path to a Migration table, if using it on the command line:

```
piccolo migrations check --migration_table=my_tables.Migration
```

Or if running it from your own script:

```python
from piccolo.apps.migrations.table import Migration as MigrationBase

class Migration(MigrationBase, schema="my_schema"):
    ...

await check_migrations(migration_table=Migration)
```

aabmets:
@dantownsend 

```python
from __future__ import annotations

from typing import Type
from functools import lru_cache

@lru_cache
def migration_table_factory(table_name: str, schema_name: str) -> Type[Table]:
   class _CustomMigration(Table, tablename=table_name, schema=schema_name):
       # class content is identical to the original Migration table
       
   return _CustomMigration
```

This solution would require replacing all direct access of the Migration table with a call to the `migration_table_factory` function. I think this solution would work even if the Migration table is stateful due to the `lru_cache` decorator, which would ensure that the same class is being returned for identical argument inputs to the factory function. This solution would allow to manage migrations separately for each schema in a database.

PR #952, opened 2024-03-14
docs: Update code example of prefetching related objects

jrycw:
To ensure equivalence between the two code snippets, I believe we may be missing the `.first()` method call in the last one.
```python
# These are equivalent:
ticket = await Ticket.objects(
    Ticket.concert.all_related()
).first()

ticket = await Ticket.objects().prefetch(
    Ticket.concert.all_related()
) # <= missing `.first()` here
```


dantownsend:
@jrycw You're right - great catch! 👍

Issue #583, opened 2022-08-11
FEATURE: Composite Unique Constraints


northpowered:
_Previous issues_
#572 #172
_Previous discussions_
#175 

### Pull request #582 

Usage:
```python3
from piccolo.colums.constraints import UniqueConstraint

class FooTable(Table):
    foo_field = Text()
    bar_field = Text()
    my_constraint_1 = UniqueConstraint(['foo_field','bar_field'])
```
or multiple constraints: 
```python3
from piccolo.colums.constraints import UniqueConstraint

class FooTable(Table):
    foo_field = Text()
    bar_field = Text()
    spam_field = Text()
    eggs_field = Text()
    my_constraint_1 = UniqueConstraint(['foo_field','bar_field'])
    my_constraint_2 = UniqueConstraint(['spam_field','eggs_field'])
```

Auto migrations are working for:

1. Creation with Table (CREATE TABLE)
2. Creation after Table creation (ALTER TABLE ADD CONSTRAINT)
3. Drop (DROP CONSTRAINT)

If You want to update constraint (ex. add new columns to UniqueConstraint()), You should, at first DROP, and then CREATE the new one. Example:
```python
 #Comment the line
 #my_constraint_1 = UniqueConstraint(['foo_field','bar_field'])

>> ...migrate...

 #Uncomment the line
 my_constraint_1 = UniqueConstraint(['foo_field','spam_field'])

>> ...migrate...
```

In progress:

1. ALTER constraint without recreation

Some important notes:

1. UniqueConstraint class inherit from Column class, so the most part of behavior is like in Column objects
2. Operations with UC are showing in migration summary like columns (Added colums: ..., Dropped colums:... etc)
3. I`d used to add extra_imports to auto drop in migrations app to be able to differ usual colums and constraints
4. The same about DropColumn class, I added **column_class** param, like in AddColumn
5. Fixed some tests (DropColumn), because **column_class** param raised an AssertionError

Help and fixes are welcome!

Issue #172, opened 2021-08-20
Composite uniqueness constraints


adriangb:
It looks like there's no way to do composite uniqueness constraints, i.e. `UNIQUE (col1, col2)`.

dantownsend:
Yep, this needs adding.

A link to a discussion about it: https://github.com/piccolo-orm/piccolo/discussions/175

PR #957, opened 2024-03-17
Support composite unique constraints in auto migration

atkei:
Related to #172 , #175 and based on #582, #583.

- Add `UniqueConstraint` class with the same behavior as a `Column` to support auto migration with composite unique constraint
- Update existing tests for internal changes and add tests for auto maigration with adding/dropping unique constraint

### DIFF
--- a/piccolo/apps/migrations/auto/diffable_table.py
+++ b/piccolo/apps/migrations/auto/diffable_table.py
@@ -146,6 +146,7 @@ class DiffableTable:
                 db_column_name=i.column._meta.db_column_name,
                 tablename=value.tablename,
                 schema=self.schema,
+                column_class=i.column.__class__,
             )
             for i in sorted(
                 {ColumnComparison(column=column) for column in value.columns}
--- a/piccolo/apps/migrations/auto/migration_manager.py
+++ b/piccolo/apps/migrations/auto/migration_manager.py
@@ -16,9 +16,11 @@ from piccolo.apps.migrations.auto.operations import (
 from piccolo.apps.migrations.auto.serialisation import deserialise_params
 from piccolo.columns import Column, column_types
 from piccolo.columns.column_types import ForeignKey, Serial
+from piccolo.columns.constraints import UniqueConstraint
 from piccolo.engine import engine_finder
 from piccolo.query import Query
 from piccolo.query.base import DDL
+from piccolo.query.methods.alter import AddUniqueConstraint, DropConstraint
 from piccolo.schema import SchemaDDLBase
 from piccolo.table import Table, create_table_class, sort_table_classes
 from piccolo.utils.warnings import colored_warning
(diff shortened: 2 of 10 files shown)


dantownsend:
Thanks a lot for this!

I did a quick review, and it looks good. I appreciate the work you've put into it.

As it's quite a big feature, I'll do a more in depth review, and play around with it a bit.

atkei:
@dantownsend 
Thank you for your great work.
I love piccolo and I'm using it in my new project.
Let me help.

dantownsend:
@sinisaos Can you help me test this if you get some spare time?

sinisaos:
@dantownsend Sure. I will look into this during the day.

sinisaos:
@dantownsend  Migrations are fine and constraints are correctly added to the database.

constraints

The problems come after that when we try to query table. First error is
```bash
File "/home/rkl/dev/unique_env/lib/python3.11/site-packages/piccolo/columns/base.py", line 831, in get_select_string
    if self._alias:
       ^^^^^^^^^^^
AttributeError: 'UniqueConstraint' object has no attribute '_alias'
```
If I prevent AttributeError with `try-except` block another error is occured
```bash
 File "asyncpg/protocol/protocol.pyx", line 166, in prepare
asyncpg.exceptions.UndefinedColumnError: column task.task_constraints does not exist
```
I think the main problem is that `UniqueConstraint` is treated as a column, but that column does not exist, which is confirmed if print querystring like this `print(Task.select().order_by(Task.id))`, result is
```bash
SELECT ALL "task"."id", "task"."name", "task"."description", "task"."completed", "task"."task_constraints" FROM "task" ORDER BY "task"."id" ASC
```

atkei:
@sinisaos @dantownsend 

Thank you for your support.
I underestimated the impact of extending column class.
Considering extensibility and ease of testing, is the following better?

- Add a Constraint class like SQLAlchemy that does not inherit from the Column class, and a UniqueConstraint inherit from it, so easy to add CheckConstraint if users want
- In the Migration Manager, run _run_add_constraints() and _run_drop_constraints() to add and drop constraints

If you agree, I will close this PR and try again with this approach.

atkei:
@sinisaos Yeah, I use raw SQL in my project, so this feature is not mandatory:)
Thank you very much.

sinisaos:
@dantownsend @atkei I have a question for both of you. Did it make sense to add and drop unique constraints through the `Piccolo CLI`? This would avoid defining unique constraints on the table definition. User input should be validated, but that's basically it. It would look like in the video. 

[constraints.webm](https://github.com/piccolo-orm/piccolo/assets/30960668/8eea58b1-cc82-41f6-819c-bb51c2c8c040)

It's just an idea, so please give your opinion.

dantownsend:
@sinisaos It's a nice idea.

I do like having constraints defined in migrations, because when collaborating with other team members they get the constraints without having to do anything besides run the migrations.

I think this PR has a lot of promise, but constraints were treated like a special form of column, which while elegant in some ways, might also be confusing. I think I'd prefer to treat them as totally different things if possible.

sinisaos:
> I do like having constraints defined in migrations, because when collaborating with other team members they get the constraints without having to do anything besides run the migrations.

You're right. I didn't think about that at all. I'm not a professional developer and have never worked in a team, but what you say absolutely makes sense if more than one person is working on the same project. Thanks for the clarification and forget about the `CLI` solution. That is not good and as you say, it is much better to have defined constraints on migrations. Sorry and thanks again.

PR #959, opened 2024-03-18
Datetime Autoconversion for Timestamptz Columns

aabmets:
I made this pull request to rectify the issue where Piccolo does not provide a way to tell Timestamptz columns in which timezones they should identify themselves as.

_Outtake from PG docs:_
> Internally, PostgreSQL stores the timestamptz in UTC value. 
> 
> When you insert a value into a timestamptz column, PostgreSQL converts the timestamptz value into a UTC value and stores the UTC value in the table.
> When you retrieve data from a timestamptz column, PostgreSQL converts the UTC value back to the time value of the timezone set by the database server, the user, or the current database connection.

The solutions provided by Postgres as by which it determines the timezone to convert the value back to was not working properly with Piccolo. Hence I made this pull request. It works, please merge. Kinda need it in my enterprise project. @dantownsend

### DIFF
--- a/piccolo/table.py
+++ b/piccolo/table.py
@@ -5,6 +5,8 @@ import itertools
 import types
 import typing as t
 import warnings
+from datetime import datetime
+from zoneinfo import ZoneInfo
 from dataclasses import dataclass, field
 
 from piccolo.columns import Column
@@ -17,6 +19,7 @@ from piccolo.columns.column_types import (
     ReferencedTable,
     Secret,
     Serial,
+    Timestamptz,
 )
 from piccolo.columns.defaults.base import Default
 from piccolo.columns.indexes import IndexMethod
@@ -436,6 +439,9 @@ class Table(metaclass=TableMetaclass):
                 ):
                     raise ValueError(f"{column._meta.name} wasn't provided")
 
+            if isinstance(column, Timestamptz) and isinstance(value, datetime):
+                value = value.astimezone(column.tz)
+                
             self[column._meta.name] = value
 
         unrecognized = kwargs.keys()
--- a/piccolo/columns/column_types.py
+++ b/piccolo/columns/column_types.py
@@ -33,6 +33,7 @@ import typing as t
 import uuid
 from dataclasses import dataclass
 from datetime import date, datetime, time, timedelta
+from zoneinfo import ZoneInfo
 from enum import Enum
 
 from piccolo.columns.base import (
@@ -1007,6 +1008,7 @@ class Timestamptz(Column):
     """
 
     value_type = datetime
+    tz_type = ZoneInfo
 
     # Currently just used by ModelBuilder, to know that we want a timezone
     # aware datetime.
@@ -1015,7 +1017,10 @@ class Timestamptz(Column):
     timedelta_delegate = TimedeltaDelegate()
 
     def __init__(
-        self, default: TimestamptzArg = TimestamptzNow(), **kwargs
+        self,
+        tz: ZoneInfo = ZoneInfo('UTC'),
+        default: TimestamptzArg = TimestamptzNow(), 
+        **kwargs
     ) -> None:
         self._validate_default(
             default, TimestamptzArg.__args__  # type: ignore
@@ -1025,10 +1030,11 @@ class Timestamptz(Column):
             default = TimestamptzCustom.from_datetime(default)
 
         if default == datetime.now:
-            default = TimestamptzNow()
+            default = TimestamptzNow(tz)
 
+        self.tz = tz
         self.default = default
-        kwargs.update({"default": default})
+        kwargs.update({"tz": tz, "default": default})
         super().__init__(**kwargs)
 
     ###########################################################################
--- a/piccolo/columns/defaults/timestamptz.py
+++ b/piccolo/columns/defaults/timestamptz.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import datetime
 import typing as t
 from enum import Enum
+from zoneinfo import ZoneInfo
 
 from .timestamp import TimestampCustom, TimestampNow, TimestampOffset
 
@@ -27,12 +28,15 @@ class TimestamptzOffset(TimestampOffset):
 
 
 class TimestamptzNow(TimestampNow):
+    def __init__(self, tz: ZoneInfo = ZoneInfo('UTC')):
+        self._tz = tz
+        
     @property
     def cockroach(self):
         return "current_timestamp"
 
     def python(self):
-        return datetime.datetime.now(tz=datetime.timezone.utc)
+        return datetime.datetime.now(tz=self._tz)
 
 
 class TimestamptzCustom(TimestampCustom):


dantownsend:
Thanks for this. I think it makes sense.

`zoneinfo` seems to be Python 3.9 and above, so we'll need to add this to `requirements.txt` for Python 3.8:

https://pypi.org/project/backports.zoneinfo/

aabmets:
@dantownsend The initial proposal contained some shortcomings which should be fully fixed by the subsequent commits.

dantownsend:
It's looking good, thanks. It seems like there's some minor linter and test issues.

sinisaos:
@aabmets Sorry to intrude into this conversation, but first you need to run linters and tests locally with this set of commands, (then you'll see what you need to fix)
1. Run formaters: `./scripts/format.sh`
2. Run linters: `./scripts/lint.sh`
3. Run the test suite with Sqlite:` ./scripts/test-sqlite.sh`
4. Run the test suite with Postgres: `./scripts/test-postgres.sh`
5. Run the test suite with Cockroach: `./scripts/test-cockroach.sh`

Code contribution guide is [here](https://piccolo-orm.readthedocs.io/en/latest/piccolo/contributing/index.html#get-the-tests-running) and code style is [here](https://piccolo-orm.readthedocs.io/en/latest/piccolo/contributing/index.html#code-style). Hope this help to pass linters check and tests

aabmets:
@sinisaos Thanks for the help, I'll try doing that option.

aabmets:
@sinisaos I am unable to run these scripts, because Piccolo workflow is incompatible with Windows platform, and I don't use Linux at all. I tried running them on WSL, but WSL just complains about invalid script contents. Would you be so kind as to run these scripts and fix my pull request so it could get merged?

aabmets:
@sinisaos I think i need help with fixing migration errors. The problem is that piccolo is not inserting 
`from zoneinfo import ZoneInfo` into autogenerated migrations file when Python version is >= 3.9 or `from backports.zoneinfo import ZoneInfo` when Python version is <= 3.8.
I tried looking at the MigrationsManager logic, but for the life of me I was unable to understand how it determines the imports that it needs to inject into the migrations files.

aabmets:
@dantownsend Okay I think I've done it, the linting and unittests should be fixed now.

sinisaos:
@aabmets Sorry for the late reply. When I start formatting, a lot of files are not well formatted and to satisfy mypy in Python3.8 I have to `# type ignore` all `from zoneinfo import ZoneInfo` lines (another way can be to add `zoneinfo` to `pyproject.toml` `tool.mypy.overrides`). That is not big deal because `./scripts/format.sh` and .`/scripts/lint.sh` will solve that.
Also `tests/columns/test_timestamptz.py::TestTimestamptzDefault::test_timestamptz_default` does not pass both Sqlite and Postgres. You must wait for @dantownsend to solve these linters and tests problems.

aabmets:
@sinisaos hey, thanks for also taking a look at possible solutions. I think my latest commits fixed both linting and test issues  waiting for approval to rerun workflow.

aabmets:
For crying out loud 😂 why is the god damn linter test suite failing. I ran the linting and style commands on the commited files.

aabmets:
@sinisaos Would you be so kind as to re-run the CI workflow, thanks.

sinisaos:
@aabmets I'm sorry, but I'm not a maintainer and I don't have permission to do that. You have to wait for @dantownsend for that. Daniel, can you please fix this issue with linters as `./scripts/format.sh` and .`/scripts/lint.sh` will fix that. Thanks.

dantownsend:
@sinisaos yes, will try and do it later tonight. I've given you extra permissions now, so in the future you should be able to approve running CI. GitHub permissions are a bit confusing.

aabmets:
@dantownsend @sinisaos Hey, at least we're seeing some improvements! I see only one failing test now.

sinisaos:
@aabmets To satisfy mypy in Python3.8 try to `# type ignore` all `from zoneinfo import ZoneInfo` lines (another way can be to add `zoneinfo` to `pyproject.toml` `tool.mypy.overrides`) That should work.

> I've given you extra permissions now, so in the future you should be able to approve running CI. GitHub permissions are a bit confusing.

@dantownsend Thanks. I also don't know much about Github permissions.

aabmets:
@sinisaos Done. I couldn't find a solution how to surgically disable MyPy from `pyproject.toml` so I went with the line-by-line solution. I don't think blanket-ignore would be the correct solution, as that could cause misses on legitimate type errors.

aabmets:
@dantownsend @sinisaos 
Is it necessary to fix the codecov complaints about this PR?  
The fixes for all impacted lines entail adding a `#pragma: no cover`  
to the except clause in ZoneInfo import statements:
``` python
try:
    from zoneinfo import ZoneInfo  # type: ignore
except ImportError:  # pragma: no cover  <-- necessary to add to silence codecov
    from backports.zoneinfo import ZoneInfo  # type: ignore  # noqa: F401
```

dantownsend:
@aabmets Yes, if you don't mind - we use `# pragma: no cover` in a few other places in the codebase. I think it makes sense here.

piccolo/table.py
```diff
@@ -55,6 +57,11 @@
 from piccolo.utils.sync import run_sync
 from piccolo.utils.warnings import colored_warning
 
+try:
+    from zoneinfo import ZoneInfo  # type: ignore
+except ImportError:  # pragma: no cover
+    from backports.zoneinfo import ZoneInfo  # type: ignore  # noqa: F401
+
```
sinisaos, line 64: @aabmets I think you should remove this imports as it is not used anywhere in the file

aabmets: @sinisaos Done

aabmets:
@dantownsend Not to be a bother, but would it be perchance possible to merge this PR into master and release a minor version in the near future?

sinisaos:
@dantownsend I tested this PR a bit. `objects` queries work fine, but `select` queries do not (results are the same as in the master branch code). Here is an example of the results from `Piccolo` shell. 
```python
# objects query
In [1]: [i.to_dict() for i in await TallinnConcerts.objects()]
Out[1]: 
[{'id': 1,
  'event_start': datetime.datetime(2050, 1, 1, 21, 0, tzinfo=zoneinfo.ZoneInfo(key='Europe/Tallinn'))},
 {'id': 2,
  'event_start': datetime.datetime(2050, 1, 1, 21, 0, tzinfo=zoneinfo.ZoneInfo(key='Europe/Tallinn'))}]
# select queries
In [2]: await TallinnConcerts.select()
Out[2]: 
[{'id': 1,
  'event_start': datetime.datetime(2050, 1, 1, 19, 0, tzinfo=datetime.timezone.utc)},
 {'id': 2,
  'event_start': datetime.datetime(2050, 1, 1, 19, 0, tzinfo=datetime.timezone.utc)}]
```

I don't know if this is expected behavior (or I'm doing something wrong), but it doesn't match this lines from the docstring.

https://github.com/piccolo-orm/piccolo/blob/db539394018bfdb1e03b2c67e295cabe61a03cf6/piccolo/columns/column_types.py#L983-L991

dantownsend:
@sinisaos Thanks for testing it - that's very helpful.

sinisaos:
@dantownsend I found another problem. When we try to bulk update a column in Piccolo Admin, the fields don't show up in dropdown with the changes from this PR.

[bulk_update.webm](https://github.com/piccolo-orm/piccolo/assets/30960668/9d55a972-e131-449a-b0a1-8664e89b3658)

dantownsend:
I've had a chance to test this now. As @sinisaos says, the issue at the moment is with `select` queries. `objects` queries work fine. It looks like we're doing the timezone conversion inside the `Table.__init__`, which is a smart idea, but doesn't work for `select`.

I think instead we might be able to modify the select query SQL to use `AT TIME ZONE`, and let the database do the heavy lifting for us. I doubt SQLite supports this, so will need a work around. Will play around with it a bit.

dantownsend:
Using `AT TIME ZONE` works OK, but the main problem is it's returned as a naive timestamp. Maybe that's OK ... but my feeling is it's better to somehow add the timezone information back.

dantownsend:
Besides writing some docs, and a few more tests, I think this is done now:

https://github.com/piccolo-orm/piccolo/pull/978

It just adds a few things to this original PR. You can override the timezone in a select query:

```python
await Concert.select(Concert.starts.at_time_zone('America/New_York'))
```

aabmets:
@dantownsend @sinisaos Thank you for taking the time to test and improve this PR. 
As I am not intimately familiar with Piccolo internals, some oversights were bound to remain in my original implementation. I'm glad you were able to find solutions to the remaining problems. 

I do have a question about the proposed solution to the select queries. If a column is designated as Timestamptz with tz information set with ZoneInfo, that column identifies itself as only existing in that timezone, meaning that whenever a value of that column is being accessed, the user can be confident that the value is in the timezone that they defined the column as. When the user is able to override the timezone information with `at_time_zone`, that defeats the whole purpose of assigning timezone information about the column at table definition, because it provides a supported workaround how to alter the values stored in the column in a non-congruent way to the column timezone definition. Maybe it should work like this, maybe it's a good idea. I do not know however, that's why I'm asking.

sinisaos:
Closed in favor of #978 which use code from this PR.

PR #971, opened 2024-03-22
Refactor `ModelBuilder` and `RandomBuilder`

jrycw:
I encountered `ModelBuilder._randomize_attribute` and initially found the numerous `if-elif-else` checks puzzling. However, further investigation revealed that certain factories necessitate additional information from the `column`. I attempted to consolidate these logics into `RandomBuilder._build`, aiming for improved clarity. Nevertheless, the final code may not be as pristine as desired. I'm looking forward to hearing the community's thoughts on whether this refactoring improves the codebase.

### DIFF
--- a/piccolo/testing/model_builder.py
+++ b/piccolo/testing/model_builder.py
@@ -6,6 +6,7 @@ import json
 import typing as t
 from decimal import Decimal
 from functools import partial
+from types import MappingProxyType
 
 from piccolo.columns import JSON, JSONB, Array, Column, ForeignKey
 from piccolo.custom_types import TableInstance
@@ -97,7 +98,7 @@ class ModelBuilder:
         persist: bool = True,
     ) -> TableInstance:
         model = table_class(_ignore_missing=True)
-        defaults = {} if not defaults else defaults
+        defaults = defaults or {}
 
         for column, value in defaults.items():
             if isinstance(column, str):
@@ -150,38 +151,46 @@ class ModelBuilder:
             Column class to randomize.
 
         """
-        random_value: t.Any
-        default_mapper = cls._get_default_mapper()
-        local_mapper = cls._get_local_mapper(column)
-        other_mapper = cls._get_other_mapper(column)
-        # order matters
-        mapper = {**default_mapper, **local_mapper, **other_mapper}
-
+        reg = cls.get_registry(column)
         if column._meta.choices:
             random_value = RandomBuilder.next_enum(column._meta.choices)
-        elif column.value_type == list:
-            length = RandomBuilder.next_int(maximum=10)
-            base_type = t.cast(Array, column).base_column.value_type
-            random_value = [mapper[base_type]() for _ in range(length)]
         else:
-            random_value = mapper[column.value_type]()
+            random_value = reg[column.value_type]()
 
         if isinstance(column, (JSON, JSONB)):
             return json.dumps({"value": random_value})
         return random_value
 
     @classmethod
-    def _get_default_mapper(cls) -> t.Dict[t.Type, t.Callable]:
+    def get_registry(
+        cls, column: Column
+    ) -> MappingProxyType[t.Type, t.Callable]:
         """
-        This classmethod encapsulates the desired logic.
+        This serves as the public API allowing users to **view**
+        the complete registry for the specified column.
+
+        :param column:
+            Column class to randomize.
+
         """
-        mapper = cls.__DEFAULT_MAPPER
-        if not mapper:  # execute once only
-            for typ, callable_name in RandomBuilder.DEFAULT_MAPPER.items():
-                # a simpler approach available?
-                func = RandomBuilder.__dict__[callable_name].__func__
-                mapper[typ] = partial(func, RandomBuilder)
-        return mapper
+        default_mapper = cls.__DEFAULT_MAPPER
+        if not default_mapper:  # execute once only
+            for typ, callable_ in RandomBuilder.get_mapper().items():
+                default_mapper[typ] = callable_
+
+        # order matters
+        reg = {
+            **default_mapper,
+            **cls._get_local_mapper(column),
+            **cls._get_other_mapper(column),
+        }
+
+        if column.value_type == list:
+            reg[list] = partial(
+                RandomBuilder.next_list,
+                reg[t.cast(Array, column).base_column.value_type],
+            )
+        return MappingProxyType(reg)
 
     @classmethod
     def _get_local_mapper(cls, column: Column) -> t.Dict[t.Type, t.Callable]:
@@ -193,19 +202,20 @@ class ModelBuilder:
             Column class to randomize.
         """
         local_mapper: t.Dict[t.Type, t.Callable] = {}
-        if column.value_type == Decimal:
-            precision, scale = column._meta.params["digits"] or (4, 2)
-            local_mapper[Decimal] = partial(
-                RandomBuilder.next_decimal, precision, scale
-            )
-        elif column.value_type == datetime.datetime:
-            tz_aware = getattr(column, "tz_aware", False)
-            local_mapper[datetime.datetime] = partial(
-                RandomBuilder.next_datetime, tz_aware
-            )
-        elif column.value_type == str:
-            if _length := column._meta.params.get("length"):
-                local_mapper[str] = partial(RandomBuilder.next_str, _length)
+
+        precision, scale = column._meta.params.get("digits") or (4, 2)
+        local_mapper[Decimal] = partial(
+            RandomBuilder.next_decimal, precision, scale
+        )
+
+        tz_aware = getattr(column, "tz_aware", False)
+        local_mapper[datetime.datetime] = partial(
+            RandomBuilder.next_datetime, tz_aware
+        )
+
+        if _length := column._meta.params.get("length"):
+            local_mapper[str] = partial(RandomBuilder.next_str, _length)
+
         return local_mapper
 
     @classmethod
@@ -241,10 +251,10 @@ class ModelBuilder:
         return other_mapper
 
     @classmethod
-    def register_random_type(cls, typ: t.Type, callable_: t.Callable) -> None:
+    def register_type(cls, typ: t.Type, callable_: t.Callable) -> None:
         cls.__OTHER_MAPPER[typ] = callable_
 
     @classmethod
-    def unregister_random_type(cls, typ: t.Type) -> None:
+    def unregister_type(cls, typ: t.Type) -> None:
         if typ in cls.__OTHER_MAPPER:
             del cls.__OTHER_MAPPER[typ]
--- a/piccolo/testing/model_builder.py
+++ b/piccolo/testing/model_builder.py
@@ -1,15 +1,22 @@
 from __future__ import annotations
 
+import datetime
+import inspect
 import json
 import typing as t
+from decimal import Decimal
+from functools import partial
 
-from piccolo.columns import JSON, JSONB, Column, ForeignKey
+from piccolo.columns import JSON, JSONB, Array, Column, ForeignKey
 from piccolo.custom_types import TableInstance
 from piccolo.testing.random_builder import RandomBuilder
 from piccolo.utils.sync import run_sync
 
 
 class ModelBuilder:
+    __DEFAULT_MAPPER: t.Dict[t.Type, t.Callable] = {}
+    __OTHER_MAPPER: t.Dict[t.Type, t.Callable] = {}
+
     @classmethod
     async def build(
         cls,
@@ -143,7 +150,101 @@ class ModelBuilder:
             Column class to randomize.
 
         """
-        random_value: t.Any = RandomBuilder._build(column)
+        random_value: t.Any
+        default_mapper = cls._get_default_mapper()
+        local_mapper = cls._get_local_mapper(column)
+        other_mapper = cls._get_other_mapper(column)
+        # order matters
+        mapper = {**default_mapper, **local_mapper, **other_mapper}
+
+        if column._meta.choices:
+            random_value = RandomBuilder.next_enum(column._meta.choices)
+        elif column.value_type == list:
+            length = RandomBuilder.next_int(maximum=10)
+            base_type = t.cast(Array, column).base_column.value_type
+            random_value = [mapper[base_type]() for _ in range(length)]
+        else:
+            random_value = mapper[column.value_type]()
+
         if isinstance(column, (JSON, JSONB)):
             return json.dumps({"value": random_value})
         return random_value
+
+    @classmethod
+    def _get_default_mapper(cls) -> t.Dict[t.Type, t.Callable]:
+        """
+        This classmethod encapsulates the desired logic.
+        """
+        mapper = cls.__DEFAULT_MAPPER
+        if not mapper:  # execute once only
+            for typ, callable_name in RandomBuilder.DEFAULT_MAPPER.items():
+                # a simpler approach available?
+                func = RandomBuilder.__dict__[callable_name].__func__
+                mapper[typ] = partial(func, RandomBuilder)
+        return mapper
+
+    @classmethod
+    def _get_local_mapper(cls, column: Column) -> t.Dict[t.Type, t.Callable]:
+        """
+        This classmethod encapsulates the desired logic, utilizing information
+        from the column.
+
+        :param column:
+            Column class to randomize.
+        """
+        local_mapper: t.Dict[t.Type, t.Callable] = {}
+        if column.value_type == Decimal:
+            precision, scale = column._meta.params["digits"] or (4, 2)
+            local_mapper[Decimal] = partial(
+                RandomBuilder.next_decimal, precision, scale
+            )
+        elif column.value_type == datetime.datetime:
+            tz_aware = getattr(column, "tz_aware", False)
+            local_mapper[datetime.datetime] = partial(
+                RandomBuilder.next_datetime, tz_aware
+            )
+        elif column.value_type == str:
+            if _length := column._meta.params.get("length"):
+                local_mapper[str] = partial(RandomBuilder.next_str, _length)
+        return local_mapper
+
+    @classmethod
+    def _get_other_mapper(cls, column: Column) -> t.Dict[t.Type, t.Callable]:
+        """
+        This is a hook that allows users to register their own random type
+        callable. If the callable has a parameter named `column`, we assist
+        by injecting `column` using `partial`.
+
+        :param column:
+            Column class to randomize.
+
+        Examples::
+
+            # a callable not utilizing column information
+            ModelBuilder.register_random_type(str, lambda: "piccolo")
+
+            # a callable utilizing the column information
+            def next_str(column: Column) -> str:
+                length = column._meta.params.get("length", 5)
+                return "".join("a" for _ in range(length))
+            )
+            ModelBuilder.register_random_type(str, next_str)
+
+        """
+        other_mapper: t.Dict[t.Type, t.Callable] = {}
+        for typ, callable_ in cls.__OTHER_MAPPER.items():
+            sig = inspect.signature(callable_)
+            if sig.parameters.get("column"):
+                other_mapper[typ] = partial(callable_, column)
+            else:
+                other_mapper[typ] = callable_
+        return other_mapper
+
+    @classmethod
+    def register_random_type(cls, typ: t.Type, callable_: t.Callable) -> None:
+        cls.__OTHER_MAPPER[typ] = callable_
+
+    @classmethod
+    def unregister_random_type(cls, typ: t.Type) -> None:
+        if typ in cls.__OTHER_MAPPER:
+            del cls.__OTHER_MAPPER[typ]
--- a/piccolo/testing/model_builder.py
+++ b/piccolo/testing/model_builder.py
@@ -1,31 +1,15 @@
 from __future__ import annotations
 
-import datetime
 import json
 import typing as t
-from decimal import Decimal
-from uuid import UUID
 
-from piccolo.columns import JSON, JSONB, Array, Column, ForeignKey
+from piccolo.columns import JSON, JSONB, Column, ForeignKey
 from piccolo.custom_types import TableInstance
 from piccolo.testing.random_builder import RandomBuilder
 from piccolo.utils.sync import run_sync
 
 
 class ModelBuilder:
-    __DEFAULT_MAPPER: t.Dict[t.Type, t.Callable] = {
-        bool: RandomBuilder.next_bool,
-        bytes: RandomBuilder.next_bytes,
-        datetime.date: RandomBuilder.next_date,
-        datetime.datetime: RandomBuilder.next_datetime,
-        float: RandomBuilder.next_float,
-        int: RandomBuilder.next_int,
-        str: RandomBuilder.next_str,
-        datetime.time: RandomBuilder.next_time,
-        datetime.timedelta: RandomBuilder.next_timedelta,
-        UUID: RandomBuilder.next_uuid,
-    }
-
     @classmethod
     async def build(
         cls,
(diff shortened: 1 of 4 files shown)


dantownsend (commented):
This is great - thanks 👍 

`ModelBuilder` is something which started fairly simple, but as more and more edge cases were discovered it has become quite complex.

I agree that your solution is cleaner and easier to follow.

I left a few comments.

piccolo/testing/random_builder.py
```diff
@@ -43,12 +81,18 @@ def next_enum(cls, e: t.Type[enum.Enum]) -> t.Any:
...
+        return cls.next_float(maximum=10 ** (precision - scale), scale=scale)
+
     @classmethod
     def next_int(cls, minimum=0, maximum=2147483647) -> int:
         return random.randint(minimum, maximum)
 
     @classmethod
-    def next_str(cls, length=16) -> str:
+    def next_str(cls, length: int | None = 16) -> str:
+        length = length or 16
```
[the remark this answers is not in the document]
jrycw: I was uncertain yesterday whether `column._meta.params.get("length")` would return `0` or `None`. However, regardless of whether it's `0` or `None`, we can still provide the default value (again...).

piccolo/testing/random_builder.py
```diff
@@ -43,12 +81,18 @@ def next_enum(cls, e: t.Type[enum.Enum]) -> t.Any:
     def next_float(cls, minimum=0, maximum=2147483647, scale=5) -> float:
         return round(random.uniform(minimum, maximum), scale)
 
+    @classmethod
+    def next_decimal(cls, digits: t.Tuple[int, int] | None = (4, 2)) -> float:
+        precision, scale = digits or (4, 2)
+        return cls.next_float(maximum=10 ** (precision - scale), scale=scale)
```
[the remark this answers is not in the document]
jrycw: Could this be related to the precision of the Decimal type? For instance:
```python=
>>> Decimal(100.1)
Decimal('100.099999999999994315658113919198513031005859375')
```

piccolo/testing/random_builder.py
```diff
@@ -1,12 +1,50 @@
...
+            datetime.timedelta: cls.next_timedelta,
+            UUID: cls.next_uuid,
+        }
+
+        random_value_callable = mapper.get(column.value_type)
+        if random_value_callable is None:
+            random_value_callable = partial(
+                cls.next_list,
+                mapper[t.cast(Array, column).base_column.value_type],
+            )
```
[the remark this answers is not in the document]
jrycw: The handling of `list` is quite complex, involving two aspects: the simple callable, such as `RandomBuilder.next_bool`, and additional logic that requires knowledge of the `column` to manufacture another callable. Since `list` relies on obtaining the type from the `mapper`, we cannot determine the exact type until the last moment.

piccolo/testing/random_builder.py
```diff
@@ -1,12 +1,50 @@
...
+from decimal import Decimal
+from functools import partial
+from uuid import UUID
+
+from piccolo.columns import Array, Column
 
 
 class RandomBuilder:
+    @classmethod
+    def _build(cls, column: Column) -> t.Any:
```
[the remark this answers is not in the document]
jrycw: I initially considered treating it as a private method, but also wanted to find a way to provide a mechanism for registering a new type.

jrycw:
> This is great - thanks 👍
> 
> `ModelBuilder` is something which started fairly simple, but as more and more edge cases were discovered it has become quite complex.
> 
> I agree that your solution is cleaner and easier to follow.
> 
> I left a few comments.

@dantownsend I completely agree with you; the task turned out to be more complex than I initially anticipated. Building the `mapper` required careful handling of many cases, such as using `partial`, dealing with the return from `get`, and modifying the current `next-ish` method with default values to becoming a callable. While this PR is functional at the moment, I acknowledge that it may become challenging to maintain in the future. I am currently exploring alternative ideas and hope to provide another version soon.

piccolo/testing/random_builder.py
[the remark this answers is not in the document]
dantownsend: Good point. Decimals are so weird. I wonder if instead we can create a decimal using a string.

Some pseudo code, where we get two random integers - one for before the decimal point, and one for after the decimal point:

```python
decimal.Decimal(f'{get_some_number()}.{get_some_other_number()}')
```

Rather than passing a float to `Decimal`.

piccolo/testing/random_builder.py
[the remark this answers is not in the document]
dantownsend: Good point.

piccolo/testing/random_builder.py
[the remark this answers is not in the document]
dantownsend: You're right - arrays are always tricky edge cases.

Maybe we could add a check if `isinstance(column, Array)`?

piccolo/testing/random_builder.py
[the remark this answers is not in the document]
dantownsend: That would be nice.

dantownsend:
> @dantownsend I completely agree with you; the task turned out to be more complex than I initially anticipated. Building the mapper required careful handling of many cases, such as using partial, dealing with the return from get, and modifying the current next-ish method with default values to becoming a callable. While this PR is functional at the moment, I acknowledge that it may become challenging to maintain in the future. I am currently exploring alternative ideas and hope to provide another version soon.

I think there's a lot of potential in this approach. Feel free to play around with other ideas, but this is a nice start.

jrycw:
> > @dantownsend I completely agree with you; the task turned out to be more complex than I initially anticipated. Building the mapper required careful handling of many cases, such as using partial, dealing with the return from get, and modifying the current next-ish method with default values to becoming a callable. While this PR is functional at the moment, I acknowledge that it may become challenging to maintain in the future. I am currently exploring alternative ideas and hope to provide another version soon.
> 
> I think there's a lot of potential in this approach. Feel free to play around with other ideas, but this is a nice start.

I've just pushed another version. This version introduces a hook for users to register their own random type. I'll review your comments tomorrow as it's already midnight here in Asia.

piccolo/testing/random_builder.py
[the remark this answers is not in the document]
jrycw: Fantastic suggestion, indeed!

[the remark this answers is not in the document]
jrycw: Maybe like:
```python
class RandomBuilder:
    ...
    @classmethod
    def next_float(cls, minimum=0, maximum=2147483647, scale=5) -> float:
        return round(random.uniform(minimum, maximum), scale)

    @classmethod
    def next_decimal(
        cls, precision: int = 4, scale: int = 2
    ) -> decimal.Decimal:
        float_number = cls.next_float(
            maximum=10 ** (precision - scale), scale=scale
        )
        return decimal.Decimal(str(float_number))
```

jrycw:
The third version may be a bit easier to understand:
* `RandomBuilder` now offers a public API called `get_mapper`, enabling users to access the default random mapper.
* Forming the `type-callable` pairs into a single dictionary involves several steps, which are encapsulated in the public API `ModelBuilder.get_registry`:
  * Initially, we import information from `RandomBuilder.get_mapper`, a one-time operation.
  * Next, we address situations where callables may require information from the `column` in `ModelBuilder._get_local_mapper`.
  * We then incorporate user-registered types into `ModelBuilder._get_other_mapper`.
  * Finally, we integrate the logic for `list` into `reg = { **default_mapper, **cls._get_local_mapper(column), **cls._get_other_mapper(column)}`, returning reg as a `MappingProxyType`.
* I made an effort to comprehend the relationship between `list` and `Array`, but I'm afraid I may have gotten a bit lost in the codebase. I'm uncertain about the extent to which `isinstance(column, Array)` can handle the edge case effectively.
* Handling `enum.Enum` remains a challenge.
* I made a minor adjustment in `ModelBuilder._build`, where `defaults = defaults or { }` seems clearer to me.

piccolo/testing/random_builder.py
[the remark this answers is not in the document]
jrycw: See my latest comment regarding this issue.

piccolo/testing/model_builder.py
```diff
@@ -159,29 +151,110 @@ def _randomize_attribute(cls, column: Column) -> t.Any:
...
+
+        :param column:
+            Column class to randomize.
+        """
+        local_mapper: t.Dict[t.Type, t.Callable] = {}
+
+        precision, scale = column._meta.params.get("digits") or (4, 2)
+        local_mapper[Decimal] = partial(
+            RandomBuilder.next_decimal, precision, scale
+        )
```
jrycw, line 209: If `RandomBuilder.next_decimal` incorporates default values for precision and scale (like this PR), the logic can be simplified to:
```python
        if precision_scale := column._meta.params.get("digits"):
            local_mapper[Decimal] = partial(
                RandomBuilder.next_decimal, *precision_scale
            )
```

piccolo/testing/model_builder.py
```diff
@@ -159,29 +151,110 @@ def _randomize_attribute(cls, column: Column) -> t.Any:
...
-            tz_aware = getattr(column, "tz_aware", False)
-            random_value = RandomBuilder.next_datetime(tz_aware=tz_aware)
-        elif column.value_type == list:
-            length = RandomBuilder.next_int(maximum=10)
-            base_type = t.cast(Array, column).base_column.value_type
-            random_value = [
-                cls.__DEFAULT_MAPPER[base_type]() for _ in range(length)
-            ]
-        elif column._meta.choices:
+        reg = cls.get_registry(column)
```
dantownsend, line 154: This can probably go into the `else` block, as we don't use it if the column has choices.

piccolo/testing/model_builder.py
```diff
@@ -159,29 +151,110 @@ def _randomize_attribute(cls, column: Column) -> t.Any:
...
+        reg = {
+            **default_mapper,
+            **cls._get_local_mapper(column),
+            **cls._get_other_mapper(column),
+        }
+
+        if column.value_type == list:
+            reg[list] = partial(
+                RandomBuilder.next_list,
+                reg[t.cast(Array, column).base_column.value_type],
```
dantownsend, line 191: I don't want to make things too insane, but multidimensional arrays are possible:

```python
Array(Array(Integer())
```

I wonder what would happen here in that situation?

piccolo/testing/model_builder.py
```diff
@@ -159,29 +151,110 @@ def _randomize_attribute(cls, column: Column) -> t.Any:
...
+        the complete registry for the specified column.
+
+        :param column:
+            Column class to randomize.
+
+        """
+        default_mapper = cls.__DEFAULT_MAPPER
+        if not default_mapper:  # execute once only
+            for typ, callable_ in RandomBuilder.get_mapper().items():
+                default_mapper[typ] = callable_
```
dantownsend, line 179: I like that some of the work is being cached - one of my concerns with the previous implementation was the work required to set up the mapping each time.

piccolo/testing/model_builder.py
```diff
@@ -159,29 +151,110 @@ def _randomize_attribute(cls, column: Column) -> t.Any:
...
+            else:
+                other_mapper[typ] = callable_
+        return other_mapper
+
+    @classmethod
+    def register_type(cls, typ: t.Type, callable_: t.Callable) -> None:
+        cls.__OTHER_MAPPER[typ] = callable_
+
+    @classmethod
+    def unregister_type(cls, typ: t.Type) -> None:
```
dantownsend, line 258: Being able to register custom types is cool, but I wonder about the main use cases.

You can specify defaults at the moment:

```python
await ModeBuilder.build(MyTable, defaults={MyTable.some_column: "foo"})
```

I'm not against being able to override how types are handled, but we just need to articulate to users when it's appropriate vs using `defaults`.

piccolo/testing/random_builder.py
```diff
@@ -43,12 +65,21 @@ def next_enum(cls, e: t.Type[enum.Enum]) -> t.Any:
...
         return round(random.uniform(minimum, maximum), scale)
 
+    @classmethod
+    def next_decimal(
+        cls, precision: int = 4, scale: int = 2
+    ) -> decimal.Decimal:
+        float_number = cls.next_float(
+            maximum=10 ** (precision - scale), scale=scale
+        )
+        return decimal.Decimal(str(float_number))
```
dantownsend, line 75: Clever!

piccolo/testing/model_builder.py
jrycw: Thank you for bringing this situation to my attention. I was curious about the behavior in our current codebase, so I conducted a quick test. It seems that the current implementation will throw a `KeyError` for this situation (please correct me if I'm mistaken).

piccolo/testing/model_builder.py
jrycw: That's a valid observation. It seems that the presence of `reg` is a result of the requirements in the previous version, where it was needed for multiple `elif-else` blocks.

piccolo/testing/model_builder.py
jrycw: Thank you for bringing up this question, it prompted me to reflect on the code and its implications.

The main difference between using `default` and `register_type` lies in their respective purposes:
* Utilizing `default` is suitable when users generally find our provided `random logic` satisfactory, but they require a hardcoded value for a specific column on a one-time basis.
* On the other hand, employing `register_type` is appropriate when users desire a custom implementation for a type, effectively overwriting our default `random logic` for that specific type.
Here are three distinct use cases:
  * (1) Types not provided by us: For instance, the type like `next_decimal` is not available in the current release version, but users can implement their own logic for the type and inject it into the `ModelBuilder`.
  * (2) User preference for specific logic: In scenarios where we introduce new features, such as the shiny `next_decimal` logic (returning `decimal.Decimal` if `column.value_type` is `decimal.Decimal`), users may prefer the previous implementation or have specific requirements. With `register_type`, they have the flexibility to override the default behavior.
  * (3) Unanticipated user cases: This aspect is particularly valuable for registration. For example, consider a user who initially builds a successful e-commerce platform in the UK using `Piccolo`. Later, they expand into Asia and encounter legal requirements necessitating the storage of customer names in local languages. If `ModelBuilder` does not support non-English characters, users can register their own implementations to address this issue.

A draft test for situations (1) and (2) might look like this:
```python
class TableWithDecimal(Table):
    numeric = Numeric()
    numeric_with_digits = Numeric(digits=(4, 2))
    decimal = Decimal()
    decimal_with_digits = Decimal(digits=(4, 2))

class TestModelBuilder(unittest.TestCase):
    ...
    def test_registry_overwritten1(self):
        table = ModelBuilder.build_sync(TableWithDecimal)
        for key, value in table.to_dict().items():
            if key != "id":
                self.assertIsInstance(value, decimal.Decimal)

        def fake_next_decimal(column: Column) -> float:
            """will return `float` instead of `decimal.Decimal`"""
            precision, scale = column._meta.params["digits"] or (4, 2)
            return RandomBuilder.next_float(
                maximum=10 ** (precision - scale), scale=scale
            )

        ModelBuilder.register_type(decimal.Decimal, fake_next_decimal)
        overwritten_table = ModelBuilder.build_sync(TableWithDecimal)
        for key, value in overwritten_table.to_dict().items():
            if key != "id":
                self.assertIsInstance(value, float)
```
A draft test for situations (3) might look like this:
```python
class TestModelBuilder(unittest.TestCase):
    ...
    def test_registry_overwritten2(self):
        choices = "一二三"  # Chinese characters

        def next_str(length: int = 3) -> str:
            # Chinese names often consist of three Chinese characters
            return "".join(random.choice(choices) for _ in range(length))

        ModelBuilder.register_type(str, next_str)
        manager1 = ModelBuilder.build_sync(Manager)
        self.assertTrue(all(char_ in choices for char_ in manager1.name))

        poster1 = ModelBuilder.build_sync(Poster)
        self.assertTrue(all(char_ in choices for char_ in poster1.content))

        ModelBuilder.unregister_type(str)
        manager2 = ModelBuilder.build_sync(Manager)
        self.assertTrue(all(char_ not in choices for char_ in manager2.name))

        poster2 = ModelBuilder.build_sync(Poster)
        self.assertTrue(all(char_ not in choices for char_ in poster2.content))
```
The scenario is as follows: `Manager1` is a locally hired individual, while `Manager2` is dispatched from the UK. Both are working on the `poster` using their native languages.

Finally, I realized I had overlooked the magic behavior of Python's name mangling rules. For instance:
```python
>>> class ModelBuilder:
...     __OTHER_MAPPER = {}
...
>>> ModelBuilder.__OTHER_MAPPER
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: type object 'ModelBuilder' has no attribute '__OTHER_MAPPER'
>>> ModelBuilder._ModelBuilder__OTHER_MAPPER
{}
```
As a result, the previous test code might be a bit off. I need to use the following code for the setup and teardown phases for each test:
```python
    def setUp(self) -> None:
        ModelBuilder._ModelBuilder__OTHER_MAPPER.clear()  # type: ignore

    def tearDown(self) -> None:
        ModelBuilder._ModelBuilder__OTHER_MAPPER.clear()  # type: ignore
```

piccolo/testing/model_builder.py
jrycw: I'm not sure if this is a good idea.
```python    
    @classmethod
    def _randomize_attribute(cls, column: Column) -> t.Any:
        reg = cls.get_registry(column)
        random_value = reg.get(enum.Enum, reg[column.value_type])()
        if isinstance(column, (JSON, JSONB)):
            return json.dumps({"value": random_value})
        return random_value


    @classmethod
    def _get_local_mapper(cls, column: Column) -> t.Dict[t.Type, t.Callable]:
	...
        if _choices := column._meta.choices:
            local_mapper[enum.Enum] = partial(
                RandomBuilder.next_enum, _choices)

        return local_mapper
```

piccolo/testing/model_builder.py
jrycw: This might be a bit off-topic, but I wanted to try out this behavior in the `Piccolo playground`, and it doesn't seem to be working. The code works fine if I just launch a terminal and enter the shell.
```
piccolo playground run --engine=sqlite3
```

```python
In [1]: from piccolo.table import Table

In [2]: from piccolo.columns import Array, BigInt

In [3]: class MyTable(Table):
   ...:     my_column = Array(Array(BigInt()))
   ...:
---------------------------------------------------------------------------
NameError                                 Traceback (most recent call last)
Cell In[3], line 1
----> 1 class MyTable(Table):
      2     my_column = Array(Array(BigInt()))

Cell In[3], line 2, in MyTable()
      1 class MyTable(Table):
----> 2     my_column = Array(Array(BigInt()))

NameError: name 'Array' is not defined
```

sinisaos: @jrycw This happens because the `Array` and `BigInt` columns are not imported into the [Playground application](https://github.com/piccolo-orm/piccolo/blob/master/piccolo/apps/playground/commands/run.py#L13-L23). If you patch the local Piccolo installation and add these two columns, everything works fine. Maybe should add all the possible columns to import and that would solve the problem..

jrycw: @sinisaos Thank you very much for identifying the source of the issue and providing the solution. It's working now.

dantownsend: You're right that the current implementation of `ModelBuilder` doesn't handle multidimensional arrays, so I wouldn't worry about it if it's a tricky fix. `Array` columns now have two methods to help with this kind of thing: `_get_dimensions` and `_get_inner_value_type`,

piccolo/testing/model_builder.py
dantownsend: Thanks for explaining the rationale behind it - it makes sense.

If I was to completely redesign `ModelBuilder`, I probably wouldn't have class methods. Instead of:

```python
await ModelBuilder.build(MyTable)
```

I would have:

```python
await ModelBuilder(some_option=True).build(MyTable)
```

So we can configure ModelBuilder's behaviour easier. For registering types we could have:

```python
custom_builder = ModelBuilder(types={...})

await custom_builder.build(MyTable)
```

We could allow the types to be passed in via the `build` method instead:

```python
await ModelBuilder.build(MyTable, types={...})
```

If `register` and `unregister` work globally, there are pros and cons. The main pro is you only need to set it up once (e.g. in a session fixture of Pytest). But if you were to somehow run your tests in parallel, it might be problematic.

What do you think?

jrycw: Thank you for sharing your thoughts with me. Personally, I am inclined towards the instance method approach. However, implementing this change might break the current interface. I propose a three-stage transition plan:

1. First stage: Utilize the concept of `descriptors` to distinguish between calls from the `class` or `instance`. Initially, we move the current implementation to the `class` branch to maintain user experience. Simultaneously, we start implementing the new concept in the `instance` branch, issuing an `experimental warning`.

2. Second stage: If the new concept gains appreciation from users or developers, we add a `deprecated warning` to the `class` branch.

3. Third stage: Remove the `class` branch and clean up the code to ensure all methods are `instance methods` by the end.

During the first two stages, we'll keep the `class` branch unchanged and encourage users to try out the new syntax and the new features. If we reach the third stage, users who prefer the `class` branch might need to adjust their habits from using `await ModelBuilder.build(...)` to `await ModelBuilder().build()` or `ModelBuilder.build_sync(...)` to `ModelBuilder().build_sync(...)`.


The concept of `descriptors` is relatively straightforward, but it can sometimes feel too magical to grasp. I often need a refresher before coding if I haven't touched it for a long time. Fortunately, we don't need the complex `__get__` and `__set__` logic for the `data descriptor`. A simple `non-data descriptor` should suffice for our use case. With the help of this [post](https://stackoverflow.com/questions/28237955/same-name-for-classmethod-and-instancemethod),  I've drafted a concept code as follows:
```python
import asyncio
import inspect
import typing as t
from concurrent.futures import ThreadPoolExecutor


def run_sync(coroutine: t.Coroutine):
    try:
        # We try this first, as in most situations this will work.
        return asyncio.run(coroutine)
    except RuntimeError:
        # An event loop already exists.
        with ThreadPoolExecutor(max_workers=1) as executor:
            future = executor.submit(asyncio.run, coroutine)
            return future.result()


class dichotomy:
    def __init__(self, f):
        self.f = f

    def __get__(self, instance, owner):
        cls_or_inst = instance if instance is not None else owner
        if inspect.iscoroutine(self.f):
            async def newfunc(*args, **kwargs):
                return await self.f(cls_or_inst, *args, **kwargs)
        else:
            def newfunc(*args, **kwargs):
                return self.f(cls_or_inst, *args, **kwargs)
        return newfunc


class ModelBuilder:
    def __init__(self, *args, **kwargs):
        self._types = "..."  # Some information for instance method

    @dichotomy
    async def build(self_or_cls, *args, **kwargs):
        if inspect.isclass(self_or_cls):
            print("called as a class method from build")
            cls = self_or_cls
            await cls._build()
        else:
            print("called as an instance method from build")
            self = self_or_cls
            await self._build()

    @dichotomy
    def build_sync(self_or_cls, *args, **kwargs):
        return run_sync(self_or_cls.build())

    @dichotomy
    async def _build(self_or_cls, *args, **kwargs):
        if inspect.isclass(self_or_cls):
            print("called as a class method from _build", end="\n"*2)
            cls = self_or_cls  # noqa: F841
            # Current implementation remains here.
        else:
            print("called as an instance method from _build")
            self = self_or_cls
            # Some information can be retrieved.
            print(f'{self._types=}', end="\n"*2)
            # Our new logics


async def main():
    print('Async ModelBuilder.build: ')
    await ModelBuilder.build()

    print('Async ModelBuilder().build: ')
    await ModelBuilder().build()

    print('Sync ModelBuilder.build: ')
    ModelBuilder.build_sync()

    print('Sync ModelBuilder().build: ')
    ModelBuilder().build_sync()

if __name__ == '__main__':
    asyncio.run(main())
```
```
Async ModelBuilder.build: 
called as a class method from build
called as a class method from _build

Async ModelBuilder().build: 
called as an instance method from build
called as an instance method from _build
self._types='...'

Sync ModelBuilder.build: 
called as a class method from build
called as a class method from _build

Sync ModelBuilder().build: 
called as an instance method from build
called as an instance method from _build
self._types='...'
```
Finally, I agree that making `register` and `unregister` work globally could make it challenging to verify test results in parallel scenarios. I might lean towards using `instance methods` for the registering issue again.

These are just rough ideas that came to mind. I'm open to further discussions and refinements.

dantownsend: Using descriptors is an interesting idea. I've used them sparingly before - as you say, they're very powerful, but can be confusing.

There's a lot of really good ideas in this PR, and I don't want to bog things down. I wonder if we could add this in a subsequent PR.

jrycw: Certainly! Here are some options to consider for closing this PR:

* Closing the PR without merging any changes.
* Keeping the current code as is, while implementing the `next_enum` functionality and updating related code.
* Utilizing the latest commit of this PR while removing the option for users to register custom types.
* Merging the PR with its latest commit.
* Considering any other suggestions or alternatives.

I'm open to any of these choices. @dantownsend , what are your thoughts on this matter?

dantownsend: @jrycw  Sorry for the delay on this - I haven't forgotten about it, I'm just trying to finish off a couple of PRs.

I'm interested to know your thoughts on this:

https://github.com/piccolo-orm/piccolo/pull/978

If you think it's a good idea or not.

jrycw: @dantownsend no worries at all. It's great to see the project progressing on various fronts. I'll make an effort to review it and share any opinions or feedback I may have.

PR #972, opened 2024-03-22
update readme for Esmerald

sinisaos:
Update `readme.md` as Esmerald is missing as a supported web framework.


dantownsend:
Looks good, thanks!

PR #978, opened 2024-04-05
`Timestamptz` autoconversion

dantownsend:
Based off https://github.com/piccolo-orm/piccolo/pull/959

Created this branch for testing. Experimenting with the `AT TIME ZONE` clause.

Remaining tasks:

- [x] Get `select` and `objects` queries working
- [x] Get SQLite working
- [x] Fix existing tests
- [ ] Add docs for `at_time_zone` method
- [ ] Add tests for `select` queries
- [ ] Add tests for the `at_time_zone` method
- [x] Optimise `Objects._get_select_query` a bit
- [ ] Add an `at_time_zone` method to `objects`


dantownsend:
I think this is almost done now. `select` queries work fine. But surprisingly `object` queries override `response_handler` from `select`, so isn't currently working. It's just a matter of shuffling some logic around.

dantownsend:
Now we need to work out what to do with SQLite - I might just raise a `ValueError` for now.

sinisaos:
@dantownsend I tried your solution and everything worked great. Here is a Piccolo `shell` results
```python
# if we declare a time zone in the table definition
# class TallinnConcerts(Table):
#   event_start = Timestamptz(at_time_zone=ZoneInfo("Europe/Tallinn"))
In [1]: await TallinnConcerts.select()
Out[1]: 
[{'id': 1,
  'event_start': datetime.datetime(2024, 4, 5, 10, 44, 46, tzinfo=zoneinfo.ZoneInfo(key='Europe/Tallinn'))},
 {'id': 2,
  'event_start': datetime.datetime(2024, 4, 5, 10, 44, 46, tzinfo=zoneinfo.ZoneInfo(key='Europe/Tallinn'))}]

In [2]: [i.to_dict() for i in await TallinnConcerts.objects()]
Out[2]: 
[{'id': 1,
  'event_start': datetime.datetime(2024, 4, 5, 10, 44, 46, tzinfo=zoneinfo.ZoneInfo(key='Europe/Tallinn'))},
 {'id': 2,
  'event_start': datetime.datetime(2024, 4, 5, 10, 44, 46, tzinfo=zoneinfo.ZoneInfo(key='Europe/Tallinn'))}]

# if we do not declare a timezone in the table definition, we can define `at_time_zone` in the  select query
# class TallinnConcerts(Table):
#   event_start = Timestamptz()
In [3]: await TallinnConcerts.select(TallinnConcerts.event_start.at_time_zone("Europe/Tallinn"))
Out[3]: 
[{'event_start': datetime.datetime(2024, 4, 6, 16, 37, 33, tzinfo=zoneinfo.ZoneInfo(key='Europe/Tallinn'))},
 {'event_start': datetime.datetime(2024, 4, 6, 16, 37, 33, tzinfo=zoneinfo.ZoneInfo(key='Europe/Tallinn'))}]

```
Piccolo Admin also work.

dantownsend:
@sinisaos Thanks a lot for testing it.

What do you think about the design of it?

I changed it to be:

```python
Timestamptz(at_time_zone=ZoneInfo("America/New_York"))
```

So the argument is `at_time_zone`, instead of `tz` to be consistent with the `at_time_zone` method name, and to make it clearer what it does.

I was thinking of adding a method to the `objects` query too. For select we can do this:

```python
await Concert.select(Concert.starts_at.at_time_zone('America/Los_Angeles'))
```

But there's currently no way of overriding it for `objects` queries. Maybe something like:

```python
await Concert.objects().at_time_zone({Concert.starts_at: 'America/Los_Angeles'})
```

sinisaos:
@dantownsend Everything looks and works great to me, as I wrote in the previous comment. 

> But there's currently no way of overriding it for `objects` queries. Maybe something like:
> 
> ```python
> await Concert.objects().at_time_zone({Concert.starts_at: 'America/Los_Angeles'})
> ```

It would be great to have that `objects` method to manipulate the results.

piccolo/columns/column_types.py
```diff
@@ -993,22 +997,59 @@ class Concert(Table):
...
     ) -> None:
         self._validate_default(
             default, TimestamptzArg.__args__  # type: ignore
         )
 
         if isinstance(default, datetime):
-            default = TimestamptzCustom.from_datetime(default)
+            default = TimestamptzCustom.from_datetime(default, at_time_zone)
 
         if default == datetime.now:
```
jrycw, line 1012: I'm a bit confused about this `if` statement. It seems that `datetime.now` is a callable, which doesn't seem to align with the `TimestamptzArg` from the signature. By the way, it seems that the test doesn't cover this scenario involving a callable.

requirements/requirements.txt
```diff
@@ -5,3 +5,5 @@ targ>=0.3.7
 inflection>=0.5.1
 typing-extensions>=4.3.0
 pydantic[email]==2.*
+tzdata>=2024.1
```
jrycw, line 8: Based on my experience with `zoneinfo` on `Windows` machines, it can exhibit some unexpected behavior even after installing [tzdata](https://docs.python.org/3/library/zoneinfo.html#data-sources). If `piccolo` aims to support cross-platform conditions, I suggest running continuous integration tests for timestamp-related functionality in the `Windows` container, at least for a period, once this PR is merged.

piccolo/columns/column_types.py
```diff
@@ -955,30 +956,33 @@ def __set__(self, obj, value: t.Union[datetime, None]):
...
 
-        class Concert(Table):
-            starts = Timestamptz()
+        class TallinnConcerts(Table):
+            event_start = Timestamptz(at_time_zone=ZoneInfo("Europe/Tallinn"))
 
         # Create
-        >>> await Concert(
-        ...    starts=datetime.datetime(
-        ...        year=2050, month=1, day=1, tzinfo=datetime.timezone.tz
```
jrycw, line 973: It seems that `tzinfo=datetime.timezone.tz` should be replaced with `tzinfo=datetime.timezone.utc` in the old documentation.

piccolo/columns/column_types.py
```diff
@@ -993,22 +997,59 @@ class Concert(Table):
...
+        return instance
+
+    ###########################################################################
+
+    def get_select_string(
+        self, engine_type: str, with_alias: bool = True
+    ) -> str:
+        select_string = self._meta.get_full_name(with_alias=False)
+
+        if self._at_time_zone != ZoneInfo("UTC"):
```
jrycw, line 1041: There might be other instances where timezone comparisons are performed similarly, and I'll use this one as an example. It seems this type of comparison could lead to unexpected outcomes. The following example might illustrate this unpredictability.
```python
>>> import datetime
>>> from zoneinfo import ZoneInfo
>>> tz1 = datetime.timezone.utc
>>> tz2 = ZoneInfo("UTC")
>>> d1 = datetime.datetime(year=2024, month=4, day=7, tzinfo=tz1)
>>> d2 = datetime.datetime(year=2024, month=4, day=7, tzinfo=tz2)
>>> d1 == d2
True
>>> str(d1) == str(d2)
True
>>> d1.tzinfo, d2.tzinfo
(datetime.timezone.utc, zoneinfo.ZoneInfo(key='UTC'))
>>> d1.tzinfo == d2.tzinfo
False
>>> d1.tzinfo is d2.tzinfo
False
```

piccolo/columns/column_types.py
```diff
@@ -993,22 +997,59 @@ class Concert(Table):
...
+        """
+        By default, the database returns the value in UTC. This lets us get
+        the value converted to the specified timezone.
+        """
+        time_zone = (
+            ZoneInfo(time_zone) if isinstance(time_zone, str) else time_zone
+        )
+        instance = self.copy()
+        instance._at_time_zone = time_zone
+        return instance
```
jrycw, line 1032: Regarding [#959](https://github.com/piccolo-orm/piccolo/pull/959#issuecomment-2040972001), it appears that we are creating a new object by copying `self` and then modifying the `_at_time_zone` attribute of that object to match the `time_zone` variable, without altering `self`. Is my understanding correct?

piccolo/columns/column_types.py
```diff
@@ -993,22 +997,59 @@ class Concert(Table):
...
+
+    def get_select_string(
+        self, engine_type: str, with_alias: bool = True
+    ) -> str:
+        select_string = self._meta.get_full_name(with_alias=False)
+
+        if self._at_time_zone != ZoneInfo("UTC"):
+            # SQLite doesn't support `AT TIME ZONE`, so we have to do it in
+            # Python instead (see ``Select.response_handler``).
+            if self._meta.engine_type in ("postgres", "cockroach"):
```
jrycw, line 1044: Occurrences of `in` or `==` checks like this one seem to be becoming increasingly common. I would recommend using data structures like `Enum` to mitigate typos and rely more on auto-completion while coding.

piccolo/columns/column_types.py
jrycw: Oops, it should be the line `if default == datetime.now`.

jrycw:
> @sinisaos Thanks a lot for testing it.
> 
> What do you think about the design of it?
> 
> I changed it to be:
> 
> ```python
> Timestamptz(at_time_zone=ZoneInfo("America/New_York"))
> ```
> 
> So the argument is `at_time_zone`, instead of `tz` to be consistent with the `at_time_zone` method name, and to make it clearer what it does.
> 
> I was thinking of adding a method to the `objects` query too. For select we can do this:
> 
> ```python
> await Concert.select(Concert.starts_at.at_time_zone('America/Los_Angeles'))
> ```
> 
> But there's currently no way of overriding it for `objects` queries. Maybe something like:
> 
> ```python
> await Concert.objects().at_time_zone({Concert.starts_at: 'America/Los_Angeles'})
> ```

I quite like this `polars-like` `select` expression (or should I say `polars` is `piccolo-like` 😊?), which feels natural to use. Additionally, I agree with @sinisaos that the syntax of the `objects` method looks good to me.

requirements/requirements.txt
jrycw: @dantownsend @sinisaos This might be a little off-topic, but have you guys heard about [uv](https://github.com/astral-sh/uv)? I've used it for a couple of personal projects and it looks promising to me.

sinisaos: @jrycw Sorry, I've never used `uv`. I always use `virtualenv` and I'm used to it.

Issue #945, opened 2024-03-11
Using `db_column_name` with auto migrations


dantownsend:
You can explicitly specify the database column name:

```python
class MyTable(Table):
    type_ = Varchar(db_column_name="type")
```

When using auto migrations to add such a column to an **existing table**, it seems like `db_column_name` isn't being used.

atkei:
@dantownsend 
I think dropping a column has `db_column_name` from an existing table does not work as well.

Perhaps need to set `db_column_name` instead of `column_name`.
https://github.com/piccolo-orm/piccolo/blob/master/piccolo/apps/migrations/auto/migration_manager.py#L668

PR #983, opened 2024-04-11
Fix adding and dropping column with db_column_name in auto migration

atkei:
Related to #945.


atkei:
I have updated my comment with my analysis.
I hope this helps with the review.

dantownsend:
@atkei That's very helpful, thanks 👍

Issue #994, opened 2024-05-24
Add support for `upper` and `lower`


dantownsend:
So we can do select queries like:

```python
await Band.select(Band.name.upper())
```

And in where clauses:

```python
await Band.select().where(Band.name.upper() == 'PYTHONISTAS')

await Band.select().where(Band.name.upper().is_in(["PYTHONISTAS", "RUSTACEANS"])
```

PR #995, opened 2024-05-24
994 Add support for `upper` and `lower`

dantownsend:
Resolves https://github.com/piccolo-orm/piccolo/issues/994


dantownsend:
Using this instead: https://github.com/piccolo-orm/piccolo/pull/996

PR #1052, opened 2024-07-16
Allow for subclasses in defaults allowed types

Bobronium:
These changes allow for extension of existing Default types, for instance, introducing UUID7 would now be possible just like this:
```python
from uuid import UUID

from piccolo import columns
from piccolo.columns.defaults import UUID4
from uuid_utils.compat import uuid7


class UUID7(UUID4):
    @property
    def postgres(self) -> str:
        return "uuid_generate_v7()"

    def python(self) -> UUID:
        return uuid7()


class User(Table, tablename="users"):
    id = columns.UUID(primary_key=True, default=UUID7())

### DIFF
--- a/piccolo/columns/base.py
+++ b/piccolo/columns/base.py
@@ -527,7 +527,7 @@ class Column(Selectable):
         elif (
             default is None
             and None in allowed_types
-            or type(default) in allowed_types
+            or isinstance(default, tuple(allowed_types))
         ):
             self._validated = True
             return True
@@ -540,7 +540,7 @@ class Column(Selectable):
                 self._validated = True
                 return True
         elif (
-            isinstance(default, Enum) and type(default.value) in allowed_types
+            isinstance(default, Enum) and isinstance(default.value, tuple(allowed_types))
         ):
             self._validated = True
             return True

### CODE REVIEW

**Comment by dantownsend** [comment]
@Bobronium Thanks for this.

Do you know the status of `uuid7` in Postgres - presumably you're using this extension: https://pgxn.org/dist/pg_uuidv7/

**Comment by Bobronium** [comment]
Actually, I'm just defining custom function for now ✨ 

<details>
<summary>Example migration</summary>

```python
from piccolo.apps.migrations.auto import MigrationManager

ID = "2024-07-13T21:20:34:913115"
VERSION = "1.13.1"
DESCRIPTION = "ADD UUIDv7"
UUID_GENERATE_V7_SQL = """\
-- Function to generate new v7 UUIDs.
-- Would be better off using https://github.com/fboulnois/pg_uuidv7, but can't on RDS.
-- Or, once the UUIDv7 spec is finalized, it will probably make it into the 'uuid-ossp' extension
-- and a custom function will no longer be necessary.
create or replace function uuid_generate_v7()
returns uuid
as $$
declare
  unix_ts_ms bytea;
  uuid_bytes bytea;
begin
  unix_ts_ms = substring(int8send(floor(extract(epoch from clock_timestamp()) * 1000)::bigint) from 3);
  uuid_bytes = uuid_send(gen_random_uuid());
  uuid_bytes = overlay(uuid_bytes placing unix_ts_ms from 1 for 6);
  uuid_bytes = set_byte(uuid_bytes, 6, (b'0111' || get_byte(uuid_bytes, 6)::bit(4))::bit(8)::int);
  return encode(uuid_bytes, 'hex')::uuid;
end
$$
language plpgsql
volatile;"""  # noqa: E501


class RawTable(Table):
    pass


async def add_uuid_v7() -> None:
    await RawTable.raw(UUID_GENERATE_V7_SQL)


async def forwards() -> None:
    manager = MigrationManager(migration_id=ID, app_name="app", description=DESCRIPTION)

    manager.add_raw(add_uuid_v7)
```

noamsto:
Will it be merged?
I want to use UUID7 as well...

Bobronium:
@dantownsend, friendly ping. This should be ready to be merged, unless you have any suggestions or a different implementation in mind?

Issue #1053, opened 2024-07-21
Refactor `_process_results` to support new DB engines


dantownsend:
> I'm one of the developers behind psqlpy. We've created a third-party library, which you can check out [here](https://github.com/qaspen-python/psqlpy-piccolo). However, we've encountered a small problem that makes the integration nearly impossible without your assistance.
> 
> The issue lies within the `_process_results` method, where some unwrapping is performed. Asyncpg returns a Record object, but psqlpy does not. Instead, it behaves more like SQLite in this context, returning a list of dictionaries. Therefore, we don't need this unwrapping. Unfortunately, there is no appropriate engine type for our use case. We cannot use the SQLite engine type for obvious reasons.
> 
> Would it be possible to add a `psqlpy-postgres` type, so we can eliminate this Record object unwrapping?
> 
> Additionally, this change needs to be reflected in the Query class within the `querystrings` property. There might be other places affected, but these are what I identified during my reverse engineering.
> 
> Looking forward to our collaboration!
> 

_Originally posted by @insani7y in https://github.com/piccolo-orm/piccolo/issues/986#issuecomment-2241261588_
            
As outlined above, there's a bit of code where we transform the response from the database into a list of dicts:

https://github.com/piccolo-orm/piccolo/blob/e7b13fbe35d2ed4723c23e049dd5f5b12500513e/piccolo/query/base.py#L51-L56

It assumes that Postgres engines will just be `asyncpg` based, which is no longer the case.

We should move this logic to the `Engine` class.

PR #1054, opened 2024-07-21
Refactor `_process_results` to support new DB engines

dantownsend:
Resolves https://github.com/piccolo-orm/piccolo/issues/1053

### DIFF
--- a/piccolo/columns/base.py
+++ b/piccolo/columns/base.py
@@ -250,11 +250,11 @@ class ColumnMeta:
 
         if self.call_chain:
             column_name = (
-                "$".join(
+                ".".join(
                     t.cast(str, i._meta.db_column_name)
                     for i in self.call_chain
                 )
-                + f"${column_name}"
+                + f".{column_name}"
             )
 
         return column_name
@@ -291,7 +291,7 @@ class ColumnMeta:
                 'band$manager.name'
 
                 >>> Band.manager.name._meta.get_full_name(with_alias=True)
-                'band$manager.name AS "manager$name"'
+                'band$manager.name AS "manager.name"'
 
         :param include_quotes:
             If you're using the name in a SQL query, each component needs to be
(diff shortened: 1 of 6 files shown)


piccolo/columns/base.py
```diff
@@ -250,11 +250,11 @@ def get_default_alias(self):
...
         if self.call_chain:
             column_name = (
-                "$".join(
+                ".".join(
                     t.cast(str, i._meta.db_column_name)
                     for i in self.call_chain
                 )
-                + f"${column_name}"
+                + f".{column_name}"
             )
```
dantownsend, line 258: There's a small risk changing this.

In the past, the following query:

```python
>>> await Band.select(Band.name, Band.manager.name)
[{'name': 'Pythonistas', 'manager.name': 'Guido'},
 {'name': 'Rustaceans', 'manager.name': 'Graydon'},
 {'name': 'C-Sharps', 'manager.name': 'Anders'}]
```

Executed this SQL:

```python
>>> print(Band.select(Band.name, Band.manager.name))
"""
SELECT ALL
    "band"."name" AS "name", 
    "band$manager"."name" AS "manager$name"    <- NOTE THIS
FROM "band"
    LEFT JOIN "manager" "band$manager"
    ON ("band"."manager" = "band$manager"."id")
"""
```

Note how we have`$` in the aliases - `AS "manager$name"`.

`$` was chosen because it's incredibly unlikely that someone would use `$` in their column names, and it still works even without the quotes around it:

```sql
SELECT name AS foo$bar from band;
-- Returns [{"foo$bar": "Pythonistas"}, ...]
```

We can use `.` instead, as long as we're careful to wrap all of our aliases in quotes (otherwise the query fails).

```sql
SELECT name AS foo.bar from band;
-- Error!
```

So our query now becomes:

```sql
SELECT ALL
    "band"."name" AS "name", 
    "band$manager"."name" AS "manager.name"    -- <- NOTE THIS
FROM "band"
    LEFT JOIN "manager" "band$manager"
    ON ("band"."manager" = "band$manager"."id")
```

This means we don't have to convert from `manager$name` to `manager.name` in Piccolo, which saves some overhead.

I checked the codebase, and it looks like we wrap all aliases in double quotes. If there's an edge case where we don't, it's something we'll have to fix.

PR #1065, opened 2024-08-08
Add support for SELECT FOR UPDATE 

### LINKED ISSUE #932: Pessimistic locking:  select for update
I need to lock an object in the database in order to prevent other transactions from modifying the same record. Within the Postgres such a lock should be done using the SELECT FOR UPDATE query.

How I can do this using Piccolo?

you can find it out in Django [here](https://docs.djangoproject.com/en/5.0/ref/models/querysets/#select-for-update) and in Pony [here](https://docs.ponyorm.org/transactions.html#pessimistic-locking)

dkopitsa:
Add clause `for_update(nowait: bool = False, skip_locked: bool = False, of=())` for `objects` and `select` query. 

Similar to django's [select_for_update](https://docs.djangoproject.com/en/5.0/ref/models/querysets/#select-for-update)
Closes #932 

Example usage:

Get one row matching the status, locks it for update, and skips any already locked rows. 
```python
async with DB.transaction():
    await Outbox.objects().where(Outbox.status == 'failed').limit(1).for_update(skip_locked=True).first()
```

Ensure that the selected object is locked and cannot be modified by other transactions
```python
async with DB.transaction():
     await org = Organization.objects().where(Organization.code == code).for_update()
     # do some work with object ...
     await org.save()
```


tests/table/test_select.py
```diff
@@ -1028,6 +1028,33 @@ def test_select_raw(self):
             response, [{"name": "Pythonistas", "popularity_log": 3.0}]
         )
 
+    def test_for_update(self):
```
sinisaos, line 1031: @dkopitsa I think you should add something like this to skip the Sqlite test
```python
@pytest.mark.skipif(
    is_running_sqlite(),
    reason="SQLite doesn't support SELECT .. FOR UPDATE.",
)
def test_for_update(self):
```

sinisaos:
@dkopitsa Can you please run locally .`/scripts/format.sh` and then `./scripts/lint.sh` to fix `linter/format` issue and then make another commit. After that everything should be ok because tests are passing.

dkopitsa:
> @dkopitsa Can you please run locally .`/scripts/format.sh` and then `./scripts/lint.sh` to fix `linter/format` issue and then make another commit. After that everything should be ok because tests are passing.

Done. Sorry about the linters; I wanted to fix the tests quickly and used GitHub's web editor.

sinisaos:
> Done. Sorry about the linters; I wanted to fix the tests quickly and used GitHub's web editor.

No worries. Thanks for your PR. Looks good to me, but you'll have to wait for @dantownsend proper review.

piccolo/query/mixins.py
```diff
@@ -784,3 +784,51 @@ def on_conflict(
...
+class ForUpdate:
+    __slots__ = ("nowait", "skip_locked", "of")
+
+    nowait: bool
+    skip_locked: bool
+    of: tuple[Table]
+
+    def __post_init__(self):
+        if not isinstance(self.nowait, bool):
+            raise TypeError("nowait must be an integer")
```
deliro, line 799: typo: bool

piccolo/query/mixins.py
```diff
@@ -784,3 +784,51 @@ def on_conflict(
...
+        if not isinstance(self.skip_locked, bool):
+            raise TypeError("skip_locked must be an integer")
```
deliro, line 801: typo: bool

piccolo/query/mixins.py
```diff
@@ -784,3 +784,51 @@ def on_conflict(
...
+        if not isinstance(self.of, tuple) or not all(
+            hasattr(x, "_meta") for x in self.of
+        ):
+            raise TypeError("of must be an tuple of Table")
```
deliro, line 805: typo: a tuple

piccolo/query/mixins.py
```diff
@@ -784,3 +784,51 @@ def on_conflict(
...
+        if self.nowait and self.skip_locked:
+            raise TypeError(
+                "The nowait option cannot be used with skip_locked"
+            )
+
+    @property
+    def querystring(self) -> QueryString:
+        sql = " FOR UPDATE"
+        if self.of:
+            tables = ", ".join(x._meta.tablename for x in self.of)
```
deliro, line 815: shouldn't the args be escaped?

dkopitsa: All other code assumes that `table._meta.tablename` is safe. An extra sanity check can be done in `piccolo.table.Table.__init_subclass__`, but it is not part of this PR

deliro:
Any updates here?

dkopitsa:
Hi @dantownsend , could you please review the PR? If there's anything that needs further work or adjustments, let me know, and I'll make the necessary changes

docs/src/piccolo/query_clauses/for_update.rst
```diff
@@ -0,0 +1,62 @@
...
+Ignore locked rows
+
+.. code-block:: python
+
+    await Band.select(Band.name == 'Pythonistas').for_update(skip_locked=True)
+
+
+
+of
+~~~~~~~
```
dantownsend, line 54: Can we add a blank line under the headings, and make the underlines match the width of the title? It's a minor thing that RST linters complain about.

dantownsend (commented):
Hi - sorry for the delay with this. I think the PR was opened while I was busy adding MFA to Piccolo Admin.

It looks really good, but my main concern is whether we should refactor it slightly, so instead of the method being `for_update`, instead it should be something like this:

```python
await Band.select().locking('for update')
```

Or even:

```python
await Band.select().for('update')
```

Just because there are multiple strengths of locks:

https://www.postgresql.org/docs/current/sql-select.html#SQL-FOR-UPDATE-SHARE

We could support them all in one method, rather than having to add more methods in the future for each one.

I quite like the idea of calling the method `for` but it clashes with the Python `for` keyword, which can sometimes cause weird bugs. So `lock_for` might be better.

What do you think?

dkopitsa:
I fixed the formatting in test_select.py, please run workflow again.

dantownsend:
@dkopitsa This looks great - thanks a lot!

I'll test it out locally.

dantownsend:
@dkopitsa I've merged this in now, via another branch.

I just changed the method name from `lock_for` to `lock_rows`. I realised when looking at the docs that this looked a bit off when no lock strength was specified:

```python
await Band.select().lock_for()
```

So now it's this instead:

```python
await Band.select().lock_rows()
```

Thanks a lot for all the effort you put into this - it's a very nice feature to have 👍

PR #1067, opened 2024-08-16
granian support in asgi templates

sinisaos:
FastAPI updated the Starlette version in the latest release, and I added support for Granian in asgi templates.
Related to [this Piccolo Admin issue](https://github.com/piccolo-orm/piccolo_admin/issues/399).


dantownsend:
@sinisaos Looks good, thanks!

Issue #1073, opened 2024-09-13
increment number count using object style


trondhindenes:
It would be good if it was possible to do something like:
```
band = await Band.objects().get(Band.name == 'Pythonistas')
assert band.count == 1
band.count.increment(1)
band = await band.update()
assert band.count == 2
```

that is, using "database-level" number increments together woth "object-style" queries. I can't find any sign of this in the documentation, so I guess only dict-style queries support "server-side" increments atm?

dantownsend:
If there's a number in the database which you want to increment, you can do it with update queries.

```python
# If we have this table
class Concert(Table):
    tickets_available = Integer()
```

We can change the value like this:

```python
# If we have this table
await Concert.update({
    Concert.tickets_available: Concert.tickets_available + 1
}).where(Concert.id == some_id)
```

If you have an object, you can do this:

```python
concert = await Concert.objects().where(Concert.id == some_id).first()
concert.tickets_available += 1
await concert.save()
```

The problem with the approach above is if there are multiple saves at the same time - the `tickets_available` might be incorrect.

So you're suggesting something like this:

```python
concert = await Concert.objects().where(Concert.id == some_id).first()
await concert.increment(Concert.tickets_available)

# Where increment effectively runs this query under the hood:
await Concert.update({
    Concert.tickets_available: Concert.tickets_available + 1
}).where(
    Concert.id == concert.id
).returning(
    Concert.tickets_available
)
```

Yeah, I can definitely see the use case.

trondhindenes:
yup, the idea was to have a "visibly separate" syntax for letting the database calculate the increment as apposed to do it in code, which as you mention might produce unexpected results

dantownsend:
I was thinking about it, and the best solution I could come up with is having an `update_self` method, which just updates the object it's called on. It's more flexible than having dedicated `increment` / `decrement` methods.

trondhindenes:
that would work!

dantownsend:
This is done now - will be in the next release.

trondhindenes:
good! Thanks for the quick feedback as always!

PR #1081, opened 2024-09-19
1073 Add `update_self` method

dantownsend:
Resolves https://github.com/piccolo-orm/piccolo/issues/1073

Remaining tasks:

- [x] Add `update_self` method
- [x] Add tests
- [x] Add docs

### DIFF
--- a/piccolo/query/methods/objects.py
+++ b/piccolo/query/methods/objects.py
@@ -27,6 +27,7 @@ from piccolo.utils.sync import run_sync
 
 if t.TYPE_CHECKING:  # pragma: no cover
     from piccolo.columns import Column
+    from piccolo.table import Table
 
 
 ###############################################################################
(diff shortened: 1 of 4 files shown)


piccolo/query/methods/objects.py
```diff
@@ -173,6 +174,93 @@ def run_sync(self, *args, **kwargs) -> TableInstance:
...
+            tickets_available = Integer()
+
+    And we fetch an object::
+
+        >>> concert = await Concert.objects().get(name="Amazing concert")
+
+    We could use the typical syntax for updating the object::
+
+        >>> concert.tickets_available += -1
+        >>> await concert.save()
```
dantownsend, line 195: TODO - update this to use `Band.popularity` as the example column, as it already exists in the playground.

PR #1089, opened 2024-09-30
1090 fix(m2m): Return empty list if there are no m2m entries

nVitius:
I ran into an issue where running `getm2m` without any existing m2m entries will throw an error that I believe is unintended.
image

### DIFF
--- a/piccolo/columns/m2m.py
+++ b/piccolo/columns/m2m.py
@@ -386,7 +386,7 @@ class M2MGetRelated:
 
         results = await secondary_table.objects().where(
             secondary_table._meta.primary_key.is_in(ids)
-        )
+        ) if len(ids) > 0 else []
 
         return results
 
--- a/piccolo/columns/m2m.py
+++ b/piccolo/columns/m2m.py
@@ -384,9 +384,13 @@ class M2MGetRelated:
             .output(as_list=True)
         )
 
-        results = await secondary_table.objects().where(
-            secondary_table._meta.primary_key.is_in(ids)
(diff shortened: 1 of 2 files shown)


nVitius:
Seems unintuitive to throw an error instead of returning an empty list...
Also, the error isn't very obvious, as it's coming from the `is_in` call within the `get_m2m` method.

dantownsend:
I'm just looking at this now.

dantownsend:
Yeah, I think this is a bug.

@sinisaos You make a good point - if someone calls `is_in` directly with no values, then I think raising an error is OK. But for this particular query, the user isn't calling `is_in` directly, so we should handle this edge case and return an empty list.

dantownsend:
Resolves https://github.com/piccolo-orm/piccolo/issues/1090

sinisaos:
> Seems unintuitive to throw an error instead of returning an empty list... Also, the error isn't very obvious, as it's coming from the `is_in` call within the `get_m2m` method.
> 
> To clarify, the error is indicating that `is_in` didn't receive any id's to check. Not that there are no entries.

@nVitius Sorry for my mistake. Thanks for the clarification.

dantownsend:
@sinisaos No need to apologise

dantownsend:
OK, I'm going to throw up a quick test, then will get this merged in. Thanks both 👍

dantownsend:
@nVitius Thanks a lot for the fix - it will be in the next release 👍

dantownsend:
The fix is in `piccolo==1.19.1`.

Issue #1094, opened 2024-10-04
Add Python 3.13 support


dantownsend:
Python 3.13 will be out very soon, which also means Python 3.8 will be end of life.

We should update our CI, so we test Python 3.13 and not 3.8.

We should also update the `classifiers` in `setup.py`.

Now we no longer support Python 3.8 we can change the type annotations from `t.List[str]` to `list[str]` etc. This will be a massive change, so best done in another PR.

PR #1104, opened 2024-10-18
1094 Add Python 3.13 support

dantownsend:
Resolves https://github.com/piccolo-orm/piccolo/issues/1094


dantownsend:
It looks like asyncpg doesn't support Python 3.13 yet.

It looks like it will soon though:

https://github.com/MagicStack/asyncpg/issues/1181

For now, this PR is blocked.

dantownsend:
The only remaining issue is BlackSheep doesn't have Python 3.13 wheels at the moment:

https://github.com/Neoteroi/BlackSheep/issues/515

We can run the integration tests in Python 3.12 for now, and I'll create a follow up issue.

Issue #1108, opened 2024-10-18
Be able to pass an engine into `TableStorage`


dantownsend:
Related to https://github.com/piccolo-orm/piccolo_admin/issues/129

PR #875, opened 2023-07-29
Add possibility to pass engine on table reflection

ArtsiomAntropau:

PR #1109, opened 2024-10-18
1108 Be able to pass an engine into `TableStorage`

dantownsend:
Resolves https://github.com/piccolo-orm/piccolo/issues/1108

### DIFF
--- a/piccolo/table_reflection.py
+++ b/piccolo/table_reflection.py
@@ -8,6 +8,8 @@ import typing as t
 from dataclasses import dataclass
 
 from piccolo.apps.schema.commands.generate import get_output_schema
+from piccolo.engine import engine_finder
+from piccolo.engine.base import Engine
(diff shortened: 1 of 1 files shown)


sinisaos:
@dantownsend When you merge this PR, can you please close [this old PR](https://github.com/piccolo-orm/piccolo/pull/875) that basically did the same thing. Thanks.

dantownsend:
@sinisaos Ah, sorry - I completely forgot about that one.

They are slightly different in that this one passes engine into `__init__`, and the other one passes it into `reflect`. Which do you prefer?

sinisaos:
@dantownsend I think both versions are fine, but I prefer your version, because I think it's more common to pass config arguments in constructor, than in methods.

dantownsend:
@sinisaos Thanks - lets go with this for now then.

PR #1111, opened 2024-10-18
1107 Fix `prefetch` when a `ForeignKey` has `db_column_name` set

dantownsend:
Resolves https://github.com/piccolo-orm/piccolo/issues/1107

### DIFF
--- a/piccolo/table.py
+++ b/piccolo/table.py
@@ -157,7 +157,11 @@ class TableMeta:
         """
         components = name.split(".")
         column_name = components[0]
-        column = [i for i in self.columns if i._meta.name == column_name]
+        column = [
+            i
+            for i in self.columns
+            if column_name in (i._meta.name, i._meta.db_column_name)
+        ]
         if len(column) != 1:
             raise ValueError(f"No matching column found with name == {name}")
         column_object = column[0]


VladislavYar:
@dantownsend Hi. Will it ever reach release? Without this fix, it's hard to switch from Django ORM.

dantownsend:
> @dantownsend Just one question. Shouldn't this be much simpler without the `match_db_column_name` function parameter and the extra table to test? Something like this:
> 
> ```python
> # your modified code without match_db_column_name param
> column_object = next(
>    (
>        i
>        for i in self.columns
>        if (i._meta.name == column_name)
>        or (i._meta.db_column_name == column_name)
>    ),
>    None,
> )
> if column_object is None:
>    raise ValueError(f"No matching column found with name == {name}")
> ```
> 
> and test like this:
> 
> ```python
> # in tests/columns/test_db_column_name.py
> def test_prefetch(self):
>     """
>     Make sure that foreign keys with a ``db_column_name`` work
>     with prefetch objects.
> 
>     https://github.com/piccolo-orm/piccolo/issues/1107
> 
>     """
>     manager = Manager.objects().create(name="Guido").run_sync()
>     self.insert_band(manager=manager)
> 
>     band = (
>         Band.objects(Band.manager)
>         .where(Band.manager == manager.id)
>         .first()
>         .run_sync()
>     )
> 
>     self.assertEqual(
>         band.to_dict(),
>         {
>             "id": band.id,
>             "name": "Pythonistas",
>             "popularity": 1000,
>             "manager": {"id": manager.id, "name": "Guido"},
>         },
>     )
> ```
> 
> Sorry if I'm wrong and don't understand this correctly but this is just suggestion.

Yeah, I need a table with a FK with `db_column_name` defined. I added a table like that to the playground too, as it's good for manual testing. My main laptop is being repaired, so setting up a new local dev environment.

Issue #1103, opened 2024-10-18
Bug: Where clause with Postgres JSONB nested field


sayanarijit:
Using nested jsonb field in `where` clause is producing unexpected result.

## How to reproduce:

```python
from piccolo.table import Table, create_db_tables_sync
from piccolo import columns as col

class Foo(Table):
    meta = col.JSONB()

def main():
    # Create the table
    create_db_tables_sync(Foo, if_not_exists=True)

    # Create a new objects
    Foo(meta={"a": True, "b": True}).save().run_sync()
    Foo(meta={"a": True, "b": False}).save().run_sync()
    Foo(meta={"a": False, "b": True}).save().run_sync()
    Foo(meta={"a": False, "b": False}).save().run_sync()

    # Query the table
    faulty_results = Foo.select().where(Foo.meta.arrow("a") != Foo.meta.arrow("b")).run_sync()
    for row in faulty_results:
        print(row)

if __name__ == "__main__":
    main()
```

## Expected behavior

```python
{'id': 2, 'meta': '{"a": true, "b": false}'}
{'id': 3, 'meta': '{"a": false, "b": true}'}
```

## Actual behavior

```python
{'id': 1, 'meta': '{"a": true, "b": true}'}
{'id': 2, 'meta': '{"a": true, "b": false}'}
{'id': 3, 'meta': '{"a": false, "b": true}'}
{'id': 4, 'meta': '{"a": false, "b": false}'}
```

Repository for convenience: https://github.com/sayanarijit/piccolobug

sinisaos:
@sayanarijit I don't know if this is a bug or not, but we can achieve the expected behavior in three ways.
1. using complex checking for differences between `meta.a` and `meta.b`
```python
faulty_results = (
      Foo.select()
      .where(
          (Foo.meta.arrow("a") == True) & (Foo.meta.arrow("b") == False)
          | (Foo.meta.arrow("a") == False) & (Foo.meta.arrow("b") == True)
      )
      .run_sync()
  )
```
2. using raw sql
```python
faulty_results = Foo.raw(
    "SELECT * FROM foo WHERE meta->'a' != meta->'b'"
).run_sync()
```
3. using [WhereRaw](https://piccolo-orm.readthedocs.io/en/latest/piccolo/query_clauses/where.html#whereraw) to combine Piccolo query with raw sql
```python
faulty_results = Foo.select().where(WhereRaw("meta->'a' != meta->'b'")).run_sync()
```

[...]

PR #1112, opened 2024-10-18
1103 Improve arrow function

dantownsend:
Resolves https://github.com/piccolo-orm/piccolo/issues/1103

### DIFF
--- a/docs/src/piccolo/schema/column_types.rst
+++ b/docs/src/piccolo/schema/column_types.rst
@@ -201,6 +201,7 @@ JSONB
 
 .. autoclass:: JSONB
 
+===========
 Serialising
 ===========
 
@@ -224,6 +225,7 @@ You can also pass in a JSON string if you prefer:
     )
     await studio.save()
 
+=============
 Deserialising
 =============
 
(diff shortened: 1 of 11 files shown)


sinisaos:
@dantownsend It seems that when we use the arrow function we need `strings` for queries (not `booleans` or other types). I think that makes sense because querystring uses strings. If we change the data type to string in these tests, everything passes. 
```python
def test_arrow_where(self):
    """
    Make sure the arrow function can be used within a WHERE clause.
    """
    RecordingStudio(
        name="Abbey Road", facilities='{"mixing_desk": true}'
    ).save().run_sync()

    self.assertEqual(
        RecordingStudio.count()
        .where(RecordingStudio.facilities.arrow("mixing_desk").eq("true")) # <-- here
        .run_sync(),
        1,
    )

    self.assertEqual(
        RecordingStudio.count()
        .where(RecordingStudio.facilities.arrow("mixing_desk").eq("false")) # <-- here
        .run_sync(),
        0,
    )
```
I also removed the [load json test](https://github.com/piccolo-orm/piccolo/blob/e2fa8f89f94bc995722ca42db856d9aad6471a46/tests/columns/test_jsonb.py#L158C9-L167C50) because the `output(load_json=True)` doesn't return a boolean after new changes.

I hope that makes sense.

dantownsend:
@sinisaos Thanks - you're right, the JSON values have to be strings.

sinisaos (approved):
@dantownsend I think this is a great enhancement :+1:

dantownsend:
@sinisaos Thanks. I went a bit crazy on this in the end.

You can now go multiple levels deep, and traverse array elements within the JSON.

```python
await RecordingStudio.select(
    RecordingStudio.facilities["a"]["b"][0]["c"]
)
```

sinisaos:
@dantownsend Wow! It's really amazing.

dantownsend:
@sinisaos Thanks. It turned into a bigger thing than I expected. But with this, we can potentially do better JSON filtering in Piccolo Admin.

PR #1125, opened 2024-11-07
Update insert.rst

daveflr:
A missing comma was added

sorry this is such an insignificant change, but my ADHD wouldn't let me ignore it.


dantownsend:
@daveflr Great catch, thanks!

PR #1138, opened 2025-01-10
add Quart asgi template

sinisaos:
Adding Quart asgi template. I was able to add a Piccolo admin using [Hypercorn DispatcherMiddleware](https://hypercorn.readthedocs.io/en/latest/how_to_guides/dispatch_apps.html)

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,7 +94,7 @@ Let Piccolo scaffold you an ASGI web app, using Piccolo as the ORM:
 piccolo asgi new
 ```
 
-[Starlette](https://www.starlette.io/), [FastAPI](https://fastapi.tiangolo.com/), [BlackSheep](https://www.neoteroi.dev/blacksheep/), [Litestar](https://litestar.dev/), [Esmerald](https://esmerald.dev/) and [Lilya](https://lilya.dev) are currently supported.
+[Starlette](https://www.starlette.io/), [FastAPI](https://fastapi.tiangolo.com/), [BlackSheep](https://www.neoteroi.dev/blacksheep/), [Litestar](https://litestar.dev/), [Esmerald](https://esmerald.dev/), [Lilya](https://lilya.dev) and [Quart](https://quart.palletsprojects.com/en/latest/) are currently supported.
 
 ## Are you a Django user?
 
--- a/piccolo/apps/asgi/commands/new.py
+++ b/piccolo/apps/asgi/commands/new.py
@@ -17,6 +17,7 @@ ROUTER_DEPENDENCIES = {
     "litestar": ["litestar"],
     "esmerald": ["esmerald"],
     "lilya": ["lilya"],
+    "quart": ["quart", "quart_schema"],
 }
 ROUTERS = list(ROUTER_DEPENDENCIES.keys())
 
--- a/piccolo/apps/asgi/commands/templates/app/_quart_app.py.jinja
+++ b/piccolo/apps/asgi/commands/templates/app/_quart_app.py.jinja
@@ -1,0 +1,119 @@
+import typing as t
+from http import HTTPStatus
+
+from hypercorn.middleware import DispatcherMiddleware
+from piccolo.engine import engine_finder
+from piccolo_admin.endpoints import create_admin
+from piccolo_api.crud.serializers import create_pydantic_model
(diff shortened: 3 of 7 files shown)


dantownsend:
@sinisaos Interesting - I've been aware of Quart for a long time, but wasn't sure how to mount a third party ASGI app. Congratulations on finding a solution 👏

sinisaos:
Same thing also works with Sanic. I will create an asgi template for Sanic, but first I need to understand how Pydantic and OpenAPI docs work in Sanic. They say it's automatic, but that's not exactly true because we need to make a lot of decorators for the configuration. I will do my best.

dantownsend:
Using the Hypercorn middleware is a nice idea, I wasn't aware it existed.

sinisaos:
@dantownsend Small offtopic. If you have some time you can check this [repo](https://github.com/sinisaos/async-orm-benchmarks). I made some benchmarks for asgi drivers and ORM's. I'm not a benchmark expert, but benchmarks show that Piccolo has the best performance (by some margin) for reading data than other ORMs. I didn't consider ORM's like Ormar or SQLModel (and many others) because those ORM's are just wrappers around SQLALchemy. Other than that, Piccolo is the easiest to use, easy to configure and set up, and has many great features (still missing `unique_constaints`, `reverse_lookup`, but there are PRs for that). I think that is great.

sinisaos:
> Using the Hypercorn middleware is a nice idea, I wasn't aware it existed.

@dantownsend Yes. That feature is really useful and allows us to enable Piccolo Admin for frameworks that don't have asgi app mounting options.

dantownsend:
> still missing `unique_constaints`, `reverse_lookup`, but there are PRs for that

Yes, I do need to get around to testing those. The big PRs are challenging because I need to find a big block of time to properly review and test them. It's also challenging balancing the features which the community needs vs what I need immediately for work. Sometimes it might look like I prioritise random stuff, but it's usually something I need for a project I'm working on.

I think the composite unique constraints one is probably highest priority. I have never actually needed that feature on a project, but I know a lot of people would expect it to exist.

sinisaos:
No worries. Piccolo already has a lot of cool features, but I see a lot of interest in those two features. A reverse lookup (in some form) would also be valuable in a situation [like this](https://github.com/sinisaos/async-orm-benchmarks/blob/main/benchmarks/piccolo_orm/main.py#L64-L86) where we need one less db query and it affects performance (results have better reqs/s for that endpoint with a reverse lookup).

sinisaos:
@dantownsend I've created a [Sanic asgi template](https://github.com/sinisaos/piccolo/tree/sanic_asgi_template), but I'm having problems with the integration test because our dummy server doesn't start the Sanic app (which throws a Sanic exception `Loop can only be retrieved after the app has started running. Not supported with create_server function`). I made changes to use [Uvicorn as the test server](https://github.com/sinisaos/piccolo/blob/sanic_asgi_template/tests/apps/asgi/commands/files/test_server.py) (based on the code found in the Uvicorn tests). Is that okay with you or?

dantownsend:
I'm not sure, I haven't experimented with Sanic recently. So you can't just run it with `AsyncClient(transport=ASGITransport(app=app))`? Does it have its own server or something?

sinisaos:
Sanic has its own server, but it is also ASGI-compliant and we can use `uvicorn` and everything works with `uvicorn`. Only the integration tests don't pass, so I just changed the `dummy_server` to use `uvicorn`. I want to know if that's okay with you or if you have a different idea.

dantownsend:
This looks good to me, thanks!

dantownsend:
@sinisaos One potential issue with this, which I only just though of, is the user has to pick Hypercorn as the server. Ideally we would not give them the option to pick a router, and just Hypercorn automatically.

sinisaos:
> One potential issue with this, which I only just though of, is the user has to pick Hypercorn as the server

Sorry for the slow reply, I wasn't home for the weekend. I don't know if I understood correctly (the user does not have to use Hypercorn exclusively). In my case all 3 servers are working fine.

dantownsend:
Ah, ok - I thought it just worked with Hypercorn. I'll give it a go, thanks!

Issue #1132, opened 2024-12-13
`create_pydantic_model` returns models with wrong fields marked as `null`


HakierGrzonzo:
When creating pydantic models with `create_pydantic_model` function, columns marked as `null` are not marked as `null` in the data. Instead the function uses the `required` property.

## How to recreate

Consider the following application:

```python
from piccolo.table import Table
from piccolo.utils.pydantic import create_pydantic_model
from piccolo import columns

# We need something that consumes the Pydantic models
from fastapi import FastAPI

import json

class SomeThing(Table):
    name = columns.Text(null=False)
    pet_name = columns.Text(null=True)
    age = columns.Integer(required=True, null=True)

some_model = create_pydantic_model(SomeThing)

app = FastAPI()

@app.get('/')
def foo() -> some_model:
    pass

# We print the OpenAPI schema generated for table `SomeThing`
print(json.dumps(app.openapi()['components']['schemas']['SomeThing']))
```

The output returned is:
```json
{
  "properties": {
    "name": {
      "anyOf": [
        {
          "type": "string"
        },
        {
          "type": "null"
        }
      ],
      "title": "Name",
      "extra": {
        "nullable": false,
        "secret": false,
        "unique": false,
        "widget": "text-area"
      }
    },
    "pet_name": {
      "anyOf": [
        {
          "type": "string"
        },
        {
          "type": "null"
        }
      ],
      "title": "Pet Name",
      "extra": {
        "nullable": true,
        "secret": false,
        "unique": false,
        "widget": "text-area"
      }
    },
    "age": {
      "type": "integer",
      "title": "Age",
      "extra": {
        "nullable": true,
        "secret": false,
        "unique": false
      }
    }
  },
  "type": "object",
  "required": [
    "age"
  ],
  "title": "SomeThing",
  "extra": {}
}
```

As we can see, the fields `name` and `pet_name` are marked as `str | None`, despite having different `null` options set.

In contrast, the `age` field is marked as an `int`, which may never be `None`, despite it being possibly `null` in the database.

## What causes this behavior:

https://github.com/piccolo-orm/piccolo/blob/7373a8490c84e3c1577d987a34dbc376474c2c30/piccolo/utils/pydantic.py#L246

In the above line, the type for the field is set as `T | None` based on the `required` property, instead of the `null` property.

This generates plausible models for writing to the database, but not for reading data from the database (we get rogue nullable properties).

## Why does this behavior need to be changed:

[...]

PR #1141, opened 2025-01-14
1132 If a column is nullable, make it optional by default in `create_pydantic_model`

dantownsend:
Resolves https://github.com/piccolo-orm/piccolo/issues/1132

### DIFF
--- a/piccolo/columns/base.py
+++ b/piccolo/columns/base.py
@@ -153,7 +153,7 @@ class ColumnMeta:
     unique: bool = False
     index: bool = False
     index_method: IndexMethod = IndexMethod.btree
-    required: bool = ...
+    required: t.Optional[bool] = None
     help_text: t.Optional[str] = None
     choices: t.Optional[t.Type[Enum]] = None
     secret: bool = False
@@ -459,7 +459,7 @@ class Column(Selectable):
         unique: bool = False,
         index: bool = False,
         index_method: IndexMethod = IndexMethod.btree,
-        required: bool = ...,
+        required: t.Optional[bool] = None,
         help_text: t.Optional[str] = None,
         choices: t.Optional[t.Type[Enum]] = None,
         db_column_name: t.Optional[str] = None,
--- a/piccolo/utils/pydantic.py
+++ b/piccolo/utils/pydantic.py
@@ -248,7 +248,7 @@ def create_pydantic_model(
 
         if all_optional:
             is_optional = True
-        elif column._meta.required is not ...:
+        elif column._meta.required is not None:
             # The user can force the field to be optional or not, irrespective
             # of whether it's nullable in the database.
             is_optional = not column._meta.required


Skelmis:
Assuming `required` becoming `None` doesn't break the admin panels behavior I like the look of this

sinisaos:
I agree with these changes and everything works, but we have one case where someone makes a mistake and marks a null field as optional. We have four options. 
```python
class Band(Table):
    name = Varchar() # null=False by default

>>> create_pydantic_model(Band).model_fields['name'].annotation
str # correct

class Band(Table):
    name = Varchar(null=True)

>>> create_pydantic_model(Band).model_fields['name'].annotation
Optional[str] # correct

class Band(Table):
    name = Varchar(required=True, null=True)

>>> create_pydantic_model(Band).model_fields['name'].annotation
str # correct

class Band(Table):
    name = Varchar(required=False) # null=False by default

>>> create_pydantic_model(Band).model_fields['name'].annotation
Optional[str] # not correct, should be str (required because is not null constraint in db)
```

Should we prevent these possible user errors or not? If we change [this line of code](https://github.com/piccolo-orm/piccolo/issues/1132#issuecomment-2543933550), we can force `required` to be `True` if `null` is `False`, to prevent user errors. After that change result is 
```python
class Band(Table):
    name = Varchar(required=False) # null=False by default

>>> create_pydantic_model(Band).model_fields['name'].annotation
str # correct

```
which is correct.

dantownsend:
@Skelmis and @sinisaos - thanks for your feedback on this.

I've been thinking about it today, and I'm a bit undecided still.

## Current behaviour

The current behaviour in Piccolo is `required=False` by default, which means that if we have this:

```python
class Band(Table):
    name = Varchar()
```

The it generates the following Pydantic model:

```python
class BandModel(BaseModel):
    name: Optional[str]
```

In terms of generating the Pydantic model, `null` is completely ignored, it's completely driven by `required`.

This is similar to Django where they have `nullable` and `blank` arguments. Only `blank` determines if the field is required in Django Admin.

Maybe we should have made `required=True` the default - it's hard to say.

## Multiple Pydantic models

The downside with putting `required` on the Piccolo `Table` is that you will often have multiple Pydantic models per table. For example:

1. A model for filtering or `PATCH` (where every field is optional)
1. A model for returning data in a `GET` request, or receiving a `POST` (where some fields will be required)

We get around 1 by having the `all_optional` argument in `create_pydantic_model`.

```python
create_pydantic_model(
    Band,
    all_optional=True
)
```

If someone required even more Pydantic models from the same Piccolo table it gets tricky. Perhaps there could be a `required_columns` arg in `create_pydantic_model`:

```python
create_pydantic_model(
    Band,
    required_columns=Band.all_columns(exclude=[Band.some_column])
)
```

That gives the ultimate flexibility.

## Conclusions (?)

I'm still a bit on the fence ... I'm interested to hear your thoughts.

I think I'll refactor https://github.com/piccolo-orm/piccolo_admin/pull/432 slightly for now, so it's decoupled from this PR, as I don't want to rush this in and make a mistake.

dantownsend:
> ```python
> >>> create_pydantic_model(Band).model_fields['name'].annotation
> Optional[str] # not correct, should be str (required because is not null constraint in db)
> ```

@sinisaos I think for that situation we just assume that the user will add the missing value before saving it to the database.

PR #1149, opened 2025-02-11
feat: add Callable typing to Timestampz default types

Skelmis:


dantownsend:
Thanks - I think this makes sense. We had `Callable` as an option for `Timestamp`, but not `Timestamptz`, strange!

https://github.com/piccolo-orm/piccolo/blob/55c2dbcf4b7aaceaf818bc566d7feea7862187c5/piccolo/columns/defaults/timestamp.py#L141

PR #1165, opened 2025-03-14
Fixes #1164 - Add support for None typehint in Varchar length attribute

### LINKED ISSUE #1164: None typehint support for Varchar
https://github.com/piccolo-orm/piccolo/blob/2a637b870bf14857d1579dcd87e8a4bf31aa2b61/piccolo/columns/column_types.py#L318

Compro-Prasad:
Resolves https://github.com/piccolo-orm/piccolo/issues/1164

### DIFF
--- a/piccolo/columns/column_types.py
+++ b/piccolo/columns/column_types.py
@@ -315,7 +315,7 @@ class Varchar(Column):
 
     def __init__(
         self,
-        length: int = 255,
+        length: t.Optional[int] = 255,
         default: t.Union[str, Enum, t.Callable[[], str], None] = "",
         **kwargs,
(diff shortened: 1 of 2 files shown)


piccolo/columns/column_types.py
```diff
@@ -315,7 +315,7 @@ class Band(Table):
 
     def __init__(
         self,
-        length: int = 255,
+        length: int | None = 255,
```
dantownsend, line 318: Can we make this `t.Optional[int]` instead? Just because we still support Python 3.9, and the `|` operator isn't available in that version.

dantownsend:
Thanks for this 👍

PR #1168, opened 2025-03-21
Fix `day` and `minute` in `TimestamptzCustom`'s `from_datetime`

splch:
Resolves https://github.com/piccolo-orm/piccolo/issues/1169

When setting a field like 

```python
class ABC(Table, db=DB):
    date = Timestamptz(
        null=False,
        default=datetime.datetime(
            9999, 12, 31, 23, 59, 59, tzinfo=datetime.timezone.utc
        ),
    )
```

the field in the database is set as `9999-12-12 23:00:59+00` instead of `9999-12-31 23:59:59+00`. This PR fixes the `TimestamptzCustom`'s `from_datetime` method.

### DIFF
--- a/piccolo/columns/defaults/timestamp.py
+++ b/piccolo/columns/defaults/timestamp.py
@@ -116,7 +116,7 @@ class TimestampCustom(Default):
         return cls(
             year=instance.year,
             month=instance.month,
-            day=instance.month,
+            day=instance.day,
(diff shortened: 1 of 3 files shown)


dantownsend:
@splch Great catch! Thanks for the PR.

splch:
no problem happy to find it :)

Issue #1171, opened 2025-03-27
Delete operations do not include `JOIN` clause in the query


HakierGrzonzo:
# Issue:

When using the `.delete()` query type
And when the condition in the `.where()` query clause includes a join
Then the generated SQL query does not include the `LEFT JOIN` clause (or a subquery, since we can't put a `JOIN` on a `DELETE` query)
And the database returns a *missing FROM-clause entry for table* error.

# How to reproduce:

```python
from piccolo.columns import UUID, ForeignKey
from piccolo.table import Table


class Foo(Table):
    id = UUID(primary_key=True)

class Bar(Table):
    id = UUID(primary_key=True)
    foo = ForeignKey(Foo)


print(Bar.delete().where(Bar.foo._.id == "foo"))
```

Output:

```sql
DELETE FROM "bar" WHERE "bar$foo"."id" = 'foo'
```

Which results in errors like these in postgresql (different table names and query):
```
ERROR:  missing FROM-clause entry for table "chat_message$chat" at character 35
STATEMENT:  DELETE FROM "chat_message" WHERE ("chat_message$chat"."owner" = $1 AND "chat_message"."id" =$2) RETURNING "id"
```

## Notes:

When using diffrent query types on the above example:
```python
print(Bar.update().where(Bar.foo._.id == "foo"))
print(Bar.select().where(Bar.foo._.id == "foo"))
```

We get the correct output:
```sql
UPDATE "bar" SET  WHERE foo IN (SELECT ALL "bar"."foo" AS "foo" FROM "bar" LEFT JOIN "foo" "bar$foo" ON ("bar"."foo" = "bar$foo"."id") WHERE "bar$foo"."id" = 'foo')
SELECT ALL "bar"."id" AS "id", "bar"."foo" AS "foo" FROM "bar" LEFT JOIN "foo" "bar$foo" ON ("bar"."foo" = "bar$foo"."id") WHERE "bar$foo"."id" = 'foo'
```

Additionaly I was looking at https://github.com/piccolo-orm/piccolo/blob/9424f97c0460f1928f491b6a3d8151a04aa5d498/piccolo/query/methods/delete.py#L19-L74 and found that the select query has some join logic https://github.com/piccolo-orm/piccolo/blob/9424f97c0460f1928f491b6a3d8151a04aa5d498/piccolo/query/methods/select.py#L487-L540 which is missing in the delete query type

sinisaos:
[...]

PR #1173, opened 2025-04-02
1171 Delete subquery

sinisaos:
Related to #1171. I renamed `querystring_for_update` to `querystring_for_joins`. If that name is not good, feel free to change it.


dantownsend:
@sinisaos Thanks for this.

I was struggling to know what to name the method.

I just went with `querystring_for_update_and_delete` in the end. Kind of boring, but it's obvious what it does.

Issue #1166, opened 2025-03-17
Set `_exists_in_db` to `False` in the `remove` method


dantownsend:
```python
band = await Band.objects().first()

>>> band._exists_in_db
True

await band.remove()

>>> band._exists_in_db
True (this should be False)
```

sinisaos:
@dantownsend If [this fix](https://github.com/sinisaos/piccolo/tree/exists_in_db_bug) is ​​good enough, I can make a PR.

dantownsend:
Resolved by https://github.com/piccolo-orm/piccolo/pull/1175

PR #1175, opened 2025-04-04
Exists in db bug

sinisaos:
Resolves to #1166


piccolo/table.py
```diff
@@ -573,6 +573,8 @@ def remove(self) -> Delete:
 
         setattr(self, self._meta.primary_key._meta.name, None)
 
+        self._exists_in_db = False
```
dantownsend, line 576: In an ideal world we would have a callback on the delete query, so `_exists_in_db` gets set to `False` after the query has completed. But we have no callbacks on delete queries at the moment, so I think this is the best we can do.

dantownsend (approved):
Looks good - thanks 👍

Issue #1178, opened 2025-04-13
`piccolo create` does not generate `piccolo_app.py` module path correctly under nested directories


Jacky56:
assume you have a project folder such as:
```
├── main.py
├── piccolo_conf.py
└── src/
    ├── posts/
    |   └── piccolo_app.py
    └── users/
```

When I create a new piccolo module such as:
```bash
piccolo create src/bands
```
the generated `src/bands/piccolo_app.py` file will contain:
```py
"""
Import all of the Tables subclasses in your app here, and register them with
the APP_CONFIG.
"""

import os

from piccolo.conf.apps import AppConfig, table_finder


CURRENT_DIRECTORY = os.path.dirname(os.path.abspath(__file__))


APP_CONFIG = AppConfig(
    app_name="src/bands", # not sure if this works correctly
    migrations_folder_path=os.path.join(
        CURRENT_DIRECTORY, "piccolo_migrations"
    ),
    table_classes=table_finder(
        modules=["src/bands.tables"], exclude_imported=True  # should be "src.bands.tables"
    ),
    migration_dependencies=[],
    commands=[],
)
```

sinisaos:
> In that case though, `table_classes` is this:
> 
> table_classes=table_finder(modules=["bands.tables"], exclude_imported=True),
> But it should really be:
> 
> table_classes=table_finder(modules=["src.bands.tables"], exclude_imported=True),

@dantownsend Yes, that's the case. Maybe we should provide a root clause for the template that will set the path correctly. Something like this in `app/new.py`

```python
templates: t.Dict[str, t.Any] = {
    "piccolo_app.py": {
        "app_name": app_name,
        "root": "" if root == "." else f"{root.replace("/", ".")}.", # new line
    },
    "tables.py": {},
}
```
This will produce the correct path in `piccolo_app.py` for `--root=src`
```python
table_classes=table_finder(modules=["src.band.tables"], exclude_imported=True),
```
and for `--root=src/apps` it will be
```python
table_classes=table_finder(modules=["src.apps.band.tables"], exclude_imported=True),
```

dantownsend:
The only problem is Windows, which uses backslashes for paths.

You can do something like:

```python
import pathlib

>>> pathlib.Pathlib(some_path).parts
```

Which will work on Unix or Windows.

An alternative is to use relative imports instead. Here's a PR for that:

https://github.com/piccolo-orm/piccolo/pull/1179

dantownsend:
[...]

PR #1179, opened 2025-04-14
1178 Improve imports when app is created in a custom root

dantownsend:
Related to https://github.com/piccolo-orm/piccolo/issues/1178

Rather than using absolute paths for `table_finder` in `piccolo_app.py`, we can use relative imports.

This is better when someone creates a Piccolo app under a sub directory using the `root` argument.

```
piccolo app new my_app --root=./src
```

### DIFF
--- a/piccolo/apps/app/commands/new.py
+++ b/piccolo/apps/app/commands/new.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import importlib
 import os
+import shutil
 import sys
 import typing as t
 
@@ -44,7 +45,7 @@ def new_app(app_name: str, root: str = "."):
             "Python module. Please choose a different name for your app."
         )
 
-    os.mkdir(app_root)
+    os.makedirs(app_root)
 
     with open(os.path.join(app_root, "__init__.py"), "w"):
         pass
--- a/piccolo/apps/app/commands/new.py
+++ b/piccolo/apps/app/commands/new.py
@@ -2,7 +2,6 @@ from __future__ import annotations
 
 import importlib
 import os
-import shutil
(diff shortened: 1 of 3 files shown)


sinisaos:
@dantownsend You are right, I don't use Windows and I forgot that Windows uses a backslash to separate directories in file paths. I tried your changes but migrations don't work in my case. Sorry maybe I'm not doing something right but if I create app with `piccolo create band --root=src` app is created correctly (right path). After that I add app in `piccolo_conf.py` app registry like this.
```python
APP_REGISTRY = AppRegistry(
  apps=[
      "home.piccolo_app",
      "piccolo_admin.piccolo_app",
      "src.band.piccolo_app", # new app
  ]
)
```
When I try to create a migration with `piccolo migrations new band --auto` the migration has not been created.
If I change the `APP_CONFIG` in the created app to this 
```python
APP_CONFIG = AppConfig(
    app_name="band",
    migrations_folder_path=os.path.join(CURRENT_DIRECTORY, "piccolo_migrations"),
    table_classes=table_finder(
        modules=[f"{get_package(__name__)}.tables"], exclude_imported=True
    ),
    migration_dependencies=[],
    commands=[],
)
```
migrations works in my case (several levels deep). Can you check this or explain to me how I should use the new changes. Thanks

dantownsend:
@sinisaos Ah, yes - you're right. There was a bug. Can you try again?

sinisaos (approved):
Now it works perfectly. A few levels deep path like this `--root=src/apps/src` also works. Great.

dantownsend:
@sinisaos Cool, thanks for testing it!

Issue #1180, opened 2025-04-14
Esmerald integration tests failing


dantownsend:
See here: https://github.com/piccolo-orm/piccolo/actions/runs/14444859902/job/40502955349?pr=1179

The error is:

```python
from esmerald.config import StaticFilesConfig
ModuleNotFoundError: No module named 'esmerald.config'
```

@sinisaos Any ideas?

PR #1181, opened 2025-04-14
fix Esmerald asgi template

sinisaos:
Related to #1180


dantownsend:
@sinisaos Perfect - thanks for this!

Issue #1155, opened 2025-02-17
`manager.alter_column`: changing `on_delete` doesn't seem to work


Jacky56:
Assume you have the scenario of changing the existing on delete cascade behaviour:
```python
class Post(Table):
    multimedia = ForeignKey(Multimedia, null=True, on_delete=OnDelete.set_null)
```
which then perform the auto migration to generates the code below:
```python
async def forwards():
    manager = MigrationManager(
        migration_id=ID, app_name="post", description=DESCRIPTION
    )
    manager.alter_column(
        table_class_name="Post",
        tablename="post",
        column_name="multimedia",
        db_column_name="multimedia",
        params={"on_delete": OnDelete.set_null},
        old_params={"on_delete": OnDelete.cascade},
        column_class=ForeignKey,
        old_column_class=ForeignKey,
        schema=None,
    )
    return manager
```

The migration ran successfully but my FK cascade behaviour did not change

CockroachDB version: Basic, Google Cloud v24.3.4
Piccolo version: piccolo==1.21.0

dantownsend:
Yes, you're right - I've noticed this too. The migrations don't detect this change at the moment, I need to fix it.

sinisaos:
@Jacky56 Can you try a potential fix from [this branch](https://github.com/piccolo-orm/piccolo/compare/master...sinisaos:piccolo:on_delete_migrations). In my case this works.

**CockroachDB**

Before migration

![Image](https://github.com/user-attachments/assets/0cc4ab4b-aa8e-4380-b682-61a803285299)

After migration

![Image](https://github.com/user-attachments/assets/f88a8d71-398a-4202-b5a0-c20b2ed04cc2)

**Postgres**

Before migration

![Image](https://github.com/user-attachments/assets/7f45753b-6833-4c56-8918-8a3d3c3ec6ea)

After migration

![Image](https://github.com/user-attachments/assets/7ca01590-65ce-481f-96e6-5afcee411391)

@dantownsend What do you think about this attempt?

Jacky56:
@sinisaos thanks for the link, ill see if i could use it somewhere in the future, for now i was throwing raw sqls in my migrations instead as a placeholder

dantownsend:
This is fixed in `piccolo==1.26.0`.

PR #1183, opened 2025-04-17
On delete migrations

sinisaos:
Related to #1155

### DIFF
--- a/piccolo/apps/migrations/auto/migration_manager.py
+++ b/piccolo/apps/migrations/auto/migration_manager.py
@@ -534,6 +534,38 @@ class MigrationManager:
 
                 ###############################################################
 
+                on_delete = params.get("on_delete")
+                on_update = params.get("on_update")
+                if on_delete is not None or on_update is not None:
+                    # use to find FK table `referenced_table_name`` and
+                    # `referenced_column_name` for alter fk constraints
+                    existing_table = await self.get_table_from_snapshot(
+                        table_class_name=table_class_name,
+                        app_name=self.app_name,
+                    )
+                    # first drop fk constraint on alter table
+                    await self._run_query(
+                        _Table.alter().drop_foreign_key_constraint(
+                            column=alter_column.column_name
+                        )
+                    )
+                    # then add fk constraint on alter table
+                    for item in existing_table._meta.foreign_key_columns:
+                        reference: t.Any = item._foreign_key_meta.references
+                        referenced_table_name = reference._meta.tablename
+                        referenced_column_name = (
+                            reference._meta.primary_key._meta.name
+                        )
+                        await self._run_query(
+                            _Table.alter().add_foreign_key_constraint(
+                                column=alter_column.column_name,
+                                referenced_table_name=referenced_table_name,
+                                referenced_column_name=referenced_column_name,
+                                on_delete=on_delete,
+                                on_update=on_update,
+                            )
+                        )
+
                 null = params.get("null")
                 if null is not None:
                     await self._run_query(
--- a/piccolo/query/methods/alter.py
+++ b/piccolo/query/methods/alter.py
@@ -194,6 +194,7 @@ class AddForeignKeyConstraint(AlterStatement):
         "constraint_name",
         "foreign_key_column_name",
         "referenced_table_name",
+        "referenced_column_name",
         "on_delete",
         "on_update",
     )
@@ -201,9 +202,9 @@ class AddForeignKeyConstraint(AlterStatement):
     constraint_name: str
     foreign_key_column_name: str
     referenced_table_name: str
+    referenced_column_name: str
     on_delete: t.Optional[OnDelete]
     on_update: t.Optional[OnUpdate]
-    referenced_column_name: str = "id"
 
     @property
     def ddl(self) -> str:
@@ -273,8 +274,8 @@ class DropTable:
 
 class Alter(DDL):
     __slots__ = (
-        "_add_foreign_key_constraint",
         "_add",
+        "_add_foreign_key_constraint",
         "_drop_constraint",
         "_drop_default",
         "_drop_table",
@@ -488,7 +489,7 @@ class Alter(DDL):
     def _get_constraint_name(self, column: t.Union[str, ForeignKey]) -> str:
         column_name = AlterColumnStatement(column=column).column_name
         tablename = self.table._meta.tablename
-        return f"{tablename}_{column_name}_fk"
+        return f"{tablename}_{column_name}_fkey"
 
     def drop_constraint(self, constraint_name: str) -> Alter:
         self._drop_constraint.append(
@@ -500,15 +501,18 @@ class Alter(DDL):
         self, column: t.Union[str, ForeignKey]
     ) -> Alter:
         constraint_name = self._get_constraint_name(column=column)
-        return self.drop_constraint(constraint_name=constraint_name)
+        self._drop_constraint.append(
+            DropConstraint(constraint_name=constraint_name)
+        )
+        return self
 
     def add_foreign_key_constraint(
         self,
         column: t.Union[str, ForeignKey],
         referenced_table_name: str,
+        referenced_column_name: str,
         on_delete: t.Optional[OnDelete] = None,
         on_update: t.Optional[OnUpdate] = None,
-        referenced_column_name: str = "id",
     ) -> Alter:
         """
         Add a new foreign key constraint::
@@ -528,9 +532,9 @@ class Alter(DDL):
                 constraint_name=constraint_name,
                 foreign_key_column_name=column_name,
                 referenced_table_name=referenced_table_name,
+                referenced_column_name=referenced_column_name,
                 on_delete=on_delete,
                 on_update=on_update,
-                referenced_column_name=referenced_column_name,
             )
         )
         return self
@@ -579,9 +583,11 @@ class Alter(DDL):
             i.ddl
             for i in itertools.chain(
                 self._add,
+                self._add_foreign_key_constraint,
                 self._rename_columns,
                 self._rename_table,
                 self._drop,
+                self._drop_constraint,
                 self._drop_default,
                 self._set_column_type,
                 self._set_unique,
--- a/tests/apps/migrations/auto/test_migration_manager.py
+++ b/tests/apps/migrations/auto/test_migration_manager.py
@@ -618,6 +618,71 @@ class TestMigrationManager(DBTestCase):
             response = self.run_sync("SELECT * FROM manager;")
             self.assertEqual(response, [{"id": id[0]["id"], "name": "Dave"}])
 
+    @engines_only("postgres", "cockroach")
+    def test_alter_fk_on_delete_on_update(self):
+        """
+        Test altering whether a column is nullable with MigrationManager.
+        """
+        # before performing migrations - OnDelete.no_action
+        on_delete_type = Band.raw(
+            """
+            select confdeltype
+            from pg_constraint
+            join pg_class c on c.oid=conrelid
+            join pg_class p on p.oid=confrelid;
+            """
+        ).run_sync()
+        self.assertEqual(on_delete_type[0]["confdeltype"], b"a")
+
+        manager = MigrationManager(app_name="music")
+        manager.alter_column(
+            table_class_name="Band",
+            tablename="band",
+            column_name="manager",
+            db_column_name="manager",
+            params={
+                "on_delete": OnDelete.set_null,
+                "on_update": OnUpdate.set_null,
+            },
+            old_params={
+                "on_delete": OnDelete.no_action,
+                "on_update": OnUpdate.no_action,
+            },
+            column_class=ForeignKey,
+            old_column_class=ForeignKey,
+            schema=None,
+        )
+
+        asyncio.run(manager.run())
+
+        # after performing migrations - OnDelete.set_null
+        on_delete_type = Band.raw(
+            """
+            select confdeltype
+            from pg_constraint
+            join pg_class c on c.oid=conrelid
+            join pg_class p on p.oid=confrelid;
+            """
+        ).run_sync()
+
+        self.assertEqual(on_delete_type[0]["confdeltype"], b"n")
+
+        # Reverse
+        asyncio.run(manager.run(backwards=True))
+
+        # after performing reverse migrations we have
+        # OnDelete.no_action again
+        on_delete_type = Band.raw(
+            """
+            select confdeltype
+            from pg_constraint
+            join pg_class c on c.oid=conrelid
+            join pg_class p on p.oid=confrelid;
+            """
+        ).run_sync()
+
+        self.assertEqual(on_delete_type[0]["confdeltype"], b"a")
+
     @engines_only("postgres")
     def test_alter_column_unique(self):
         """


piccolo/query/methods/alter.py
```diff
@@ -488,7 +489,7 @@ def set_length(self, column: t.Union[str, Varchar], length: int) -> Alter:
     def _get_constraint_name(self, column: t.Union[str, ForeignKey]) -> str:
         column_name = AlterColumnStatement(column=column).column_name
         tablename = self.table._meta.tablename
-        return f"{tablename}_{column_name}_fk"
+        return f"{tablename}_{column_name}_fkey"
```
dantownsend, line 492: Why did we go from `_fk` to `_fkey`?

piccolo/query/methods/alter.py
```diff
@@ -500,15 +501,18 @@ def drop_foreign_key_constraint(
...
+        return self
 
     def add_foreign_key_constraint(
         self,
         column: t.Union[str, ForeignKey],
         referenced_table_name: str,
+        referenced_column_name: str,
         on_delete: t.Optional[OnDelete] = None,
         on_update: t.Optional[OnUpdate] = None,
-        referenced_column_name: str = "id",
```
dantownsend, line 511: Why are we removing the default value?

piccolo/apps/migrations/auto/migration_manager.py
```diff
@@ -534,6 +534,38 @@ async def _run_alter_columns(self, backwards: bool = False):
...
+                        app_name=self.app_name,
+                    )
+                    # first drop fk constraint on alter table
+                    await self._run_query(
+                        _Table.alter().drop_foreign_key_constraint(
+                            column=alter_column.column_name
+                        )
+                    )
+                    # then add fk constraint on alter table
+                    for item in existing_table._meta.foreign_key_columns:
```
dantownsend, line 553: I haven't had a chance to properly think about this yet - but why do we loop through all of the foreign key columns on the table?

dantownsend (commented):
I haven't had a chance to properly test this yet, but it's looking good so far - thanks!

piccolo/query/methods/alter.py
sinisaos: I changed this because FK constraints ends with `_fkey`, e.g. `band_manager_fkey` (not `_fk`) and because we first need to drop FK constraints that are not dropped with the `_fk` suffix because they do not exist in the database.

piccolo/query/methods/alter.py
sinisaos: I changed this because with the default value I get an `'AddForeignKeyConstraint' object attribute 'referenced_column_name' is read-only` error. I don't know another way to solve it and I remove the default value and put the `referenced_column_name` in the slots. If there is a better way, please change it. Thanks.

piccolo/query/methods/alter.py
dantownsend: The challenge with this  is the constraint name is generated by the database, so we're not in control of what the constraint name is (it could be different depending on the Postgres version).

Ideally we would explicitly set the name of the FK constraint when creating the table.

Alternatively, we could query the information schema tables to find the constraint name.

sinisaos: Isn't [fkey](https://stackoverflow.com/questions/4107915/postgresql-default-constraint-names/4108266#4108266) the default suffix for fk constraint? We are [test multiple Postgres](https://github.com/piccolo-orm/piccolo/blob/7cbd871c7b2fcdabbcfed0db9c69c6a72cdb3b37/.github/workflows/tests.yaml#L89) versions, and all test pass (also CockroachDB) which means that `fkey` is the standard suffix of fk constraints in all those versions.

dantownsend: OK, good point - I thought that `fk` was used in older versions of Postgres, but that doesn't seem to be the case.

piccolo/query/methods/alter.py
dantownsend: OK, fair enough - we should be passing in this value anyway.

piccolo/query/methods/alter.py
dantownsend: I wonder what happens if a column or table name is changed? Does the constraint name automatically update in the database?

dantownsend: I did a quick test, and it looks like the constraint name stays the same, which is annoying.

sinisaos: Yes, you are right. I didn't think about renaming the columns and tables. I also did a quick test and I found another problem with indexes and constraints names on renaming the columns and tables. If we have a table like this
```python
class Task(Table):
    name = Varchar()
    task_user = ForeignKey(
        BaseUser, on_delete=OnDelete.set_default, on_update=OnUpdate.set_default
    )
```
After we create and make migration, `indexes` and `fk constraints` it is like this which is correct
```bash
Indexes:
    "task_pkey" PRIMARY KEY, btree (id)
Foreign-key constraints:
    "task_task_user_fkey" FOREIGN KEY (task_user) REFERENCES piccolo_user(id) ON UPDATE SET DEFAULT ON DELETE SET DEFAULT
```
Now, if we rename the column `task_user` to `user_task` the fk constraints remain the same (because when renaming a column, only the renaming is performed without altering the column) and the fk constraints name are not changed.
After we create and make migration, `indexes` and `fk constraints` it is like this
```bash
Indexes:
    "task_pkey" PRIMARY KEY, btree (id)
Foreign-key constraints:
    # should be task_user_task_fkey
    "task_task_user_fkey" FOREIGN KEY (user_task) REFERENCES piccolo_user(id) ON UPDATE SET DEFAULT ON DELETE SET DEFAULT
```
After that we cannot drop the fk constraint because the fk constraint name is wrong (does not exist).

The same goes for renaming the table. If we change the table name to `Test` the indexes and fk constraints names are wrong.
```
Indexes:
    "task_pkey" PRIMARY KEY, btree (id) # should be test_pkey
Foreign-key constraints:
    # should be test_user_task_fkey and the constraint was not dropped because it did not exist in the db with that name    
    "task_task_user_fkey" FOREIGN KEY (user_task) REFERENCES piccolo_user(id) ON UPDATE SET DEFAULT ON DELETE SET DEFAULT
    # a new constraint is created which is fine, but this is not correct because the old tablename fk constraints are not 
    dropped because name issue
    "test_user_task_fkey" FOREIGN KEY (user_task) REFERENCES piccolo_user(id) ON UPDATE CASCADE ON DELETE CASCADE
```
Sorry for the long comments. I hope you understand what I mean (details are in the bash comments).

dantownsend: I think the only solution is to modify `drop_foreign_key_constraint`.

Here's a way of finding the name of a foreign key constraint:

https://stackoverflow.com/a/3907999

Since it uses the `information_schema` tables, it should work across different databases (e.g. MySQL if we ever support that in the future).

Ideally we could do `ALTER TABLE some_table DROP CONSTRAINT (SELECT ...)`. Where the sub query gets the name of the constraint, but this gives a syntax error.

Here's one potential solution:

https://stackoverflow.com/a/12396684

Probably the easiest thing to do is just add a select query before this line:

https://github.com/piccolo-orm/piccolo/blob/b5c14905328b64c381f2103b851a8d55a3107596/piccolo/apps/migrations/auto/migration_manager.py#L541

Which gets the constraint name, and then rather than using `drop_foreign_key_constraint`, we use `drop_constraint`, passing in the constraint name.

sinisaos: Or should we rename the `FK constraints` when we change the name of the fk column. This is just an idea and we should probably have a `RenameColumnConstraint` class in `query/methods/alter.py`
A working example is this
```python
async def _run_rename_columns(self, backwards: bool = False):
    for table_class_name in self.rename_columns.table_class_names:
        columns = self.rename_columns.for_table_class_name(table_class_name)

        if not columns:
            continue
        # get existing table
        existing_table = await self.get_table_from_snapshot(
            table_class_name=table_class_name,
            app_name=self.app_name,
        )

        _Table: t.Type[Table] = create_table_class(
            class_name=table_class_name,
            class_kwargs={
                "tablename": columns[0].tablename,
                "schema": columns[0].schema,
            },
        )

        for rename_column in columns:
            column = (
                rename_column.new_db_column_name
                if backwards
                else rename_column.old_db_column_name
            )
            new_name = (
                rename_column.old_db_column_name
                if backwards
                else rename_column.new_db_column_name
            )

            await self._run_query(
                _Table.alter().rename_column(
                    column=column,
                    new_name=new_name,
                )
            )
            # check if column is FK column
            is_fk_constraint = [
                i
                for i in existing_table._meta.foreign_key_columns
                if i._meta.name == new_name
            ]
            # if column is FK rename constraint, pass for all other columns
            if len(is_fk_constraint) > 0:
                old_constraint_name = f"{columns[0].tablename}_{column}_fkey"
                new_constraint_name = f"{columns[0].tablename}_{new_name}_fkey"
                await self._run_query(
                    _Table.raw(
                        f"""
                        ALTER TABLE {columns[0].tablename} 
                        RENAME CONSTRAINT {old_constraint_name} 
                        TO {new_constraint_name}
                        """
                    )
                )
```
This example works and the `FK constraint` name changes when renaming the FK column without affecting other columns. Something similar should be applied to `rename_table` to rename `PK constraints`. This is just an idea. That should be written much better. I hope that makes sense.

dantownsend: Yes, renaming the constraint when a column or table is renamed is an option.

I pushed a prototype for looking up the constraint name in the database. I haven't had a chance to test it.

sinisaos: I was able to try the latest changes and in my case the FK constraint names are changing correctly when renaming a column or table. I also added PK constraint name changes when renaming a table. If that's not good enough or it doesn't make sense, please change or remove that.

piccolo/query/methods/alter.py
```diff
@@ -372,6 +390,20 @@ def rename_table(self, new_name: str) -> Alter:
         self._rename_table = [RenameTable(new_name=new_name)]
         return self
 
+    def rename_pk_constraint(
```
dantownsend, line 393: I would change this to `rename_constraint`, as it could work for any kind of constraint.

piccolo/query/methods/alter.py
```diff
@@ -36,6 +36,21 @@ def ddl(self) -> str:
         return f"RENAME TO {self.new_name}"
 
 
+@dataclass
+class RenamePkConstraint(AlterStatement):
```
dantownsend, line 40: I would call this `RenameConstraint`.

piccolo/query/methods/alter.py
```diff
@@ -36,6 +36,21 @@ def ddl(self) -> str:
...
+    __slots__ = ("old_constraint_name", "new_constraint_name")
+
+    old_constraint_name: str
+    new_constraint_name: str
```
dantownsend, line 44: Lets change these to `old_name` and `new_name`. I don't think we need to be any more explicit than that.

piccolo/apps/migrations/auto/migration_manager.py
```diff
@@ -713,6 +747,13 @@ async def _run_rename_tables(self, backwards: bool = False):
                 },
             )
 
+            await self._run_query(
+                _Table.alter().rename_pk_constraint(
+                    old_constraint_name=f"{tablename}_pkey",
+                    new_constraint_name=f"{new_tablename}_pkey",
+                )
+            )
```
dantownsend, line 755: I think we should leave this out for now. Because renaming tables currently works with SQLite, but renaming constraints doesn't.

piccolo/query/methods/alter.py
dantownsend: OK, thanks.

I'll make a few final tweaks and will try and get it merged in.

piccolo/apps/migrations/auto/migration_manager.py
sinisaos: We could leave this and skip the renaming constraint for `Sqlite`. Something like this
```python
engine = engine_finder()

if engine.engine_type in ("postgres", "cockroach"):
    await self._run_query(
        _Table.alter().rename_constraint(
            old_constraint_name=f"{tablename}_pkey",
            new_constraint_name=f"{new_tablename}_pkey",
        )
    )

await self._run_query(
    _Table.alter().rename_table(new_name=new_tablename)
)
```

dantownsend: Yeah, that works, but I'm tempted to just leave out the renaming of the primary key constraint though - we don't need it for the `on delete` / `on update` changes. This is already a pretty big PR, and I'm always nervous about changing too much stuff with migrations.

sinisaos: Sorry. I will revert the latest commit.

dantownsend:
@sinisaos I'm going to merge this in. It's getting pretty big, and any remaining stuff can be added in future PRs. Thanks!

AlexanderMakarov:
I think that this code should make v2 of Piccolo ORM. Because it introduces backward incompatible change - it adds support of FK-s in `MigrationManager.alter_column(...)`. And due to before they haven't worked (at least properly) it causes migrations written for previous Piccolo versions to fail runtime (when trying to apply them).

For example I've met with a number of issues in our quite complex multi-tenant app (one "master" schema-tenant and multiple generated-on-create "regular tenant" schemas, PostgreSQL) and mostly auto-generated migrations caused by:
- quite questionable code
  ```python

    async def get_table_from_snapshot(
        self,
        table_class_name: str,
        app_name: t.Optional[str],
        offset: int = 0,
        migration_id: t.Optional[str] = None,
    ) -> t.Type[Table]:
        ...
        from piccolo.apps.migrations.commands.base import BaseMigrationManager
        ...
        ... BaseMigrationManager() ...
  ```
  which ignores "app_config" provided into `ForwardsMigrationManager.run_migrations(app_config)".
- `get_fk_constraint_name` function in new "constraints.py" file which gets schema name via `table._meta.schema or "public"` while it would be more Piccolo-ish (IMHO) to deduce constraint names from previous migrations (i.e. from code, not DB). Here we are getting "public" always because schema name is provided only via "app_config".

dantownsend:
@AlexanderMakarov I didn't intend this to be a breaking change - lets try and fix it so it works for your use case.

Can you give me a bit more detail on how you use it for multi-tenancy? For each tenant, do you pass in a different `AppConfig` to `run_migration` or something?

AlexanderMakarov:
>I didn't intend this to be a breaking change

Yes, I see, but for our use case it is quite a change.

For multi-tenancy we use quite a number of patches for Piccolo. We need to support 2 types of tenants:
- "master" - always one tenant/schema, different schema.
- "regular" - many tenants with schema names generated on "create tenant" events", the same schema for all tenants.

Therefore we have 2 pairs of files:
- "master": "piccolo_conf.py" file with DB and APP_REGISTRY which points on "piccolo_app.py" file. "piccolo_app.py" file declares TEMPLATE_APP_CONFIG and APP_CONFIG pointing on folder with "master" migrations. Also this file contains `patch_piccolo_migrations(APP_CONFIG, ...)` call (explained below). 
- "regular": "piccolo_conf_tenants.py" file is a wrapper around "piccolo_conf.py" (to have them both imported via `PICCOLO_CONF=app.db.piccolo_conf` pointing on "piccolo_conf.py") with its own DB and APP_REGISTRY which points on "piccolo_app_tenants.py" file. "piccolo_app_tenants.py" file declares TEMPLATE_APP_CONFIG and APP_CONFIG pointing on folder with "regular" migrations. With `patch_piccolo_migrations(APP_CONFIG, ...)` as well.

This complexity (4 files) mostly due to 2 use cases - when running Python app and when running something like `PICCOLO_CONF=app.db.piccolo_conf_tenants piccolo migrations new app.piccolo_app.master --auto --trace` (note that this is code for "regular" tenant migrations generation, not "master" one).

`patch_piccolo_migrations` executes:
- `Migration._meta.tablename = migration_table_name` (we are using custom table name).
- `forwards_mm = ForwardsMigrationManager(app_config.app_name)` and next loop `for module in forwards_mm.get_migration_modules(app_config.migrations_folder_path).values()` which patches each module with `setattr(module, "schema_name", schema_name)` and `setattr(module, "forwards", patched_forwards)`.
- Common `Migration._meta.schema = schema_name`.

When we creating a tenant (i.e. creating new schema run-time) then we are running almost the same `patch_piccolo_migrations` and next mentioned  `await forwards_mm.run_migrations(app_config)`.

When we are running a query then we use `piccolo.table.TableMeta.get_formatted_tablename = get_formatted_tablename` which takes schema name from ContextVar. ContextVar is managed by context manager (here we are close to usual SQLAlchemy-ish approach).

Tenants are separated by users as an extra fence for any data leaks.

So in general we have a lot of monkey-patching to set and replace required schema name inside Piccolo to manage migrations and execute queries with the same code. Also our SQL schema is quite complex too, 50+ tables and many FK-s between them.

Until 1.26.1 updates were quite seamless but here we have changes which use db directly as a source of data (not previous migrations) and ignores initially specified `app_name` for ForwardMigrationManagar (via this `BaseMigrationManager()` call).

dantownsend:
The reason we get the FK constraint name from the database is because we don't explicitly set the name of the constraint when creating the FK column. We could look back at the migrations, find when the FK column was created, and work out the FK constraint name based off the name of the column.

Also, I wonder if this will be fixed by using the `app_cofig` passed into `ForwardsMigrationManager.run_migrations(app_config)`.

We should really support multi-tenant out of the box. If you have any ideas let me know.

AlexanderMakarov:
>we don't explicitly set the name of the constraint when creating the FK column.

I see. But you are right about "work out the FK constraint name" - PostgreSQL generates quite good FK names and I believe (haven't checked thoroughly though) format is same since at least v9.6.

>Also, I wonder if this will be fixed by using the app_cofig passed into ForwardsMigrationManager.run_migrations(app_config).

This is what we are doing. Unfortunately `BaseMigrationManager()` ignores this parameter - see [BaseMigrationManager constructor which gets nothing](https://github.com/piccolo-orm/piccolo/blob/1e69d7b16bad0a31ba329e79581723f95df5195b/piccolo/apps/migrations/commands/base.py#L22) and [its `get_table_from_snapshot` realization which deduces `app_config` by optional `app_name`](https://github.com/piccolo-orm/piccolo/blob/1e69d7b16bad0a31ba329e79581723f95df5195b/piccolo/apps/migrations/commands/base.py#L128), i.e. something default - one per app, not per schema/tenant.

Issue #1192, opened 2025-05-12
BlackSheep integration tests failing


dantownsend:
https://github.com/piccolo-orm/piccolo/actions/runs/14974812809/job/42064322020

dantownsend:
@sinisaos have you go any idea what could be causing this?

sinisaos:
@dantownsend I don't know where the problem is for now. Uvicorn works with all versions, but `dummy_server` doesn't pass the integration test with the latest version of BlackSheep (`v2.3.0`). That latest version had some breaking changes and I'll try to figure it out tomorrow. For now we can pin the `BlackSheep` version to `<=2.2.0` and that works and the integration tests pass.

dantownsend:
@sinisaos Yes, version pinning will work for now. Hopefully there's a simple fix. Thanks!

dantownsend:
Fixed by https://github.com/piccolo-orm/piccolo/issues/1192

PR #1195, opened 2025-05-13
update BlackSheep template

sinisaos:
Related to #1192. In version 2.3.0 BlackSheep introduces the `apply_routes()` method to make the routes effective after the application is launched, so we need to update the BlackSheep template with that method for the integration test to work. More info [here](https://github.com/Neoteroi/BlackSheep/releases/tag/v2.3.0)


dantownsend:
@sinisaos Looks great - thanks a lot of looking into this!

PR #1201, opened 2025-05-23
Composite index

### LINKED ISSUE #1188: Compound indexes are not supported by (auto)-migrations
`MigrationManager` is missing `add_index` method, and there doesn't seem to be any way to specify a multi-column index should be created in the schema format.

sinisaos:
Related to #1188. This is almost identical PR as for [composite unique constraints](https://github.com/piccolo-orm/piccolo/pull/984), only it doesn't use custom ddl but modified Piccolo methods `create_index` and `drop_index`. Although it serves a different purpose, feel free to remove it if it represents too much code duplication and adds complexity, as adding and dropping a composite index can easily be done with manual migrations as described in the comment.

Usage is:

```python
from piccolo.columns import Varchar, Boolean
from piccolo.composite_index import CompositeIndex


class Task(Table):
    name = Varchar()
    composite_name_completed = CompositeIndex(["name", "completed"])
```


dantownsend (commented):
I like the API - using `CompositeIndex`.

I agree that using strings is just much easier than actual column references.

```python
class Album(Table):
    name = Varchar()
    released = Boolean(default=False)
    name_released_idx = CompositeIndex(["name", "released"])
```

Is there a way to specify the index type?

It looks like B-tree, GiST, GIN, and BRIN can be used with multi-column indexes:

https://www.postgresql.org/docs/current/indexes-multicolumn.html

sinisaos:
@dantownsend Great idea. I'll try to implement it. Thanks

sinisaos:
Tests pass locally without any problems. Maybe the next update with the master branch will be better.

sinisaos:
I'm closing this PR just to clean up my open requests in the repo. Since it has been open for quite a while, I'll remove it from the list to keep things tidy. Feel free to use it if needed. Thanks.

Issue #1143, opened 2025-01-16
Missing constraint while creating PK and unique column


mkorlaga:
Problem described causes errors when running migrations and modifying piccolo models because of missing constraint in database (postgres 17).

Tl;tr
Modifying unique argument on a column causes migration to generate dropping constraint that was not created in the first place.

Recreate steps:

1. Create Table with column A.
2. Set `primary_key = True` and `unique = True`

e.g. 
    some_uuid_field = UUID(
        default=UUID4(),
        null=False,
        primary_key=True,
        unique=True,
        index=True,
        index_method=IndexMethod.btree,
        db_column_name=None,
        secret=False,
    )

3. Create and run migration (table and column is created. Notice that in database only 1 constraint is created, which is PK for some_uuid_field column, which is obviously expected behavior)
4. On model modify `unique=True` -> `unique=False`.
5. Generate migration (notice that we have drop unique constraint step)
6. Run migration (causes error on trying to drop constraint that doesn't exist as it was not created in first place)

Solution:
1. When creating a pk column validate if passed arguments contain both pk and unique set to True. If so, then when generating migration for column creation squash unique value as pk is already set to True.
2. If both values are set to True, in migration create both constraints (doesn't make sense as we already would try to have column to be a pk which should imply being unique)

sinisaos:
@mkorlaga I think you are right. When we use a custom primary key (`UUID`, `Varchar` etc.) the index is created but not the unique constraint. After that the migrations do not work properly because we cannot drop a constraint that does not exist.

![Image](https://github.com/user-attachments/assets/d3e6434c-97f2-49f7-acb8-0260d6be8130)

I think the problem is in MigrationManager` _run_add_tables` method. 

https://github.com/piccolo-orm/piccolo/blob/1e69d7b16bad0a31ba329e79581723f95df5195b/piccolo/apps/migrations/auto/migration_manager.py#L816-L817

If I add changes like this, everything works.

```python
for _Table in sorted_table_classes:
    await self._run_query(_Table.create_table())
    # Missing unique constraint while creating custom PK 
    # and unique column set to True
    # https://github.com/piccolo-orm/piccolo/issues/1143
    for add_column in add_columns:
        if (
            not isinstance(add_column.column, Serial)
            and add_column.column._meta.params["unique"] is True
            and add_column.column._meta.params["primary_key"] is True
        ):
            await self._run_query(
                _Table.alter().set_unique(
                    column=add_column.column._meta.name,
[...]

PR #1202, opened 2025-05-30
check uniqueness of the custom pk

sinisaos:
Related to #1143


dantownsend:
I think the problem here is if someone does this:

```python
class MyTable(Table):
    pk = UUID(primary_key=True, unique=True)
```

A primary key is already unique, so `unique=True` is redundant.

If someone then does this for some reason:

```python
class MyTable(Table):
    pk = UUID(primary_key=True, unique=False)
```

The auto migration tries to drop the unique constraint, which doesn't exist. I think this is an unlikely scenario ... however ...

There are a few options:

1. Validate that if `primary_key=True` then `unique=False`
2. In this PR, you add the unique constraint, but as it's effectively the same thing as a primary key constraint anyway, it's kind of duplicating functionality.
3. If a column is a primary key, and the migration thinks it should drop a unique constraint then ignore it???

I'm not sure ... I need to think about it some more.

tests/apps/migrations/auto/integration/test_migrations.py
```diff
@@ -1086,7 +1086,7 @@ def test_target_column(self):
 class TestForeignKeySelf(MigrationTestCase):
     def setUp(self) -> None:
         class TableA(Table):
-            id = UUID(primary_key=True)
+            id = UUID(primary_key=True, unique=True)
```
sinisaos, line 1089: That doesn't seem right?

tests/apps/migrations/auto/integration/test_migrations.py
```diff
@@ -1136,14 +1136,16 @@ def test_add_column(self, get_app_config):
                 [
                     create_table_class(
                         class_name="MyTable",
-                        class_members={"id": UUID(primary_key=True)},
+                        class_members={
+                            "id": UUID(primary_key=True, unique=True)
```
sinisaos, line 1140: Also here.

tests/apps/migrations/auto/integration/test_migrations.py
```diff
@@ -1136,14 +1136,16 @@ def test_add_column(self, get_app_config):
...
+                        },
                     )
                 ],
                 [
                     create_table_class(
                         class_name="MyTable",
                         class_members={
-                            "id": UUID(primary_key=True),
+                            "id": UUID(primary_key=True, unique=True),
```
sinisaos, line 1148: Also here.

sinisaos:
Yes, you are right. I did not think about it in this way and I was guided by the statement from issue.

> There are a few options:
> 
> 1. Validate that if `primary_key=True` then `unique=False`

I also think we should prevent adding `primary_key=True` and `unique=False` in column definition by adding a `ValueError` so there is no confusion that this is possible. I have added a new commit with that check (I don't know if it's good, it's just an attempt) and everything works except migrations when the table is a reference to itself and I need to add `unique=True` to the primary key to make the test pass. If that doesn't make sense, I'll close this PR if you agree, because as you say, this whole thing is an unlikely scenario.

PR #1203, opened 2025-05-30
Invalid column arguments

### LINKED ISSUE #1114: piccolo args to allow only valid arguments
its let you say  `Float(nul=True)` when it should raise invalid argument

sinisaos:
Resolves #1114


dantownsend:
I think this is an interesting idea.

The only downside is that it doesn't help with auto completion in editors. To do that we would have to declare all of the possible options in `__init__` for each `Column` subclass.

```python
class Boolean(Column):
    def __init__(
        self,
        null: bool = False,
        primary_key: bool = False,
        unique: bool = False,
        ...
    ):
        ...
```

We could also try using `Unpack`, to declare the types of the `kwargs`:

https://docs.python.org/3/library/typing.html#typing.Unpack

dantownsend:
> and then autocomplete works when we want to write the argument to the columns in the table definition. The biggest problem with this approach is that `Unpack` is only available in `Python 3.11 and later`, but we also support `Python 3.9 and 3.10`.

You should be able to import it from `typing_extensions`.

sinisaos:
> > and then autocomplete works when we want to write the argument to the columns in the table definition. The biggest problem with this approach is that `Unpack` is only available in `Python 3.11 and later`, but we also support `Python 3.9 and 3.10`.
> 
> You should be able to import it from `typing_extensions`.

Yes, you are right. Thanks.

I made an [experimental commit](https://github.com/piccolo-orm/piccolo/pull/1203/commits/9e27722fa18f167663a4a0734cc132e75692ecc7) (can be made in a separate PR, if needed). Using `TypeDict` and `Unpack` we have improvements in autocompletion. Main issue is that I had to mark all lines `kwargs.update` with `# type: ignore` because of Mypy error `Unexpected TypedDict keys`. I also tried with `total=False` but the error remains because keys from `Column.__init__` (`default`, `length` etc.) are missing in `TypeDict`. We need these lines to properly update kwargs with `Column.__init__` values.

dantownsend:
> I made an [experimental commit](https://github.com/piccolo-orm/piccolo/pull/1203/commits/9e27722fa18f167663a4a0734cc132e75692ecc7) (can be made in a separate PR, if needed). Using `TypeDict` and `Unpack` we have improvements in autocompletion. Main issue is that I had to mark all lines `kwargs.update` with `# type: ignore` because of Mypy error `Unexpected TypedDict keys`. I also tried with `total=False` but the error remains because keys from `Column.__init__` (`default`, `length` etc.) are missing in `TypeDict`. We need these lines to properly update kwargs with `Column.__init__` values.

Thanks! Instead of the `type: ignore`, we can do something like this:

```python
super().__init__(default=default, **kwargs)
```

I am very optimistic about this `Unpack` approach. It's not just useful for autocompletion - VSCode, MyPy etc will detect typos. We might be able to remove the run time checks (i.e. raising `ValueError`), as linters should detect any misspelt column args before it reaches production.

piccolo/columns/column_types.py
```diff
@@ -77,6 +80,23 @@ class Band(Table):
     from piccolo.table import Table
 
 
+class Arguments(t.TypedDict, total=False):
```
dantownsend, line 83: I would call this something more descriptive, like `ColumnKwargs`.

tests/columns/test_boolean.py
```diff
@@ -6,7 +6,7 @@
 
 
 class MyTable(Table):
-    boolean = Boolean(boolean=False, null=True)
+    boolean = Boolean(default=False, null=True)
```
dantownsend, line 9: Good catch!

piccolo/columns/column_types.py
```diff
@@ -77,6 +80,23 @@ class Band(Table):
...
+    unique: NotRequired[bool]
+    index: NotRequired[bool]
+    index_method: NotRequired[IndexMethod]
+    required: NotRequired[bool]
+    help_text: NotRequired[str]
+    choices: NotRequired[t.Type[Enum]]
+    secret: NotRequired[bool]
+    auto_update: NotRequired[t.Any]
+    digits: NotRequired[t.Tuple[int, int]]
+    db_column_name: NotRequired[str]
```
dantownsend, line 97: Do you need `NotRequired`? As you've specified `total=False`?

sinisaos: > I am very optimistic about this Unpack approach. It's not just useful for autocompletion - VSCode, MyPy etc will detect typos. We might be able to remove the run time checks (i.e. raising ValueError), as linters should detect any misspelt column args before it reaches production.

I made another commit with your suggestions and removed the runtime typos checking (raising `ValueError` for invalid column arguments). I still think it would be useful and if you think so too, I can revert it. If not, we can leave it as is and let linters to catch those typos.

piccolo/columns/column_types.py
```diff
@@ -77,6 +80,23 @@ class Band(Table):
     from piccolo.table import Table
 
 
+class ColumnKwargs(t.TypedDict, total=False):
+    null: bool
+    primary_key: bool
+    primary: bool
+    key: bool
```
dantownsend, line 87: I would drop these two. We don't want people passing them in any more.

tests/columns/test_base.py
```diff
@@ -45,6 +45,16 @@ def test_secret_parameter(self):
...
 
 
+class TestInvalidColumnArgument(TestCase):
+    def test_invalid_arguments(self):
+        """
+        Test invalid argument to a column.
+        """
+        with self.assertRaises(ValueError):
+            column = Varchar(nul=True)
+            self.assertEqual(column._meta.name, column)
```
dantownsend, line 55: I think we can drop this for now.

piccolo/columns/column_types.py
sinisaos: I would remove it if you agree to change this test.

https://github.com/piccolo-orm/piccolo/blob/1e69d7b16bad0a31ba329e79581723f95df5195b/tests/columns/test_primary_key.py#L77-L85

to `primary_key=True`. Otherwise we get a Mypy error.

sinisaos: I made a new commit with the latest changes. I hope it's okay.

Issue #1208, opened 2025-06-06
Use ``built-in`` types and ``collections`` for type hints


sinisaos:
Since we are using Python 3.9 and above, I think we should use [built-in types and collections for type hints](https://docs.python.org/3/library/typing.html#deprecated-aliases). This can also be done in the Piccolo API. For Piccolo Admin, we need to first drop Python 3.8, and then we can update the type hints in that repository as well. 
Can I make these changes if you agree with them? It looks like a lot of changes, but nothing changes in the logic of the library. Here's the [branch](https://github.com/piccolo-orm/piccolo/compare/master...sinisaos:piccolo:improve_type_hints).

dantownsend:
Implemented in https://github.com/piccolo-orm/piccolo/pull/1209

PR #1209, opened 2025-06-08
Improve type hints

sinisaos:
Related to #1208

### DIFF
--- a/piccolo/apps/app/commands/new.py
+++ b/piccolo/apps/app/commands/new.py
@@ -5,7 +5,7 @@ import os
 import pathlib
 import string
 import sys
-import typing as t
+from typing import Any
 
 import black
 import jinja2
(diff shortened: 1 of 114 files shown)


dantownsend (approved):
Looks great - thanks!

This is probably the largest PR ever merged into this repo.

sinisaos:
Yes, it was a pretty big PR (without any changes to the library logic), but it was actually easy to do. I just make a simple script that maps the changes to a dictionary, then iterates through the directories and replaces the key-value pairs in the all `.py files` from that dictionary. Then the script runs a ruff linter in the subprocess to remove unnecessary imports. I ended up just running the Piccolo linters (there was very little manual changes after that) and that was it. I used the same script for all Piccolo repositories. There is still Piccolo API repo left and I will make PR today. I think this is useful because with these changes we prevent deprecation warnings in the future.

Issue #1212, opened 2025-06-24
Add an easy way to remove Array value in the style of Array.cat


Quik2007:
Let's say I want to remove some value from multiple db objects, e.g. remove "apple" as a favorite fruit from all users.

Something like this would be nice:
```py
await Person.update(
    {
        Person.favorite_fruits.remove(["apple"])
    }
).where(
    Person.favorite_fruits.any("apple")
)
```

Here, an additional `Array.remove` method in the style of `Array.cat` would arguably create more friendly/easier code (I honestly don't even know what the best way to do this would be right now, maybe using `M2M` even though it's just a list of values)

sinisaos:
@Quik2007 For now, you can use [Piccolo QueryString](https://github.com/piccolo-orm/piccolo/blob/73651e0e118056bf1adbde6aed3cf167cb98ea66/piccolo/querystring.py#L58) and the Postgres `array_remove` method to remove an item from an array. Something like this

```python
from piccolo.querystring import QueryString

await Person.update(
    {
        Person.favorite_fruit: QueryString(
            f'array_remove("{Person.favorite_fruit._meta.name}", {{}})',
            "apple",
        )
    }
).where(Person.favorite_fruit.any("apple"))
```
A simpler usage would be (as you said in your comment) to add a `remove` method to the `Array` column. This could be done with something like this

```python
# in piccolo/columns/colummn_types.py

def remove(self, value: t.Any) -> QueryString:
    """
    Used in an ``update`` query to remove item from array.

    .. code-block:: python

        >>> await Ticket.update({
        ...     Ticket.seat_numbers: Ticket.seat_numbers.remove(1000)
        ... }).where(Ticket.id == 1)


    """
    engine_type = self._meta.engine_type
    if engine_type != "postgres" and engine_type != "cockroach":
        raise ValueError(
            "Only Postgres and Cockroach support array appending."
        )

    db_column_name = self._meta.db_column_name
    return QueryString(f'array_remove("{db_column_name}", {{}})', value)

def __delete__(self, value: t.Any) -> QueryString:
    return self.remove(value)
```
I hope that makes sense.

sinisaos:
@dantownsend The query function could be something like this. 

```python
# in piccolo/query/functions/array.py

from piccolo.columns.base import Column
from piccolo.querystring import QueryString
from typing import Any


class ArrayRemove(QueryString):
    def __init__(self, column: Column, value: Any):
        """
        Remove item from array.

        """
[...]

PR #1213, opened 2025-06-27
add some useful array functions

sinisaos:
Some useful array functions. Related to #1212

### DIFF
--- a/docs/src/piccolo/functions/array.rst
+++ b/docs/src/piccolo/functions/array.rst
@@ -1,0 +1,26 @@
+Array functions
+===============
+
+.. currentmodule:: piccolo.query.functions.array
+
+.. note:: This is for Postgres / Cockroach only.
+
+ArrayAppend
+-----------
+
+.. autoclass:: ArrayAppend
+
+ArrayPrepend
+------------
+
+.. autoclass:: ArrayPrepend
+
+ArrayReplace
+------------
+
+.. autoclass:: ArrayReplace
+
+ArrayRemove
+-----------
+
+.. autoclass:: ArrayRemove
--- a/docs/src/piccolo/functions/index.rst
+++ b/docs/src/piccolo/functions/index.rst
@@ -12,3 +12,4 @@ Functions can be used to modify how queries are run, and what is returned.
     ./datetime
     ./type_conversion
     ./aggregate
+    ./array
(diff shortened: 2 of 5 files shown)


dantownsend:
@sinisaos This looks great, thanks a lot!

dantownsend:
@sinisaos In the issue you originally suggested doing something like this:

```python
def remove(self, value: t.Any) -> QueryString:
    """
    Used in an ``update`` query to remove item from array.

    .. code-block:: python

        >>> await Ticket.update({
        ...     Ticket.seat_numbers: Ticket.seat_numbers.remove(1000)
        ... }).where(Ticket.id == 1)


    """
    engine_type = self._meta.engine_type
    if engine_type != "postgres" and engine_type != "cockroach":
        raise ValueError(
            "Only Postgres and Cockroach support array appending."
        )

    db_column_name = self._meta.db_column_name
    return QueryString(f'array_remove("{db_column_name}", {{}})', value)

def __delete__(self, value: t.Any) -> QueryString:
    return self.remove(value)
```

On reflection, I think this makes sense (especially as we already have `cat`). But the method could under the hood just call the new function.

```python
def remove(self, value: t.Any) -> QueryString:
    return ArrayRemove(self, value)
```

Instead of the `__delete__` magic method, you could try `__sub__`. So it works like this:

```python
await Ticket.update({
    Ticket.seat_numbers: Ticket.seat_numbers - [1000]
}).where(Ticket.id == 1)
```

sinisaos:
@dantownsend I think we should decide if we want to use array functions like any other query functions (aggregation, math, etc.) or we should use `Array` column methods like `cat`. Adding both array functions and array methods confuses me (and will probably confuse every other user) that we can do the same thing in two ways. I would remove the `cat` method (possibly leave the `cat` method if needed for backward compatibility) and create an `ArrayCat` class and use it like other array query functions. If we go with `Array` column methods, I don't see the point of array query functions (in `query/functions/array.py`) because we can do all of that in array column methods and have everything in one place. I can update the PR, but I need to know which way we're going.

sinisaos:
@dantownsend I have created a #1214 with an alternative approach using `Array` column methods for array functions. That seems like a better approach, but I leave it to you to decide which approach suits you better.

PR #1217, opened 2025-07-12
Allow a subquery to be passed into `is_in` and `not_in`

sinisaos:
Resolves https://github.com/piccolo-orm/piccolo/issues/785

Based on #786. Also adds a subquery to the `not_in` clause, tests and docs. These subqueries will be useful in `m2m` columns to [replace this code](https://github.com/piccolo-orm/piccolo/blob/master/piccolo/columns/m2m.py#L376-L394) with a subquery with just one database query (and no additional empty list check) like this
```python
# use a subquery to make only one db query (for efficiency)
results = await secondary_table.objects().where(
    secondary_table._meta.primary_key.is_in(
        joining_table.select(
            getattr(
                self.m2m._meta.secondary_foreign_key,
                secondary_table._meta.primary_key._meta.name,
            )
        ).where(self.m2m._meta.primary_foreign_key == self.row)
    )
)
```

### DIFF
--- a/piccolo/columns/base.py
+++ b/piccolo/columns/base.py
@@ -600,7 +600,7 @@ class Column(Selectable):
 
         return True
 
-    def is_in(self, values: Union[Select, list[Any]]) -> Where:
+    def is_in(self, values: Union[Select, QueryString, list[Any]]) -> Where:
         from piccolo.query.methods.select import Select
 
         if isinstance(values, list):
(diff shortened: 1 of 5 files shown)


dantownsend (commented):
Thanks for this. Sub selects is something I really want to add.

I'll try this PR out.

piccolo/columns/base.py
```diff
@@ -599,18 +600,38 @@ def _validate_choices(
 
         return True
 
-    def is_in(self, values: list[Any]) -> Where:
-        if len(values) == 0:
-            raise ValueError(
-                "The `values` list argument must contain at least one value."
-            )
+    def is_in(self, values: Union[Select, list[Any]]) -> Where:
```
dantownsend, line 603: I'd like to change this up a bit, so rather than passing a `Select` into `is_in` we directly pass the `Querystring`. 

It makes it a bit more flexible.

dantownsend:
I think this is pretty much ready to go - we just need a couple of more tests.

We have some tests to make sure the generated SQL is correct, but we need some sanity checks to make sure the query works.

dantownsend:
@sinisaos Thanks for this 👍

Issue #1228, opened 2025-07-18
Add CockroachDB support to the playground


dantownsend:
We currently just support Postgres and SQLite:

https://github.com/piccolo-orm/piccolo/blob/51242469afa158e7800a8604567187295f33edf2/piccolo/apps/playground/commands/run.py#L297-L298

Even though the `PostgresEngine` and `CockroachEngine` are very similar, it's still best to use the correct engine.

PR #1229, opened 2025-07-23
adds CockroachDB as an option to the Piccolo playground

sinisaos:
Related to #1228. Also add `orjson` to the `Lilya` asgi template as it seems that `orjson` is now a hard dependency for `Lilya` (without that change our integration test would fail for Lilya with `ModuleNotFoundError: No module named 'orjson'`).

### DIFF
--- a/piccolo/apps/asgi/commands/new.py
+++ b/piccolo/apps/asgi/commands/new.py
@@ -15,7 +15,7 @@ ROUTER_DEPENDENCIES = {
     "blacksheep": ["blacksheep"],
     "litestar": ["litestar"],
     "esmerald": ["esmerald"],
-    "lilya": ["lilya", "orjson"],
+    "lilya": ["lilya"],
     "quart": ["quart", "quart_schema"],
     "falcon": ["falcon"],
     "sanic": ["sanic", "sanic_ext"],
--- a/piccolo/apps/asgi/commands/new.py
+++ b/piccolo/apps/asgi/commands/new.py
@@ -15,7 +15,7 @@ ROUTER_DEPENDENCIES = {
     "blacksheep": ["blacksheep"],
     "litestar": ["litestar"],
     "esmerald": ["esmerald"],
-    "lilya": ["lilya"],
+    "lilya": ["lilya", "orjson"],
     "quart": ["quart", "quart_schema"],
     "falcon": ["falcon"],
     "sanic": ["sanic", "sanic_ext"],
--- a/piccolo/apps/playground/commands/run.py
+++ b/piccolo/apps/playground/commands/run.py
@@ -25,7 +25,7 @@ from piccolo.columns import (
     Varchar,
 )
 from piccolo.columns.readable import Readable
-from piccolo.engine import PostgresEngine, SQLiteEngine
+from piccolo.engine import CockroachEngine, PostgresEngine, SQLiteEngine
 from piccolo.engine.base import Engine
 from piccolo.table import Table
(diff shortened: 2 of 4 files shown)


piccolo/apps/asgi/commands/new.py
```diff
@@ -15,7 +15,7 @@
     "blacksheep": ["blacksheep"],
     "litestar": ["litestar"],
     "esmerald": ["esmerald"],
-    "lilya": ["lilya"],
+    "lilya": ["lilya", "orjson"],
```
dantownsend, line 18: Interesting that `lilya` would do this without having it as one of their own dependencies, but I'm OK with adding this if it works. 👍

piccolo/apps/playground/commands/run.py
```diff
@@ -329,6 +330,16 @@ def run(
...
             }
         )
+    elif engine.upper() == "COCKROACH":
+        db = CockroachEngine(
+            {
+                "host": host,
+                "database": database,
+                "user": "root",
+                "password": "",
+                "port": 26257,
```
dantownsend, line 340: Port is kind of problematic, because we do allow the user to pass it in as an arg. I think we need to make port a nullable argument, which defaults to `None`. And if not passed in, then it's `5432` for Postgres and `26257` for Cockroach.

dantownsend (commented):
Looking good 👍

piccolo/apps/playground/commands/run.py
dantownsend: Yes, I think we should give the user the option to specify the port. So like you've done, except we can simplify it:

```python
def run(
    ...
    port: Optional[int] = None,
    ...
):
    if engine.upper() == "POSTGRES":
        db: Engine = PostgresEngine(
            {
                ...
                "port": port or 5432,
            }
        )
    elif engine.upper() == "COCKROACH":
        db = CockroachEngine(
            {
                ...
                "port": port or 26257,
            }
        )
```

sinisaos: Ok, I'll change that. Thanks.

piccolo/apps/playground/commands/run.py
```diff
@@ -326,7 +328,17 @@ def run(
...
+                "port": port or 5432,
+            }
+        )
+    elif engine.upper() == "COCKROACH":
+        db = CockroachEngine(
+            {
+                "host": host,
+                "database": database,
+                "user": "root",
+                "password": "",
```
dantownsend, line 340: We have a `password` param too, so we might need to do the same as with `port` - make it optional, and if not specified then default to an empty string?

Or in the docs tell them to create the database with the same password as Postgres (`piccolo`).

piccolo/apps/playground/commands/run.py
```diff
@@ -288,14 +289,15 @@ def run(
...
 ):
     """
     Creates a test database to play with.
 
     :param engine:
-        Which database engine to use - options are sqlite or postgres
+        Which database engine to use - options are sqlite, postgres or
+        cockroach
     :param user:
         Postgres user
```
dantownsend, line 302: We need to update these params - e.g. `Postgres/CockroachDB user`.

Or `Database user (ignored for SQLite)`?

piccolo/apps/asgi/commands/new.py
sinisaos: Reverted this because [Lilya now uses standard library json and no longer orjson](https://github.com/dymmond/lilya/releases/tag/0.18.1).

Issue #865, opened 2023-07-14
name 'Serial' is not defined


hoosnick:
When creating a new migration, I think this bug(error) only occurs with ForeignKeys, details below:
```
my_app/tables.py
```
```python
class Order(Table, tablename="orders"):
    user = ForeignKey(
        LazyTableReference(
            table_class_name="Users",
            module_path="users.tables"
        )
    )
    tariff = ForeignKey(
        LazyTableReference(
            table_class_name="Tariff",
            module_path="course.tables"
        )
    )
```
```
my_app/piccolo_migrations/my_app_2023_06_30t18_31_03_962663.py
```
<img src='https://github.com/piccolo-orm/piccolo/assets/73847672/01d68eba-d3a1-401e-a2d3-e588e926b9f6' width='500'>

***
<details><summary><b>Traceback</b></summary>

```
The command failed.
name 'Serial' is not defined
Traceback (most recent call last):
  File "path\to\project\env\Lib\site-packages\targ\__init__.py", line 448, in run      
    command.call_with(arg_class)
  File "path\to\project\env\Lib\site-packages\targ\__init__.py", line 229, in call_with
    asyncio.run(self.command(**cleaned_kwargs))
  File "path\to\programs\Python\Python311\Lib\asyncio\runners.py", line 190, in run
    return runner.run(main)
           ^^^^^^^^^^^^^^^^
  File "path\to\programs\Python\Python311\Lib\asyncio\runners.py", line 118, in run
    return self._loop.run_until_complete(task)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "path\to\programs\Python\Python311\Lib\asyncio\base_events.py", line 653, in run_until_complete
    return future.result()
           ^^^^^^^^^^^^^^^
  File "path\to\project\env\Lib\site-packages\piccolo\apps\migrations\commands\forwards.py", line 159, in forwards
    response = await run_forwards(
               ^^^^^^^^^^^^^^^^^^^
  File "path\to\project\env\Lib\site-packages\piccolo\apps\migrations\commands\forwards.py", line 120, in run_forwards
    response = await manager.run()
               ^^^^^^^^^^^^^^^^^^^
  File "path\to\project\env\Lib\site-packages\piccolo\apps\migrations\commands\forwards.py", line 97, in run
    return await self.run_migrations(app_config)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #1231, opened 2025-07-28
fixed missing primary key import for LazyTableReference

sinisaos:
Related to #865. Adds missing primary key to `extra_imports` when creating a migration with a ForeignKey that uses a LazyTableReference.

### DIFF
--- a/docs/src/piccolo/migrations/create.rst
+++ b/docs/src/piccolo/migrations/create.rst
@@ -79,8 +79,8 @@ If you want to run raw SQL within your migration, you can do so as follows:
     from piccolo.table import Table
 
 
-    ID = "2022-02-26T17:38:44:758593"
-    VERSION = "0.69.2"
+    ID = "2025-07-28T09:51:54:296860"
+    VERSION = "1.27.1"
     DESCRIPTION = "Updating each band's popularity"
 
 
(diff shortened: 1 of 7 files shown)


piccolo/apps/migrations/auto/serialisation.py
```diff
@@ -667,6 +667,21 @@ def serialise_params(params: dict[str, Any]) -> SerialisedParams:
...
+                Import(
+                    module=primary_key_class.__module__,
+                    target=primary_key_class.__name__,
+                    expect_conflict_with_global_name=getattr(
+                        UniqueGlobalNames,
+                        f"COLUMN_{primary_key_class.__name__.upper()}",
+                        None,
+                    ),
+                )
+            )
```
dantownsend, line 684: Thanks for this - I tested it, and it works.

I think the problem goes even deeper though.

Let's say you have this (note the `UUID` primary key):

```python
class TableA(Table):
    table_b = ForeignKey(LazyTableReference('TableB', module_path=__module__))

class TableB(Table):
    id = UUID(primary_key=True)
```

If you create the migration, it now has the import for `UUID`, but it's missing the import for `UUID4` (the default value for `UUID`).

The problem is, we use `SerialisedTableType`, which itself calls `serialise_params`. We need to access the `extra_imports` from that `serialise_params` call.

Sorry, it sounds very confusing.

But basically, I think the ideal solution would be this:

```python
serialised_table_type = SerialisedTableType(table_type=table_type)
extra_definitions.append(serialised_table_type)
extra_imports.extend(serialised_table_type.extra_imports)
```

So we need to refactor `SerialisedTableType` slightly so can access its extra imports.

Does that make sense?

Issue #770, opened 2023-02-22
Document `MigrationManager.get_table_from_snapshot`


dantownsend:
We have a feature which is currently undocumented.

By using `MigrationManager.get_table_from_snapshot` you can get a table from the migration history.

```python
from piccolo.apps.migrations.auto.migration_manager import MigrationManager
from piccolo.table import Table


ID = "2023-02-09T12:53:36:698501"
VERSION = "0.105.0"
DESCRIPTION = "Delete unpopular bands."


async def forwards():
    manager = MigrationManager(
        migration_id=ID, app_name="", description=DESCRIPTION
    )

    async def run():
        Band = await manager.get_table_from_snapshot('Band')
        await Band.delete().where(Band.popularity < 10)

    manager.add_raw(run)

    return manager
```

In the docs we recommend not directly importing a `Table` from `tables.py` into a migration file, as the migrations can break in the future if the `Table` being imported is deleted from `tables.py`.

The solution recommended in the docs is to copy and paste the `Table` from `tables.py` into the migration file (which is generally OK, but a bit annoying if the table is very large, with lots of foreign keys). An alternative is using `MigrationManager.get_table_from_snapshot` instead.

sinisaos:
Resolved by #1232

PR #1232, opened 2025-07-29
Documenting ``MigrationManager.get_table_from_snapshot``

sinisaos:
Resolves to #770


dantownsend:
> No problem. I went through some old issues and closed the ones that were completed or duplicated. I made a few PRs for the still current issues including this one, because I think it's very useful. I hope that's okay with you.

That's great, thank you!

I'll do a release now, because there's a lot of stuff waiting to be released. Will merge some more stuff in later.

docs/src/piccolo/migrations/create.rst
```diff
@@ -198,6 +198,40 @@ it's better to copy the relevant tables into your migration file:
...
+        )
+
+        async def run():
+            # We get a table from the migration history.
+            Task = await manager.get_table_from_snapshot(
+                app_name="home", table_class_name="Task"
+            )
+            await Task.raw(
+                "CREATE UNIQUE INDEX unique_name_completed ON task(name, completed)"
+            )
```
dantownsend, line 229: I think it would be nice to show how this table can do more than raw queries.

How about this as an example:

```python
Band = await manager.get_table_from_snapshot(
    app_name="music", table_class_name="Band"
)
await Band.update({"popularity": 1000}, force=True)
```

Which mirrors the example above it, except without having to manually define the tables.

sinisaos: Implemented your suggestions. Thanks

dantownsend:
Looks great, thanks!

Issue #554, opened 2022-06-28
Add `readable` and M2M examples to playground


dantownsend:
When I'm experimenting with Piccolo queries I typically use the playground:

```
piccolo playground run
```

I try and make the schema used by the playground as fully featured as possible, so I can experiment with most Piccolo features.

Some features the playground schema doesn't have which need adding:

- [ ] Readable for each table (see [docs](https://piccolo-orm.readthedocs.io/en/latest/piccolo/schema/advanced.html#readable))
- [ ] M2M (see [docs](https://piccolo-orm.readthedocs.io/en/latest/piccolo/schema/m2m.html))

PR #1234, opened 2025-07-29
adds m2m column to the playground

sinisaos:
Related to #554. Also a little refactoring for the `M2M` test.

### DIFF
--- a/piccolo/apps/playground/commands/run.py
+++ b/piccolo/apps/playground/commands/run.py
@@ -12,6 +12,7 @@ from typing import Optional
 
 from piccolo.columns import (
     JSON,
+    M2M,
     UUID,
     Array,
     Boolean,
@@ -19,6 +20,7 @@ from piccolo.columns import (
     ForeignKey,
     Integer,
     Interval,
+    LazyTableReference,
     Numeric,
     Serial,
     Text,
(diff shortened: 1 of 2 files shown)


tests/columns/m2m/base.py
```diff
@@ -51,54 +51,25 @@ def _setUp(self, schema: Optional[str] = None):
...
-            Band.insert(
-                Band(name="Pythonistas"),
-                Band(name="Rustaceans"),
-                Band(name="C-Sharps"),
-            ).run_sync()
+        bands = Band.insert(
+            Band(name="Pythonistas"),
+            Band(name="Rustaceans"),
+            Band(name="C-Sharps"),
+        ).run_sync()
```
dantownsend, line 58: Interesting - this is much cleaner, thanks!

piccolo/apps/playground/commands/run.py
```diff
@@ -161,6 +163,39 @@ def get_readable(cls) -> Readable:
...
+            template="%s",
+            columns=[cls.name],
+        )
+
+
+class GenreToMovie(Table):
+    id: Serial
+    movie = ForeignKey(Movie)
+    genre = ForeignKey(Genre)
+    reason = Text(help_text="For testing additional columns on join tables.")
```
dantownsend, line 196: I would change this to `GenreToBand`, as we already have a `Band` table.

Or `GenreToAlbum`, whichever makes more sense.

dantownsend (commented):
Thanks for this 👍

piccolo/apps/playground/commands/run.py
sinisaos: I'll use `GenreToBand` because as you say the `Band` table is already here. Thanks

dantownsend:
Looks great, thanks! 👍

Issue #918, opened 2024-01-10
Add columns with `secret=True` parameter to `Table._meta.secret_columns`


patrickneise:
In working with `PydanticDTO` from [litestar](https://github.com/litestar-org/litestar), realized that columns with `secret=True` parameter were not being added to `Table._meta.secret_columns`

Seems like a relatively small change:

```python
# piccolo/table.py line 311
if isinstance(column, Secret) or column._meta.params.get("secret"):
    secret_columns.append(column)
```

With new test:

```python
# tests/table/test_metaclass.py
def test_secret_columns_with_param(self):
    """
    Make sure TableMeta.secret_columns include columns with `secret=True`
    """

    class Classified(Table):
        top_secret = Varchar(secret=True)

    self.assertEqual(
        Classified._meta.secret_columns, [Classified.top_secret]
```

Changes staged for PR [here](https://github.com/patrickneise/piccolo/tree/secret_columns).

dantownsend:
Sorry for the slow response on this. I was mulling over what to do.

I think your changes makes perfect sense - if you can open a PR, I'll merge it in. Thanks!

sinisaos:
Resolved by #1235

PR #1235, opened 2025-07-30
Adds a column with the argument ``secret=True`` to ``Table._meta.secret_columns``

sinisaos:
Related to #918. Code written by @patrickneise ([here](https://github.com/piccolo-orm/piccolo/compare/master...patrickneise:piccolo:secret_columns)) but never create PR. Thanks @patrickneise.

### DIFF
--- a/scripts/test-cockroach.sh
+++ b/scripts/test-cockroach.sh
@@ -5,7 +5,7 @@
 # To run a single test tests/test_foo.py::TestFoo::test_foo
 
 export PICCOLO_CONF="tests.cockroach_conf"
-python3 -m pytest \
+python -m pytest \
     --cov=piccolo \
     --cov-report=xml \
     --cov-report=html \
--- a/tests/utils/test_pydantic.py
+++ b/tests/utils/test_pydantic.py
@@ -29,10 +29,10 @@ from piccolo.utils.pydantic import create_pydantic_model
 
 class TestVarcharColumn(TestCase):
     def test_varchar_length(self):
-        class Director(Table):
+        class Manager(Table):
             name = Varchar(length=10)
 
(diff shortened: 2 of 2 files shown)


piccolo/table.py
```diff
@@ -315,7 +315,7 @@ def __init_subclass__(
                 if isinstance(column, Email):
                     email_columns.append(column)
 
-                if isinstance(column, Secret):
+                if isinstance(column, Secret) or column._meta.secret:
```
dantownsend, line 318: I don't think we need to check `isinstance(column, Secret)` - just `column._meta.secret` alone should work, because `Secret` columns should have `column._meta.secret = True`.

tests/table/test_metaclass.py
```diff
@@ -109,6 +109,19 @@ class Classified(Table):
...
         )
 
+    def test_secret_columns_with_argument(self):
+        """
+        Make sure TableMeta.secret_columns are setup correctly
+        with ``secret=True`` argument.
+        """
+
+        class Classified(Table):
+            top_secret = Varchar(secret=True)
```
dantownsend, line 119: As a sanity check, add a non-secret column here too, to make sure only secret columns are added to `_meta.secret_columns`. For example:

```python
class Classified(Table):
    top_secret = Varchar(secret=True)
    public = Varchar()
```

You might also consider merging this test with the above test, so we have this:

```python
def test_secret_columns(self):
    """
    Make sure ``TableMeta.secret_columns`` are setup correctly.
    """

    class Classified(Table):
        top_secret = Secret()
        confidential = Varchar(secret=True)
        public = Varchar()

    self.assertEqual(
        Classified._meta.secret_columns,
        [Classified.top_secret, Classified.confidential]
    )

Issue #1242, opened 2025-08-01
Update ``asgi`` templates


sinisaos:
I think we should update `ASGI` templates with these features:

1. Use `built-in` types for type hints (since we are using Python 3.9 and above)
2. We need to fix the `Falcon` template because `PiccoloCRUD` is not working. We need to use the Falcon `Resource Class` to get the correct web API.
3. Using the Piccolo [callback](https://piccolo-orm.readthedocs.io/en/latest/piccolo/query_clauses/callback.html) handler to check for `None` (for a record not found)
```python
# Check if the record is None. Use for query callback (in this case 
# for FastAPI, but this can be applied to other frameworks as well)
def check_record_not_found(result: dict[str, Any]) -> dict[str, Any]:
    if result is None:
        raise HTTPException(
            detail="Record not found",
            status_code=status.HTTP_404_NOT_FOUND,
        )
    return result

@app.put("/tasks/{task_id}/", response_model=TaskModelOut, tags=["Task"])
async def update_task(task_id: int, task_model: TaskModelIn):
    task = (
        await Task.objects()
        .get(Task._meta.primary_key == task_id)
        .callback(check_record_not_found)
    )
    for key, value in task_model.model_dump().items():
        setattr(task, key, value)

    await task.save()
    return task.to_dict()
```
Here is example off FastAPI `app.py`

<details>
<summary>FastAPI app.py</summary>

```python
from contextlib import asynccontextmanager
from typing import Any

from fastapi import FastAPI, status
from fastapi.exceptions import HTTPException
from piccolo.engine import engine_finder
from piccolo_admin.endpoints import create_admin
from piccolo_api.crud.serializers import create_pydantic_model
from starlette.routing import Mount, Route
from starlette.staticfiles import StaticFiles

from home.endpoints import HomeEndpoint
from home.piccolo_app import APP_CONFIG
from home.tables import Task


async def open_database_connection_pool():
    try:
        engine = engine_finder()
        await engine.start_connection_pool()
    except Exception:
        print("Unable to connect to the database")


async def close_database_connection_pool():
    try:
        engine = engine_finder()
        await engine.close_connection_pool()
    except Exception:
        print("Unable to connect to the database")


@asynccontextmanager
async def lifespan(app: FastAPI):
    await open_database_connection_pool()
    yield
    await close_database_connection_pool()


app = FastAPI(
    routes=[
        Route("/", HomeEndpoint),
        Mount(
            "/admin/",
            create_admin(
[...]

PR #1243, opened 2025-08-04
Update asgi templates

sinisaos:
Related to #1242

### DIFF
--- a/piccolo/apps/asgi/commands/templates/app/_blacksheep_app.py.jinja
+++ b/piccolo/apps/asgi/commands/templates/app/_blacksheep_app.py.jinja
@@ -37,7 +37,6 @@ TaskModelIn: Any = create_pydantic_model(
     table=Task,
     model_name="TaskModelIn",
 )
-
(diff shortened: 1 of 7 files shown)


sinisaos:
> * Whether we should be returning a 404 if a PUT tries to update a resource which doesn't exist (some people wouldn't treat this as an error). But this is the behaviour we already had in the previous templates, so lets leave it.

I think that's the right behavior and using a `callback` method is really handy in those cases.

> * We use `Task._meta.primary_key`. An alternative is to do `Task.id`, but then to add the following to the `Task` table definition:

Usually when I'm not using a custom primary key, I let Piccolo to [automatically create the primary key column](https://piccolo-orm.readthedocs.io/en/latest/piccolo/schema/defining.html#primary-key). In that case `Task._meta.primary_key` is a convenient way to satisfy a tool like mypy. I think we should leave this, but if you want, I can change it.

dantownsend:
Looks good, thanks!

Issue #1303, opened 2025-11-13
`add_m2m` now return None


diklios5768:
Picoolo `add_m2m` method use `M2MAddRelated` , and method `run` now show as follow:

```python
async def run(self):
    """
    Run the queries, making sure they are either within an existing
    transaction, or wrapped in a new transaction.
    """
    engine = self.rows[0]._meta.db
    if engine.transaction_exists():
        await self._run()
    else:
        async with engine.transaction():
            await self._run()
```

It should be `return await self._run()`, so it can run as stated in the docs that return with `id`

```shell
>>> band = await Band.objects().get(Band.name == "Pythonistas")
>>> await band.add_m2m(
...     Genre(name="Punk rock"),
...     m2m=Band.genres
... )
[{'id': 1}]
```
Or it only return `None`

diklios5768:
Just add `return ` for `insert` method as follow and works well:

```python
async def run(self):
    """
    Run the queries, making sure they are either within an existing
    transaction, or wrapped in a new transaction.
    """
    engine = self.rows[0]._meta.db
    if engine.transaction_exists():
        return await self._run()
    else:
        async with engine.transaction():
            return await self._run()
```

PR #1305, opened 2025-11-15
1303 `add_m2m` return ID of joining table row

dantownsend:
Resolves https://github.com/piccolo-orm/piccolo/issues/1303

When using `add_m2m`, the docs says that the query returns the ID of the row added to the joining table. This isn't the current behaviour of the code though.

https://piccolo-orm.readthedocs.io/en/latest/piccolo/schema/m2m.html#add-m2m

There's no harm in adding it - it could be useful to know.

### DIFF
--- a/piccolo/columns/m2m.py
+++ b/piccolo/columns/m2m.py
@@ -311,11 +311,8 @@ class M2MAddRelated:
         transaction, or wrapped in a new transaction.
         """
         engine = self.rows[0]._meta.db
-        if engine.transaction_exists():
-            await self._run()
-        else:
-            async with engine.transaction():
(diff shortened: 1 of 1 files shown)


piccolo/columns/m2m.py
```diff
@@ -311,11 +311,8 @@ async def run(self):
         transaction, or wrapped in a new transaction.
         """
         engine = self.rows[0]._meta.db
-        if engine.transaction_exists():
-            await self._run()
```
dantownsend: We don't need this check any more, because Piccolo no longer raises an exception if trying to create a transaction when one already exists (it's a no-op).

PR #1312, opened 2025-12-03
MySQL support

sinisaos:
Related to [this discussion](https://github.com/piccolo-orm/piccolo/discussions/1245). Any help and feedback is welcome.


.github/workflows/tests.yaml
```diff
@@ -194,3 +194,56 @@ jobs:
             - name: Upload coverage
               uses: codecov/codecov-action@v1
               if: matrix.python-version == '3.13'
+              
+    mysql-tests:
+        runs-on: ubuntu-latest
+        strategy:
+            matrix:
+                python-version: ["3.10", "3.11", "3.12", "3.13"]
```
dantownsend: Should this have 3.14 now?

docs/src/piccolo/migrations/create.rst
```diff
@@ -230,6 +230,11 @@ especially if the table is large, with many foreign keys.
 
         return manager
 
+.. warning:: For MySQL we need to run migrations outside transaction due to 
+    MySQL limitations. We can achive that by setting the ``MigrationManager`` 
+    argument ``wrap_in_transaction`` to ``False`` so that the migration 
+    is not wrapped in a transaction.
```
dantownsend: I think this is fine - but we might be able to automatically do this by detecting that MySQL is the engine type.

docs/src/piccolo/playground/advanced.rst
```diff
@@ -94,6 +94,42 @@ When you have the database setup, you can connect to it as follows:
...
+
+See :ref:`the docs on settings up MySQL <setting_up_mysql>`.
+
+Create database
+~~~~~~~~~~~~~~~
+
+By default the playground expects a local database to exist with the following
+credentials:
+
+
```
dantownsend: There's an extra line here.

docs/src/piccolo/playground/advanced.rst
```diff
@@ -94,6 +94,42 @@ When you have the database setup, you can connect to it as follows:
...
+.. code-block:: bash
```
dantownsend: Rather than a bash codeblock, this could be a list or table.

piccolo/query/mixins.py
```diff
@@ -664,24 +665,36 @@ def to_string(value) -> str:
...
             if action == OnConflictAction.do_nothing:
                 return QueryString(OnConflictAction.do_nothing.value)
             elif action == OnConflictAction.do_update:
                 values = []
-                query = f"{OnConflictAction.do_update.value} SET"
+                if engine is not None:
+                    if engine.engine_type == "mysql":
+                        query = ""
+                    else:
+                        query = f"{OnConflictAction.do_update.value} SET"
```
dantownsend: I get linter errors in this function, because `query` is potentially undefined, and lower down we append to `query`.

dantownsend (commented):
This is a lot of work, I appreciate the effort!

Here are my initial comments, thanks!

piccolo/columns/column_types.py
```diff
@@ -1027,6 +1074,14 @@ def __init__(
         self.default = default
         super().__init__(default=default, **kwargs)
 
+    @property
+    def column_type(self):
+        engine_type = self._meta.engine_type
+        if engine_type == "mysql":
+            return "TIMESTAMP(6)"
```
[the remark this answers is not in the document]
sinisaos: `TIMESTAMP(6)` is same as `TIMESTAMP`, but with `microsecond` precision which we use in Postgres

piccolo/columns/column_types.py
```diff
@@ -1274,6 +1329,8 @@ def column_type(self):
             # make it an integer - but we need a text field.
             # https://sqlite.org/datatype3.html#determination_of_column_affinity
             return "SECONDS"
+        elif engine_type == "mysql":
+            return "TIME(6)"
```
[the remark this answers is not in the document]
sinisaos: Same as above

piccolo/query/mixins.py
dantownsend: The reason I get a linter error is in this logic if we don't enter this block:

```python
if engine is not None
```
Then `query` is undefined.

You could just put in an `assert engine` and drop the `if`.

```python
assert engine
if engine.engine_type == "mysql":
    query = ""
else:
    query = f"{OnConflictAction.do_update.value} SET"
```

.github/workflows/tests.yaml
```diff
@@ -194,3 +194,56 @@ jobs:
...
+                    echo "Waiting for MySQL… ($i/40)"
+                    sleep 2
+                    if [ "$i" -eq 40 ]; then
+                        echo "MySQL did not become ready in time!" >&2
+                        exit 1
+                    fi
+                  done     
+            - name: Setup MySQL
+              run: |
+                  mysql -hmysql -uroot -prootpassword -e "CREATE DATABASE piccolo"
```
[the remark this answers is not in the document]
sinisaos: MySQL CI finally works

tests/testing/test_model_builder.py
```diff
@@ -85,7 +85,20 @@ def setUpClass(cls):
...
+            Band,
+            Manager,
+            Poster,
+            RecordingStudio,
+            Shirt,
+            Venue,
+            TableWithArrayField,
+            TableWithDecimal,
+            BandWithRecursiveReference,
+        )
```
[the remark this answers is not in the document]
sinisaos: Ahh. There was a subtle bug (single line of code) with sql generation for foreign keys that caused a problem with droping referenced tables. That was the reason I had to rearrange the tables to pass the unittest. I revert all the tests where I changed that order.

dantownsend:
I was thinking with auto migrations - even though they can't run in a transaction, we could still roll them back for the user by undoing the changes (i.e. if the migration added a column, then we delete the column if the migration fails before the end).

Not essential for a v1, just a thought.

sinisaos:
MySQL has limitations on ddl in transactions, but I was wrong. All the errors I was getting were due to incorrectly generated ddl for foreign keys (that's fixed now). In fact, what I'm seeing is that tables are being dropped if the migration fails, which is great. But there are a number of other issues that directly impact auto migrations and it would be great if you could help with them. 
1. `Non-primary` foreign keys don't work (easiest way would be to first create the table and then add the constraints in a separate alter statement (two queries), but I don't know how to do that in Piccolo)
2. `null/not null` doesn't work because MySQL has a different syntax that I don't know how to handle in alter statements
3. Indexes in MySQL are dropping differently than in Postgres, and the syntax is different.
4. Default values for text, blob etc. Error `1101, "BLOB, TEXT, GEOMETRY or JSON column 'my_column' can't have a default value"`

docs/src/piccolo/getting_started/database_support.rst
```diff
@@ -17,6 +17,17 @@ together in production. The main missing feature is support for
 :ref:`automatic database migrations <AutoMigrations>` due to SQLite's limited
 support for ``ALTER TABLE`` ``DDL`` statements.
 
+`MySQL <https://www.mysql.com/>`_ has limited support due to some MySQL limitations.
+Except that MySQL doesn't have specific column types (like Postgres), the main missing feature 
+is support for :ref:`automatic database migrations <AutoMigrations>`. MySQL ``DDL``
```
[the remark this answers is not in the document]
sinisaos: I changed this because automatic migrations work, but we need to be careful.

docs/src/piccolo/getting_started/setup_mysql.rst
```diff
@@ -0,0 +1,28 @@
...
+Creating a database
+*******************
+
+Using ``mysql``:
+
+.. code-block:: bash
+
+    mysql -u root -p
+
+Enter the your password and create database:
```
dantownsend: Can we change this to:

`Enter your password and create the database:`

docs/src/piccolo/getting_started/setup_mysql.rst
```diff
@@ -0,0 +1,28 @@
...
+
+.. code-block:: bash
+
+    CREATE DATABASE "my_database_name";
+
+For Windows you can use some GUI tool.
```
dantownsend: This is also true of MacOS etc, so maybe change it to:

`Alternatively, use a GUI tool.`

dantownsend (commented):
It's going to take me a while to go through it all, so will do it a bit at a time.

Looking good so far - just left some comments about tweaks to the docs 👍

piccolo/columns/column_types.py
```diff
@@ -442,9 +454,23 @@ def __init__(
...
+        """
+        MySQL does not allow unquoted TEXT literals in the DEFAULT
+        clause, so we use the expression in parentheses.
+        Only works in CREATE TABLE. MySQL does not allow default
+        values for TEXT columns in ALTER statements.
+        """
+        engine_type = self._meta.engine_type
+
+        if engine_type == "mysql":
+            return QueryString(f"('{self.default}')")
```
dantownsend: I think this could be problematic.

`self.default` could be all sorts of different things - like `None` or a function, so if we embed it in a string we might end up with invalid DDL.

What if this logic lives in `get_sql_value` instead?

piccolo/columns/column_types.py
```diff
@@ -502,6 +528,15 @@ class Band(Table):
 
     value_type = uuid.UUID
 
+    @property
+    def column_type(self):
+        engine_type = self._meta.engine_type
+        if engine_type in ("postgres", "cockroach", "sqlite"):
+            return "UUID"
+        elif engine_type == "mysql":
+            return "CHAR(36)"
```
dantownsend: Wow, crazy that MySQL doesn't have a native UUID column type.

tests/engine/test_version_parsing.py
```diff
@@ -27,3 +28,27 @@ def test_version_parsing(self):
             ),
             12.4,
         )
+
+
+@engines_only("mysql")
+class TestVersionParsingMysql(TestCase):
```
dantownsend: MySQL

docs/src/piccolo/getting_started/database_support.rst
```diff
@@ -17,6 +17,13 @@ together in production. The main missing feature is support for
 :ref:`automatic database migrations <AutoMigrations>` due to SQLite's limited
 support for ``ALTER TABLE`` ``DDL`` statements.
 
+`MySQL <https://www.mysql.com/>`_ is supported with some MySQL limitations.
+MySQL doesn't have specific column types (like Postgres) and there may be some
```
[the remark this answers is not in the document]
sinisaos: `TimeStamptz` and `JSONB` columns are not supported. Should we add that or remove `MySQL doesn't have specific column types` completely? Array column is supported thru `JSON`.

docs/src/piccolo/getting_started/setup_mysql.rst
sinisaos: Thanks. I'm not a native English speaker and these mistakes always creep in. I would change that.

docs/src/piccolo/getting_started/setup_mysql.rst
sinisaos: Thanks.  I would change that.

docs/src/piccolo/migrations/create.rst
```diff
@@ -256,9 +256,10 @@ Creating an auto migration:
     aren't supported by auto migrations, or to modify the data held in tables, as
     opposed to changing the tables themselves.
 
-.. warning:: Auto migrations aren't supported in SQLite, because of SQLite's
-    extremely limited support for SQL Alter statements. This might change in
-    the future.
+.. warning:: Auto migrations for SQLite and MySQL are supported, with limitations.
+    SQLite has extremely limited support for SQL Alter statements and MySQL DDL triggers
```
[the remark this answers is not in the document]
sinisaos: Thanks. I would change that.

docs/src/piccolo/query_clauses/on_conflict.rst
```diff
@@ -192,11 +194,15 @@ update should be made:
     ...     where=Band.popularity < 1000
     ... )
 
+.. warning:: Not supported for MySQL. A workaround is possible by using an  
+    ``IF`` or ``CASE`` condition in the ``UPDATE`` clause or by first
+    performing a separate ``UPDATE``, but this is not covered in Piccolo.
```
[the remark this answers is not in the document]
sinisaos: Thanks. I would change that.

tests/engine/test_pool.py
```diff
@@ -69,6 +70,59 @@ def test_many_queries(self):
         asyncio.run(self._make_many_queries())
 
 
+@engines_only("mysql")
+class TestPoolMysql(DBTestCase):
```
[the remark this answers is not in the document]
sinisaos: Ok. I will change all `Mysql` -> `MySQL`

tests/query/functions/test_functions.py
```diff
@@ -31,8 +31,11 @@ def test_nested_within_querystring(self):
         If we wrap a function in a custom QueryString - make sure the columns
         are still accessible, so joins are successful.
         """
+        # Use Concat() for compatibility with all databases
```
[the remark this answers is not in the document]
sinisaos: I will remove this. Thanks

piccolo/columns/column_types.py
sinisaos: But we need to somehow change the default value because `TEXT`, `BLOB`, and `JSON` do not allow unquoted literals in default . This is not allowed.
```sql
CREATE TABLE example (
    description TEXT DEFAULT 'hello'
);
```
We need to use an expression in parentheses.
```sql
CREATE TABLE example (
    description TEXT DEFAULT ('hello') -- expression
);
```
Only works in `CREATE TABLE` because MySQL does not allow default values ​​(at all, with or without expressions) for `TEXT`, `BLOB`, and `JSON`  columns in `ALTER` statements

piccolo/query/methods/exists.py
```diff
@@ -29,11 +29,18 @@ async def response_handler(self, response) -> bool:
...
+            return [
+                QueryString(
+                    "SELECT EXISTS({}) AS `exists`", select.querystrings[0]
+                )
+            ]
+        else:
+            return [
+                QueryString(
+                    'SELECT EXISTS({}) AS "exists"', select.querystrings[0]
+                )
```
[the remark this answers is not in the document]
sinisaos: We can actually remove this repetition. It seems I wrote that code before I wrote the utility function in the engine. I'll remove that. Thanks.

requirements/requirements.txt
```diff
@@ -5,3 +5,4 @@ targ>=0.3.7
 inflection>=0.5.1
 typing-extensions>=4.3.0
 pydantic[email]==2.*
+PyMySQL==1.1.2
```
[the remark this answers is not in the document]
dantownsend: OK, interesting. I wonder if it's possible to lazy import `pymysql` to avoid this problem. I need to try it.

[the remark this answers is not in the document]
sinisaos: This is done.

.github/workflows/tests.yaml
```diff
@@ -194,3 +194,77 @@ jobs:
...
+                pip install -r requirements/extras/mysql.txt
+
+            - name: Install MySQL Client
+              run: sudo apt-get update && sudo apt-get install -y mysql-client
+
+            - name: Wait for MySQL
+              run: |
+                set -e
+                for i in {1..60}; do
+                    if mysqladmin ping -h127.0.0.1 -P3306 -uroot -prootpassword > /dev/null 2>&1; then
```
dantownsend: Can there be spaces here - so `-h 127.0.0.1` instead of `-h127.0.0.1`?

docs/src/piccolo/getting_started/setup_mysql.rst
```diff
@@ -0,0 +1,28 @@
...
+Creating a database
+*******************
+
+Using ``mysql``:
+
+.. code-block:: bash
+
+    mysql -u root -p
+
+Enter the your password and create the database:
```
dantownsend: Can we change this to `Enter your password` .

tests/base.py
```diff
@@ -230,6 +239,57 @@ def get_postgres_varchar_length(
...
+    def get_mysql_is_nullable(self, tablename, column_name: str) -> bool:
+        """
+        Fetches whether the column is defined as nullable, from the database.
+        """
+        return (
+            self.get_mysql_column_definition(
+                tablename=tablename, column_name=column_name
+            ).is_nullable.upper()
+            == "YES"
+        )
```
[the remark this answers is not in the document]
sinisaos: But column definitions of MySQL and Postgres are not the same. We can do something similar using a single function like this

```python
def get_is_nullable(self, tablename, column_name: str) -> bool:
    if engine_is("mysql"):
        return (
            self.get_mysql_column_definition(
                tablename=tablename, column_name=column_name
            ).is_nullable.upper()
            == "YES"
        )
    return self.get_postgres_column_definition(
        tablename=tablename, column_name=column_name
    ).character_maximum_length
```
but I think the current way is much clearer with the distinction between Postgres and MySQL.

.github/workflows/tests.yaml
sinisaos: That doesn't work. I'll revert to the previous commands.

piccolo/apps/schema/commands/generate.py
```diff
@@ -67,11 +67,22 @@ class RowMeta:
...
+        engine = engine_finder()
+        assert engine
+        if engine.engine_type == "mysql":
+            return ", ".join(
+                i.name
+                for i in dataclasses.fields(cls)
+                if i.name != "numeric_precision_radix"
+            )
+        else:
+            return ", ".join(i.name for i in dataclasses.fields(cls))
```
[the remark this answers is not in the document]
sinisaos: This is great. Thanks.

tests/mysql_conf.py
```diff
@@ -0,0 +1,22 @@
+import os
+
+from piccolo.conf.apps import AppRegistry
+from piccolo.engine.mysql import MySQLEngine
+
+DB = MySQLEngine(
+    config={
+        "host": os.environ.get("MY_HOST", "127.0.0.1"),
+        "port": int(os.environ.get("MY_PORT", 3306)),
+        "user": os.environ.get("MY_USER", "root"),
```
dantownsend: What do you think about making the user `piccolo` like the other engines?

dantownsend:
I've got the tests running locally now. It's impressive that they're running.

I've tried the playground too, and it seems to work well. I do get things like this printed out in the console:

```
aiomysql/cursors.py:239: Warning: Unknown table 'piccolo_playground.band'
```

Do you see them too?

piccolo/columns/defaults/uuid.py
```diff
@@ -27,6 +27,10 @@ def cockroach(self):
     def sqlite(self):
         return "''"
 
+    @property
+    def mysql(self):
+        return f"'{uuid.uuid4()}'"
```
dantownsend: I see where you're coming from. but ideally we need a different solution here. This will give each row the same value by default.

MySQL has a `uuid()` function, but it's uuid v1.

```
SELECT uuid();
```

https://dev.mysql.com/doc/refman/8.4/en/miscellaneous-functions.html#function_uuid

We could define our own uuid4 function in the database, or just default to an empty string.

sinisaos:
I'm closing this PR just to clean up my open requests in the repo. Since it has been open for quite a while, I'll remove it from the list to keep things tidy. Feel free to use it if needed. Thanks.

Issue #1254, opened 2025-08-24
applying `auto migration` does not rolled back the migration


Jacky56:
Hello all,

Assume you have a auto migration which adds new columns:
```text
1. manager.add_column(...)
2. manager.add_column(...)
3. manager.add_column(Some foreign key that doesn't exist as of yet will make the manager throw an error)
4. manager.add_column(...)
```

Now, what the `piccolo.apps.migrations.auto.migration_manager.MigrationManager` what will happen on the database:
- add column 1
- add column 2

now the manager will evaluate 3. but crash into an error, and will not rollback the 2 added columns on the database

dantownsend:
Ah ok, I need to test it. Are you running Postgres? Each migration should run inside a transaction, and if something fails then the entire migration gets rolled back.

sinisaos:
The migration is executed in a transaction

https://github.com/piccolo-orm/piccolo/blob/5139ec0644675bc3b195d46de2d50571cfa9e4c7/piccolo/apps/migrations/auto/migration_manager.py#L982-L986

and if something goes wrong, the migration is rolled back. At least in my case everything works correctly. As in the example from the previous comment, if `column 3` has an error, the migration transaction is rolled back and `column 1` and `column 2` are not added to the database. I think this is the correct behavior, at least in my case.

Jacky56:
hello @dantownsend @sinisaos 

i forgotten to add the versions in the description:
```
piccolo version: piccolo[postgres]==1.24.2
cockroachDB version: Basic, Google Cloud v25.2.4
```


so the [code mentioned](https://github.com/piccolo-orm/piccolo/blob/5139ec0644675bc3b195d46de2d50571cfa9e4c7/piccolo/apps/migrations/auto/migration_manager.py#L982-L986) by @sinisaos looks correct, so I can't give any assumptions from my end.

Here roughly the migration which did not rollback:
```python
from piccolo.apps.migrations.auto.migration_manager import MigrationManager
from piccolo.columns.base import OnDelete
from piccolo.columns.base import OnUpdate
from piccolo.columns.column_types import ForeignKey
from piccolo.columns.column_types import Text
from piccolo.columns.column_types import Varchar
from piccolo.columns.indexes import IndexMethod
from piccolo.table import Table


# note that due to how migrations with multiple apps are ran https://github.com/piccolo-orm/piccolo/issues/1253
# This table was not created yet
class Geolocation(Table, tablename="geolocation", schema=None):
[...]

PR #1199, opened 2025-05-17
allow changes to the array base column type in migrations

sinisaos:
Related to #1076

PR #1318, opened 2025-12-30
Cockroach upgrade

sinisaos:
Piccolo CI uses an unmaintained version. This PR upgrades Cockroach to the latest regular release [v25.4.2](https://www.cockroachlabs.com/docs/releases/v25.4#v25-4-2). Also related to #1254


piccolo/engine/cockroach.py
```diff
@@ -35,15 +37,14 @@ def __init__(
...
+        """
+        Set `autocommit_before_ddl` to off (enabled by default since v25.2)
+        to prevent automatic DDL commits in transactions and enable rollback
+        """
+        connection = await super().get_new_connection()
+        await connection.execute(
+            "SET autocommit_before_ddl = off;"
+            "ALTER ROLE ALL SET autocommit_before_ddl = false;"
+        )
+        return connection
```
dantownsend: Is there some way of setting `autocommit_before_ddl = off` for all sessions, so we don't have to do it for each connection?

sinisaos: If this PR is merged, [this PR can also](https://github.com/piccolo-orm/piccolo/pull/1199) be merged because that PR works without issues with this PR changes (works with the latest Cockroach regular release `v25.4.2`)

sinisaos: You're actually right. There is also a third option with `SET LOCAL` which limits the setting to the current transaction only. With that, we can make a setting `after the transaction starts`, **not before** (sorry, I didn't know that). Once the transaction is commited or rolled back, the setting is automatically reset. Can you check the [last commit](https://github.com/piccolo-orm/piccolo/pull/1318/commits/6f33b8a7e4f2204b198efa519e203b36fd7de8a1)? Did you mean something like that?

dantownsend: Yes, I think something like that, but I'll double check tomorrow.

sinisaos: @dantownsend I made [another commit](https://github.com/piccolo-orm/piccolo/pull/1318/commits/3e9300d4b9589bc9e3e53bfe4800561b383780a1) where I made small changes (the default is always `False` for both `transaction` and `atomic` so we don't have to change the tests) that you can easily try out. I hope that's better.

Issue #1344, opened 2026-02-08
Piccolo CLI registers user app automatically


pelid:
The documentation promises that the user app will not connect automatically: [docs](https://piccolo-orm.readthedocs.io/en/latest/piccolo/projects_and_apps/included_apps.html#optional-includes). But Piccolo CLI ignores this and registers the application anyway.

```py
# piccolo_conf.py
APP_REGISTRY = AppRegistry(apps=[])
```

Console output:

```
$ piccolo
...
user create
Create a new user. 

user list
List existing users. 

user change_password
Change a user's password. 

user change_permissions
Change a user's permissions. 
...
```

The problem is that my project uses another similar application instead of the standard user application. When Piccolo CLI is launched, confusion begins in the names of commands and applications.

sinisaos:
Maybe we should remove the `user` app from the included apps because, as already stated in the authentication documentation, [we need to register the user app to use it](https://piccolo-orm.readthedocs.io/en/latest/piccolo/authentication/index.html#registering-the-app)?

PR #1346, opened 2026-02-10
remove user app from auto included apps

sinisaos:
Related to #1344


sinisaos:
@dantownsend Have you considered merging this? This small change will allow the user to use a custom user app without interfering with the Piccolo user app. The documentation already states that the [user app is optional](https://piccolo-orm.readthedocs.io/en/latest/piccolo/projects_and_apps/included_apps.html#optional-includes) and if we want use a user app, [we need to register it in the AppRegistry](https://piccolo-orm.readthedocs.io/en/latest/piccolo/authentication/index.html#registering-the-app). I have also updated the `ASGI` template so that the user app is included by default. Thanks.

PR #1395, opened 2026-06-23
Add boolean input param to enable WAL mode

TimothyAllman:
Hi there

adding boolean input to enable WAL mode
as mentioned in this discussion https://github.com/piccolo-orm/piccolo/discussions/1394#discussioncomment-17218171

let me know if I mis-interpretted anything or if anything needs more work

I was thinking as a first pass (this PR) just add the boolean input param.
then as a second pass, maybe generalize for any PRAGMA statment (wondered if you had any ideas or first thoughts as to how you would want to go about this)

love the work on piccolo
Many thanks

### DIFF
--- a/piccolo/engine/sqlite.py
+++ b/piccolo/engine/sqlite.py
@@ -590,6 +590,7 @@ class SQLiteEngine(Engine[SQLiteTransaction]):
         path: str = "piccolo.sqlite",
         log_queries: bool = False,
         log_responses: bool = False,
+        enable_wal_mode: bool = False,
         **connection_kwargs,
     ) -> None:
         """
@@ -618,6 +619,7 @@ class SQLiteEngine(Engine[SQLiteTransaction]):
 
         self.log_queries = log_queries
         self.log_responses = log_responses
+        self.enable_wal_mode = enable_wal_mode
         self.connection_kwargs = {
             **default_connection_kwargs,
             **connection_kwargs,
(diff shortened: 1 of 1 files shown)


piccolo/engine/sqlite.py
```diff
@@ -590,6 +590,7 @@ def __init__(
         path: str = "piccolo.sqlite",
         log_queries: bool = False,
         log_responses: bool = False,
+        enable_wal_mode: bool = False,
```
dantownsend: What do you think of changing this to:

```python
journal_mode: Literal['DELETE', 'TRUNCATE', 'PERSIST', 'MEMORY', 'WAL', 'OFF'] | None = None
```

It provides a bit more flexibility if people want alternative journal modes.

dantownsend (commented):
This looks good - thanks!

Just one suggestion about making it more flexible.

piccolo/engine/sqlite.py
TimothyAllman: that makes a lot more sense.
have changed it accordingly, updated the PR and pushed the changes.

I did notice some discussion around this here https://github.com/piccolo-orm/piccolo/discussions/1394#discussioncomment-17452479.
Even so, I think the above change might be a good first pass to get into piccolo and useful for use-cases that I and others might want to try out with piccolo (like litestream https://litestream.io/)

thanks so much for all the hard work Dan and the Piccolo Team.
Let me know if there is anything else.

PR #1400, opened 2026-07-08
Warn when --fake marks multiple migrations as run

### LINKED ISSUE #1255: applying `migration forward <module> --fake` should not fake all migrations
hello all,

so there was some slight issue with the migrations process, where its incredibly buggy due to no [rollbacks](https://github.com/piccolo-orm/piccolo/issues/1254) when migrations breaks midway through

the `--fake` parameter for `migrations forward` should ideally only fake the next migrations or at the very least warn the developer before applying

AnayGarodia:
Closes #1255

`piccolo migrations forwards --fake` marks pending migrations as run without applying them. When several migrations are pending, running it with `--fake` silently fakes all of them, which is an easy way to lose track of which schema changes were actually applied.

As agreed in #1255, this adds a warning when `--fake` will mark more than one migration as run, printed right after the "Running N migrations" line and before they are faked. A single faked migration is unchanged.

The regression test (`test_forwards_fake_multiple_warns`) fakes the `music` app's migrations with `fake=True` and asserts the warning is printed; it fails on `master` and passes with this change. The rest of the forwards/backwards suite still passes.

### DIFF
--- a/piccolo/apps/migrations/commands/forwards.py
+++ b/piccolo/apps/migrations/commands/forwards.py
@@ -70,6 +70,12 @@ class ForwardsMigrationManager(BaseMigrationManager):
             n = len(subset)
             print(f"🚀 Running {n} migration{'s' if n != 1 else ''}:")
 
+            if self.fake and n > 1:
+                print(
+                    f"⚠️  --fake will mark all {n} migrations as run without "
+                    "applying them."
+                )
+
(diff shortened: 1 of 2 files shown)


dantownsend:
Thanks for this. I made a tweak - so we use `input` to ask the user if they want to proceed.

PR #1401, opened 2026-07-15
fix: don't treat an already-run migration as unrecognised

pctablet505:
Fixes #1359

When a specific migration ID passed to `piccolo migrations forwards` has already been run, the command now exits successfully and reports that no migrations need to be run, instead of printing `<id> is unrecognised` and returning an error status.


dantownsend:
Thanks for this. I think what you did was great, but I tweaked it a tiny bit just for personal preference.

PR #1408, opened 2026-07-20
1399 `load_json` and array values with `prefetch`

### LINKED ISSUE #1399: `load_json` and array values with `prefetch`
Hi Dan, would like the fix you merged here if possible https://github.com/piccolo-orm/piccolo/pull/1392

Would it be possible to release soon? If not, roughly when do you think you could?

Thank you

dantownsend:
Resolves https://github.com/piccolo-orm/piccolo/issues/1399

### DIFF
--- a/piccolo/query/base.py
+++ b/piccolo/query/base.py
@@ -95,7 +95,7 @@ class Query(Generic[TableInstance, QueryResponseType]):
                 new_row = {**row}
                 for json_column_name in json_column_names:
                     value = new_row.get(json_column_name)
-                    if isinstance(value, (str, bytes, bytearray)):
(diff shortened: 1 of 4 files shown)


piccolo/query/base.py
```diff
@@ -105,28 +105,27 @@ async def _process_results(self, results) -> QueryResponseType:
...
+                    ],
+                )
+            else:
+                return cast(
+                    QueryResponseType,
+                    [
+                        self.table(**columns, _exists_in_db=True)
+                        for columns in raw
+                    ],
+                )
```
dantownsend, line 128: Just removed one level of nesting here - no other change.

PR #1409, opened 2026-07-21
Add GroupByRaw query expression

### LINKED ISSUE #1387: Support `GroupByRaw` similar  to `OrderByRaw`, `WhereRaw` , `SelectRaw`
Sometimes we use `SelectRaw` create new column and we want to use like this:

```python
rows = Table.select(
    Count(),
    SelectRaw('DATE({} AT TIME ZONE {}) AS stat_date', Table.created_at, TZ_NAME),
).where(where).group_by(
    GroubyRaw('stat_date')
)
```

Now if use `.group_by('stat_date')` which not a column in `Table` will raise `ValueError(f"No matching column found with name == {name}")`

Now I'm working in this way:

```python
from dataclasses import dataclass

@dataclass
class GroupByRaw:
    sql: str

    @property
    def querystring(self) -> QueryString:
        return QueryString(f' GROUP BY {self.sql}')


def group_by_raw(query: Any, sql: str) -> Any:
    query.group_by_delegate._group_by = GroupByRaw(sql)
    return query

query = Table.select(
    Count(),
    SelectRaw('DATE({} AT TIME ZONE {}) AS stat_date', Table.created_at, TZ_NAME),
).where(where)

rows = await group_by_raw(query, 'stat_date').run()
```
And perhaps there should be a `CountRaw` class? 😄

AnayGarodia:
Fixes #1387.

Adds a public `GroupByRaw` expression following the existing `OrderByRaw` pattern. `Select.group_by` now accepts columns, column-name strings, and trusted raw grouping expressions, and mixed raw/column clauses render in order.

The documentation includes a `SelectRaw` alias example and an explicit trusted-SQL warning.

Verification:
- `PICCOLO_CONF=tests.sqlite_conf python -m pytest tests/table/test_select.py -q` — 74 passed, 7 skipped
- `python -m compileall -q piccolo/query`
- Black/isort applied to changed code
- `git diff --check`

I also ran mypy over the changed modules; it reaches an existing unrelated return-type error at `piccolo/table.py:194` (`Column | None` vs `Column`).

### DIFF
--- a/docs/src/piccolo/query_clauses/group_by.rst
+++ b/docs/src/piccolo/query_clauses/group_by.rst
@@ -39,3 +39,25 @@ Other aggregate functions
 -------------------------
 
 These work the same as ``Count``. See :ref:`aggregate functions <AggregateFunctions>`.
+
+-------------------------------------------------------------------------------
+
+``GroupByRaw``
+--------------
+
+For more complex cases, use ``GroupByRaw`` to group by a raw SQL expression or
+an alias created by ``SelectRaw``. For example:
+
+.. code-block:: python
+
+    from piccolo.query import GroupByRaw, SelectRaw
+    from piccolo.query.functions.aggregate import Count
+
+    await Band.select(
+        SelectRaw("DATE(created_at) AS created_date"),
+        Count(),
+    ).group_by(
+        GroupByRaw("created_date")
+    )
+
+As with other raw query helpers, only use trusted SQL strings.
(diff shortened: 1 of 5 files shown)


dantownsend:
Thanks for this - it looks good to me, but I'll give it a more thorough review.

dantownsend:
This looks great to me - thanks!

The only thing I might do in a follow up PR is convert all of the Raw classes (`OrderByRaw`, `WhereRaw` etc) to just be subclasses of `QueryString`.

PR #1422, opened 2026-08-27
Add `ArrayOverlap` function (`&&`)

regiscamimura:
`piccolo.query.functions.array` covers `array_cat`, `array_append`,
`array_prepend`, `array_replace` and `array_remove`, but not the `&&` overlap
operator.

The gap it fills: "does this array contain any of these values?". Today that has
to be written as an `OR` of `any()` calls, one per value, which can't use a GIN
index. `&&` is a single indexable test.

```python
from piccolo.query.functions import ArrayOverlap

await Ticket.select().where(
    ArrayOverlap(Ticket.seat_numbers, [510, 511])
)
```

There's also a column method, matching `cat` / `remove` / `append` / `prepend` /
`replace`:

```python
await Ticket.select().where(Ticket.seat_numbers.overlap([510, 511]))
```

### Notes

* Postgres / CockroachDB only, raising the same `ValueError` on SQLite as the
  other array functions.
* `ArrayOverlap` subclasses `QueryString` rather than `ArrayQueryString` —
  `ArrayQueryString` overrides `__add__` / `__radd__` to mean concatenation,
  which doesn't apply here as `&&` returns a boolean, not an array.
* On CockroachDB the array needs an explicit type, the same as `any()` /
  `not_any()` — the tests use `QueryString("{}::INTEGER[]", [...])` for that
  reason. On Postgres a plain list works, since the type is inferred from the
  column.

### Testing

Added `test_overlap`, `test_overlap_plain_list`, `test_overlap_two_columns` and
`test_overlap_sqlite` to `tests/columns/test_array.py`:

```
$ ./scripts/test-postgres.sh tests/columns/test_array.py
18 passed, 6 skipped

$ ./scripts/test-sqlite.sh tests/columns/test_array.py
12 passed, 12 skipped
```

I don't have a CockroachDB instance to hand, so that path is untested beyond
following the existing convention.

### DIFF
--- a/docs/src/piccolo/functions/array.rst
+++ b/docs/src/piccolo/functions/array.rst
@@ -8,6 +8,11 @@ ArrayCat
 
 .. autoclass:: ArrayCat
 
+ArrayOverlap
+------------
+
+.. autoclass:: ArrayOverlap
+
 ArrayAppend
 -----------
 
--- a/docs/src/piccolo/schema/column_types.rst
+++ b/docs/src/piccolo/schema/column_types.rst
@@ -444,6 +444,12 @@ all
 
 .. automethod:: Array.all
 
+=======
+overlap
+=======
+
+.. automethod:: Array.overlap
+
 ===
 cat
 ===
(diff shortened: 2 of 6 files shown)


dantownsend:
Nice idea, thanks

dantownsend:
Looks great, thanks 👍

PR #1424, opened 2026-08-27
Add Postgres full text search functions

regiscamimura:
**Postgres only.** SQLite's FTS is a separate, unrelated extension, and
CockroachDB's `tsvector` support is incomplete — so passing a column belonging
to either engine into any of these raises
`ValueError("Only Postgres supports full text search.")`, following the same
pattern as the array functions. A SQLite user gets that error at query-build
time, not a confusing database error.

There's currently no way to express Postgres full text search through Piccolo,
so it has to be written as raw SQL. This adds the pieces as composable
`QueryString`s.

```python
from piccolo.query.functions import Matches, ToTsVector, WebsearchToTsQuery

await Band.select(Band.name).where(
    Matches(
        ToTsVector(Band.name, config='english'),
        WebsearchToTsQuery('pythonistas', config='english'),
    )
)
```

| Class | SQL |
| --- | --- |
| `ToTsVector` | `to_tsvector` |
| `ToTsQuery` | `to_tsquery` |
| `PlainToTsQuery` | `plainto_tsquery` |
| `WebsearchToTsQuery` | `websearch_to_tsquery` |
| `Matches` | `@@` |

### The text search config can't be a query parameter

Postgres only resolves `to_tsvector(config, text)` when the config is a
constant, and an expression index over it is only `IMMUTABLE` because of that.
So `config` is a `str` which is written into the query rather than bound.

Since it isn't parameterised, it's validated against `^[A-Za-z_][A-Za-z0-9_]*$`
before it goes in, so it can't be used for injection. `config=None` (the
default) emits the single-argument form, which uses the database's
`default_text_search_config` — the docstring notes that this form is `STABLE`
rather than `IMMUTABLE`, so it can't be used in an expression index.

### Sanitising user input

`to_tsquery` raises a database error on a stray operator, so it's the wrong
thing to hand a raw search box to. `PlainToTsQuery` and `WebsearchToTsQuery`
accept any text, so they're the safe entry points — that's called out in
`ToTsQuery`'s docstring with a `.. warning::` pointing at them.

**A question for you:** the one thing `to_tsquery` can do that the safe
variants can't is prefix matching (`python:*`), which is what you want for
type-ahead search. Doing that safely means a helper which strips the input to
word characters, `AND`s them, and appends `:*` to the last word. I've left it
out — it's a policy decision rather than a wrapper around a Postgres function,
and it's the kind of thing that belongs in the docs as a recipe if you don't
want it in the library. Happy to add it if you do.

### Deliberately left out

* `ts_rank` / `ts_rank_cd` / `ts_headline` — the obvious follow-ups, but they
  have fussier signatures (normalization flags, option strings) and this PR is
  already the whole matching path.
* `tsvector` / `tsquery` **column types**, and GIN index support for them.
  Right now the vector has to be computed per query, or the user has to add the
  column and index through a raw migration. That's the bigger piece of work and
  I didn't want to bundle it in.
* Relaxing the engine check for CockroachDB — it has had partial `tsvector`
  support since v23.2, but I have no instance to verify against, so I've been
  conservative. It's a one-line change to `check_engine` if you'd rather allow
  it.

### Naming

`Matches` is the least obvious name here — the operator has no function name in
Postgres. Alternatives would be `TsMatch` or `TextSearchMatch`. Happy to rename.

### Testing

New file `tests/query/functions/test_text_search.py` — 7 Postgres-only
behaviour tests plus config validation and a check that the wrong engine
raises.

```
$ ./scripts/test-postgres.sh tests/query/functions/
43 passed, 2 skipped

$ ./scripts/test-sqlite.sh tests/query/functions/
31 passed, 14 skipped
```

Full suite against Postgres — `1 failed, 891 passed, 34 skipped`. The one
failure is `tests/table/test_update.py::TestOperators::test_operators`, which
fails on `master` for me too.

### DIFF
--- a/docs/src/piccolo/functions/index.rst
+++ b/docs/src/piccolo/functions/index.rst
@@ -17,5 +17,6 @@ Functions can be used to modify how queries are run, and what is returned.
     ./datetime
     ./math
     ./string
+    ./text_search
     ./type_conversion
     ./custom
--- a/docs/src/piccolo/functions/text_search.rst
+++ b/docs/src/piccolo/functions/text_search.rst
@@ -1,0 +1,52 @@
+Text search functions
+=====================
+
+.. currentmodule:: piccolo.query.functions.text_search
+
+.. note:: Postgres only. SQLite has its own, unrelated FTS extension, and
+    CockroachDB's support is incomplete. Passing a column belonging to any
+    other engine into these functions raises a ``ValueError``.
+
+Postgres full text search works in two steps - the text being searched is
+converted into a ``tsvector``, the search terms are converted into a
+``tsquery``, and the two are compared with the ``@@`` operator.
+
+.. code-block:: python
+
+    from piccolo.query.functions import (
+        Matches,
+        ToTsVector,
+        WebsearchToTsQuery,
+    )
+
+    await Band.select(Band.name).where(
+        Matches(
+            ToTsVector(Band.name, config='english'),
+            WebsearchToTsQuery('pythonistas', config='english'),
+        )
+    )
+
+ToTsVector
+----------
+
+.. autoclass:: ToTsVector
+
+ToTsQuery
+---------
+
+.. autoclass:: ToTsQuery
+
+PlainToTsQuery
+--------------
+
+.. autoclass:: PlainToTsQuery
+
+WebsearchToTsQuery
+------------------
+
+.. autoclass:: WebsearchToTsQuery
+
+Matches
+-------
+
+.. autoclass:: Matches
--- a/piccolo/query/functions/__init__.py
+++ b/piccolo/query/functions/__init__.py
@@ -28,6 +28,13 @@ from .string import (
     Rtrim,
     Upper,
 )
+from .text_search import (
+    Matches,
+    PlainToTsQuery,
+    ToTsQuery,
+    ToTsVector,
+    WebsearchToTsQuery,
+)
 from .type_conversion import Cast
 
 __all__ = (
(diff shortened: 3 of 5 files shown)


lachiewalker:
@regiscamimura have you had a look at #1390 ?

regiscamimura:
Hi @lachiewalker — Just did :grimacing: And yeah, if I have looked at that I wouldn't have created this one. 

A proposal, if you and @dantownsend are open to it: treat #1424 as the small first slice and rebase #1390 on top of it.

  - #1424 is narrow — the five text search functions plus Matches, no column types, no engine changes — and currently merges clean. #1390 is +1394/−30 across three features and is showing conflicts with master, so it'll need a rebase either way.
- Once #1424 is in, your Tsvector/Tsquery column types, TsRank/TsRankCd/TsHeadline, GIN operator_class, pgvector and pg_trgm all land on top with less to review in one go — the function names and file already exist, so that part of your diff mostly disappears.

Two things I'd flag from having gone down the same road, since they'd carry over:

1. config as a bound parameter. In #1390 ToTsvector(document, config=...) passes config through as a $1 placeholder. Postgres only treats to_tsvector(config, text) as IMMUTABLE when the config is a constant, so a bound value can't be used in an expression index (CREATE INDEX ... ON t USING gin (to_tsvector('english', body))), which is the main reason anyone wants the two-arg form. #1424 writes it inline as a validated identifier for that reason — happy to talk through it if you see a downside.
2. Naming. I used ToTsVector / ToTsQuery / PlainToTsQuery / WebsearchToTsQuery (camel-casing each word); you used ToTsvector / PlaintoTsquery. Either is fine, but we should pick one before both are in the tree. I don't mind switching to yours if Dan prefers it.

If you'd rather keep #1390 as the single PR, that's fine too — I'll close #1424 and can help review or split yours if useful. Just don't want the two to sit blocking each other.

PR #1425, opened 2026-09-01
Reversing a migration that adds a table fails on SQLite with a CASCADE syntax error

### LINKED ISSUE #1420: Drop table migration tries to CASCADE on SQLite
When downgrading a migration that added a table with the SQLite backend, Piccolo tries to run a query like 

```
DROP TABLE "table_name" CASCADE
```

SQLite does not support CASCADE so this results in an error

```
sqlite3.OperationalError: near "CASCADE": syntax error
```

I'm not familiar with Piccolo's internals but a quick look around shows that there is some existing logic for this issue here

https://github.com/piccolo-orm/piccolo/blob/3599b756a2e70e18e7758b5bb475d33f5192ca4e/piccolo/table.py#L1667-L1674

Maybe we could push that logic down into the Alter class and solve both paths without duplicating code?

https://github.com/piccolo-orm/piccolo/blob/3599b756a2e70e18e7758b5bb475d33f5192ca4e/piccolo/query/methods/alter.py#L378-L393

juneja-varun:
Closes #1420

I create tables through Piccolo migrations against a local SQLite database, and every so often I need to reverse one - for example rolling back a bad migration during development. Doing that on SQLite currently blows up:

\`\`\`
sqlite3.OperationalError: near "CASCADE": syntax error
\`\`\`

The reverse migration always runs `drop_table(cascade=True)`, which appends `CASCADE` to the `DROP TABLE` statement regardless of the database backend. Postgres and CockroachDB support that syntax, but SQLite doesn't, so the rollback fails outright with no way to complete it.

This PR makes the `cascade` flag depend on the engine, matching the pattern already used elsewhere in the codebase (e.g. `Table.drop_tables()`), so it's only passed on backends that actually support it:

\`\`\`python
cascade=_Table._meta.db.engine_type != "sqlite"
\`\`\`

I reproduced the bug against a real SQLite database with a migration that creates a table and then reverses it, confirmed it throws the exact error above on the current code, and confirmed the fix resolves it. I also added a regression test (`test_add_table_sqlite`) covering this exact forwards/backwards flow on SQLite, since the existing `test_add_table` test is scoped to Postgres/CockroachDB only, which is how this went uncaught. Ran the full SQLite test suite locally - all passing.


dantownsend:
Looks great, thanks 👍

PR #1427, opened 2026-09-02
Automigration silently drops one of two same-named tables in different schemas

### LINKED ISSUE #1426: Automigration creation of same table but different schema
Hello

I found a small bug inside of the DiffableTable logic.
The equals check in DiffableTable should also check the schema for *true* equality.

Proposal:
```
def better_eq(self, value) -> bool:
    if not isinstance(value, DiffableTable):
        return False
    return (
        (self.schema == value.schema)
        and (self.class_name == value.class_name)
        and (self.tablename == value.tablename)
    )


DiffableTable.__eq__ = better_eq
```

Thanks for taking this up.

Cheers

juneja-varun:
Closes #1426

I have the same table name (e.g. `Band`) defined in two different Postgres schemas - a pattern that comes up naturally in multi-tenant setups where each tenant gets their own schema. When I ran `piccolo migrations new my_app --auto` for the first time, only one of the two tables showed up in the generated migration - the other one just silently vanished, with no error or warning.

The cause: `DiffableTable.__eq__`/`__hash__` only compare `class_name` and `tablename`, never `schema`. Auto-diffing puts `DiffableTable` instances into sets to work out what's new/dropped/changed, so two tables that share a name but live in different schemas get treated as the exact same table and deduplicated - one of them just disappears.

I made the equality/hash schema-aware, matching what was proposed on the issue. But that alone isn't quite enough - `SchemaDiffer` also has a separate mechanism (`check_table_schema_changes`) for detecting when a *single* table's schema changes (e.g. moving `Band` from `public` to `reporting`), which matches purely on `class_name` and produces a clean `change_table_schema` operation. Once equality became schema-aware, that legitimate schema-change case started *also* showing up in the add/drop-table set difference, and (for tables with any columns) even got misdetected as a bogus no-op table rename (`rename_table(old_class_name='Band', ..., new_class_name='Band', ...)`) alongside the correct `change_table_schema` statement. So this PR also excludes tables already captured by `check_table_schema_changes` from the add/drop/rename detection, the same way renames are already excluded there.

Reproduced both scenarios directly against `SchemaDiffer` before writing any fix: the reported same-name-different-schema case (confirmed one table silently missing from `create_tables`), and the schema-change regression risk (confirmed the bogus rename statement appearing after a naive fix). Added a regression test for the reported bug, and strengthened the existing `test_change_schema` test (added columns, since the rename-misdetection only triggers when there's an overlapping column name) with an assertion that no bogus rename is generated. Both fail on unpatched code and pass with the fix. Full test suite (718 tests) green, isort/black/flake8 clean.


dantownsend:
Looks good, thanks!

Issue #1070, opened 2024-08-22
Schema generation assumes incorrect default values for UUID columns


Keating950:
## Summary

Database: PostgreSQL 16
Version: 1.17.0

UUID columns generated by running `piccolo schema generate` have incorrect default values. This appears to be because `piccolo.columns.column_types.UUID` assumes that UUID columns with defaults are generated by a database-specific built-in function. This results in incorrect default values being generated.

### Schema definition

```sql
-- Source: https://gist.github.com/kjmph/5bd772b2c2df145aa645b837da7eca74
create or replace function uuid_generate_v7()
returns uuid
as $$
begin
  -- use random v4 uuid as starting point (which has the same variant we need)
  -- then overlay timestamp
  -- then set version 7 by flipping the 2 and 1 bit in the version 4 string
  return encode(
    set_bit(
      set_bit(
        overlay(uuid_send(gen_random_uuid())
                placing substring(int8send(floor(extract(epoch from clock_timestamp()) * 1000)::bigint) from 3)
                from 1 for 6
        ),
        52, 1
      ),
      53, 1
    ),
    'hex')::uuid;
end
$$
language plpgsql
volatile;

create table example (
    id uuid primary key default gen_uuid_v7()
);
```

### Generated table definition

```python
from piccolo.table import Table
from piccolo.columns.column_types import Serial
from piccolo.columns.column_types import Timestamp
from piccolo.columns.column_types import UUID
from piccolo.columns.column_types import Varchar
from piccolo.columns.defaults.timestamp import TimestampNow
from piccolo.columns.defaults.uuid import UUID4
from piccolo.columns.indexes import IndexMethod


class Example(Table, tablename="example"):
    id = UUID(
        default=UUID4(),  # incorrect: Translates to gen_random_uuid on the server
        null=False,
        primary_key=True,
        unique=False,
        index=True,
        index_method=IndexMethod.btree,
        db_column_name=None,
        secret=False,
    )


class Migration(Table, tablename="migration"):
  ...  # etc
```

dantownsend:
Thanks for reporting this, and the detailed description.

With ``UUID`` columns, this is what we use for default values:

https://github.com/piccolo-orm/piccolo/blob/e14b8a567b44028db9959bd742871fbb7fda4ee9/piccolo/columns/defaults/uuid.py#L8-L22

If generated on the database side, it uses `uuid_generate_v4()`, and if on the Python side it uses `uuid.uuid4()`.

[...]

PR #1431, opened 2026-09-07
Fix schema generate ignoring UUID column defaults

juneja-varun:
Relates to #1070.

I ran `piccolo schema generate` against a table with a `uuid` column defaulting to `gen_random_uuid()`, and the generated model came back with `default=UUID4()` regardless of what the database default actually was - even a completely custom function would get silently replaced the same way. Tracing it down: `COLUMN_DEFAULT_PARSER` maps `UUID` to `None`, so `get_column_default()` never even looks at the column's actual default text for UUID columns, and just falls back to the column constructor's own default. There's dead code a few lines down (`elif column_type is UUID: return uuid.uuid4`) that was clearly meant to handle this and never runs.

Fixed by giving UUID a real pattern that recognizes the builtin functions Piccolo's own `UUID4`/`UUID7` defaults use (`gen_random_uuid()`, the older `uuid_generate_v4()`, and the new Postgres 18 `uuidv7()`), verified against a real local Postgres. This doesn't solve the original report's exact scenario (a custom `uuid_generate_v7()` from an extension) - Piccolo can't guess what an arbitrary function does, so that still falls back to `UUID4()` as before. But it fixes the broader bug of every recognized builtin default being silently dropped, and adds native `uuidv7()` recognition now that `UUID7` exists as a first-class default.


dantownsend:
Makes sense - so we're now distinguishing between UUID4 and UUID7 when auto generating the schema from an existing DB. Thanks!

PR #1434, opened 2026-09-11
fix: use default values when loading fixtures with missing columns

### LINKED ISSUE #1432: Fixtures are inserted with null values while default values can be used
First of all, I'd like to thank you for this awesome library that does just what it is expected from it and which is pretty easy to read! 👏

When loading fixtures, it fails if a column with a default value is not provided. For example, having the following schema:

```py
class User(Table, tablename="users"):
    id = UUID(primary_key=True)
    created_at = Timestamptz()
    updated_at = Timestamptz()
```

The following fixtures file will fail:

```json
{
  "app": {
    "User": [
      {
        "id": "b43392b0-d694-4720-980c-8b1020c92749"
      }
    ]
  }
}
```

... with the following error:

```
asyncpg.exceptions.NotNullViolationError: null value in column "created_at" of relation "users" violates not-null constraint
DETAIL:  Failing row contains (b43392b0-d694-4720-980c-8b1020c92749, null, null).
```

The fixtures app should instead insert `(b43392b0-d694-4720-980c-8b1020c92749)` as `created_at` and `updated_at` are not provided, which is not a problem as `Timestamptz` has a default value.

I see several things in the code that may lead to a fix:

- It my be useful to remove the parameter `include_default_columns` in [apps/fixtures/commands/shared.py#L41](https://github.com/piccolo-orm/piccolo/blob/master/piccolo/apps/fixtures/commands/shared.py#L41). But, we probably want to validate default column when there are provided so that may not be the solution.
- However, digging into it shows that there is not difference between `TableMeta.columns` and `TableMeta.non_default_columns` as both are filled the same way in [table.py#L339](https://github.com/piccolo-orm/piccolo/blob/master/piccolo/table.py#L339). It seems to be a legacy inherited from a [pretty old commit](https://github.com/piccolo-orm/piccolo/commit/6f3e4f9aaaa4a23945aab1601a2f9d42c3284096). Also, `TableMeta.default_columns` are almost never filled. But maybe here my comprehension is wrong and "default" does not mean column with a default value?
- It seems that generated Pydantic models are always defaulted to null, which leads to these conflicting null values (see [utils/pydantic.py#L270](https://github.com/piccolo-orm/piccolo/blob/master/piccolo/utils/pydantic.py#L270). Maybe this default value should be adapted based on the default value of the `Column` instance. Or set to `None` only if `is_optional and not column._meta.default`?

mlbiche:
This PR fixes an issue when columns with default values are missing in a loaded fixtures file.

Closes #1432


UsmanGhias (commented):
Hey there! This is a really nice fix for the fixture loading logic. Utilizing `column.get_default_value()` instead of hardcoding `None` ensures that optional columns properly fall back to their defined schema defaults when omitted from input files.

From a technical standpoint, I appreciate how clean the diff is. It integrates smoothly with the existing Pydantic model creation flow in `piccolo/utils/pydantic.py`. The added test coverage in `test_dump_load.py` alongside the schema updates in the mega app tables gives me a lot of confidence that this behaves correctly across different database types.

One minor thing to keep an eye on is making sure dynamic callables used as defaults (like timestamps or UUID generators) evaluate as expected during the Pydantic validation phase, but overall this looks solid and ready to ship.

Best,
Usman

